PR #6, opened 2022-01-15
Add tx.origin support for Hoax

abigger87:


src/stdlib.sol
```diff
@@ -11,12 +11,12 @@ abstract contract stdCheats {
     // Setup a prank from an address that has some ether
     function hoax(address who) public {
         vm_std_cheats.deal(who, 1 << 128);
-        vm_std_cheats.prank(who);
+        vm_std_cheats.prank(who, who);
```
brockelmore, line 14: default shouldnt be to override origin. We can overload these as (address, address) or (address, address, uint256) for the give example

brockelmore (changes requested):

brockelmore (approved):
lgtm

PR #11, opened 2022-01-31
add deployCode helper

ncitron:
Adds helpers to make deployment using `vm.getCode` easier. At some point I would like to add a third overload where you can specify the address that the contract will be deployed at, but I think this will require some extra functionality to make constructors work properly.

### DIFF
--- a/src/stdlib.sol
+++ b/src/stdlib.sol
@@ -64,7 +64,7 @@ abstract contract stdCheats {
 
     // Deploys a contract by fetching the contract bytecode from
     // the artifacts directory
-    function transcribe(string memory what, bytes memory args)
+    function deployCode(string memory what, bytes memory args)
         public
         returns (address addr)
     {
@@ -74,7 +74,7 @@ abstract contract stdCheats {
         }
     }
 
-    function transcribe(string memory what)
+    function deployCode(string memory what)
         public
         returns (address addr)
     {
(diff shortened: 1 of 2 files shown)


gakonst (approved):
LGTM - wdyt @brockelmore

src/stdlib.sol
```diff
@@ -61,6 +61,28 @@ abstract contract stdCheats {
         vm_std_cheats.deal(who, give);
         vm_std_cheats.startPrank(who, origin);
     }
+
+    // Deploys a contract by fetching the contract bytecode from
+    // the artifacts directory
+    function transcribe(string memory what, bytes memory args)
```
mds1, line 67: `transcribe` feels too abstract IMO, what about `deployCode` which nicely parallels the `getCode` method used?

brockelmore: ye if we can change it to `deployCode` im 1000% on board this is great. I like maker naming when it isnt easily captured what it does without being overly verbose but `deployCode` isnt too verbose

brockelmore:
also tests are failing - not entirely sure why... if we fix up tests & change name will merge. ty for ur pr ser

ncitron:
It should be good to go. The tests failing must have been a fluke since I just set up the workflow to run on my fork and its working fine now.

PR #12, opened 2022-02-02
Support bytes/string keys in stdStorage

shuklaayush:
- Changes `StdStorage._keys` to a `bytes` array instead of `bytes32[]` to support bytes/string keys
- Removes `flatten` (can just use `abi.encodePacked`)
- Replaces `stdStorage.with_key` functions with a single `stdStorage.with_keys` that takes an abi encoded input

I'm not sure about the last one since it adds a bit of bloat (extra `abi.encode`) in the calling context. I can add the old functions back if required

### DIFF
--- a/src/stdlib.sol
+++ b/src/stdlib.sol
@@ -101,12 +101,11 @@ library stdError {
 struct StdStorage {
     mapping (address => mapping(bytes4 => mapping(bytes32 => uint256))) slots;
     mapping (address => mapping(bytes4 =>  mapping(bytes32 => bool))) finds;
-    
-    bytes32[] _keys;
+
+    bytes _keys;
     bytes4 _sig;
     uint256 _depth;
     address _target;
-    bytes32 _set;
 }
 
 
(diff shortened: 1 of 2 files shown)


brockelmore:
this is brutal for UX. maybe instead have a second set of keys that is the bytes arrays and further overload with_keys to write to this and update the surrounding stuff.

in general `string` and `bytes` keys are dumb and developers shouldn't use them if at all possible as under the hood it just keccaks them anyway

brockelmore:
alternatively, keep keys as bytes, and replace `with_keys` to just concatenate the input to the bytes

shuklaayush:
Closing this since an overloaded `with_key(bytes)` doesn't work for calls like `with_key("abc")`. "abc" is ambiguous in solidity since it's implicitly convertible to both bytes32 and bytes, so there's a compilation error. I could add a with_bytes(string) but that doesn't feel like good ux either (since both with_key and with_bytes would work for "abc")

PR #17, opened 2022-02-20
feat: add `tip` stdcheat

ZeroEkkusu:
Gm! This is a new `tip` stdcheat.

### Motivation
Simplify funding an account with a set amount of an arbitrary ERC20 token.

### Solution
A convenient solution you can use in tests:
```solidity
tip(alice, 10000e18, dai);

// Let's test it
assertEq(daiToken.balanceOf(alice), 10000e18); // [PASS]
```
Especially useful when forking.

Edit: Need to make a few changes to make it work with ~~every~~ a majority of tokens. **✔️ DONE**

### DIFF
--- a/src/stdlib.sol
+++ b/src/stdlib.sol
@@ -5,9 +5,12 @@ import "./Vm.sol";
 
 // Wrappers around Cheatcodes to avoid footguns
 abstract contract stdCheats {
-    // we use a custom name that is unlikely to cause collisions so this contract
+    using stdStorage for StdStorage;
+
+    // we use a custom names that are unlikely to cause collisions so this contract
     // can be inherited easily
     Vm constant vm_std_cheats = Vm(address(uint160(uint256(keccak256('hevm cheat code')))));
+    StdStorage std_store_std_cheats;
 
     // Skip forward or rewind time by the specified number of seconds
     function skip(uint256 time) public {
@@ -65,13 +68,10 @@ abstract contract stdCheats {
     // Allows you to set the balance of an account for a majority of tokens
     // Be careful not to break something!
     function tip(address token, address to, uint256 give) public {
-        vm_std_cheats.record();
-        (bool sent, ) = token.call(abi.encodeWithSelector(0x70a08231, to));
-        require(sent, "Failed to access `balanceOf`. Make sure this is an ERC20 token.");
-        (bytes32[] memory reads, ) = vm_std_cheats.accesses(token);
-        vm_std_cheats.store(token, reads[0], bytes32(give));
-        ( , bytes memory data) = token.call(abi.encodeWithSelector(0x70a08231, to));
-        require(uint256(bytes32(data)) == give, "Could not tip. You will have to set the balance manually.");
+        std_store_std_cheats.target(token)
+            .sig(0x70a08231)
+            .with_key(to)
+            .checked_write(give);
     }
 
     // Deploys a contract by fetching the contract bytecode from
--- a/src/stdlib.sol
+++ b/src/stdlib.sol
@@ -75,8 +75,8 @@ abstract contract stdCheats {
         vm_std_cheats.store(which, reads[0], bytes32(give));
     }
     
-    // If a token's `balanceOf` method does not return the balance immediately,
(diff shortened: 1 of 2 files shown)


ZeroEkkusu:
# Update

- [EIP-20: Token Standard](https://eips.ethereum.org/EIPS/eip-20) requires only that the token has the [`balanceOf`](https://eips.ethereum.org/EIPS/eip-20#balanceof) method, not how to store balances or report them.
- We can only assume that the token will report a balance honestly. Also, different tokens store balances in different ways. For example, DAI stores them in `balanceOf`, 1INCH in `_balances`, MATIC in `balances`.

## Solution

- A majority of tokens read a balance  immediately when the `balanceOf` method is called, but we must not assume this to always be the case.
- I have changed how the `tip` stdcheat works, so it can always set a balance to what we want, when the balance is read immediately after calling `balanceOf`. If `balanceOf` reads some other values before reading a balance, you must specify the name of the mapping that holds balances:
```solidity
// WHEN READ IMMEDIATELY:

/// function balanceOf(address account) public view override returns (uint256) {
///     return _balances[account];
/// }

tip(address(1inch), 10000e18, alice);


// WHEN NOT READ IMMEDIATELY:

/// function balanceOf(address account) public view override returns (uint256) {
///     bytes32 something = someStateVariable;
///     ...
///     return _balances[account];
/// }

tip(address(atypicalToken), "_balances", 10000e18, alice);
```

 # Ready for review!

brockelmore:
is this ready for review now?

ZeroEkkusu:
> is this ready for review now?

Yes, it is.

src/stdlib.sol
```diff
@@ -62,6 +62,18 @@ abstract contract stdCheats {
         vm_std_cheats.startPrank(who, origin);
     }
 
+    // Allows you to the balance of an account for a majority of tokens
+    // Be careful not to break something!
+    function tip(address to, uint256 give, address which) public {
```
brockelmore, line 67: Why dont you just use stdstore instead? i imagine there will be more useful things like this that pop up

brockelmore: can we change the variable arrangement as well? i think (token, to, give) is most intuitive

brockelmore (changes requested):

src/stdlib.sol
brockelmore: https://mirror.xyz/brocke.eth/PnX7oAcU4LJCxcoICiaDhq_MUUu9euaM8Y5r465Rd2U < see here. I use stdstore

ZeroEkkusu: thanks, brock. i didn't know u can use it like that.

the article was informative!

ZeroEkkusu: it's ready, ser @brockelmore

brockelmore (approved):
lgtm!

shuklaayush:
Hey
Thought I'd just comment here instead of creating a new issue. Isn't the term "tip" a bit misleading since the cheat code is setting the balance and not adding to it? Maybe an overloaded `deal` would be better. Also, might make sense to update `totalSupply` along with the balance

ZeroEkkusu:
@shuklaayush

I had considered making `tip` add to a balance, but setting a balance is more useful. Do you have a better name for it? I don't want hevm cheatcodes and stdcheats to be confused, which is why it is not called `deal`.

Since this stdcheat takes the address of an *arbitrary* ERC20 token, we aren't aware how the token operates. For example, it's `totalSupply` might be essential for some internal logic and changing it could break something, which is why I went with a minimal approach.

shuklaayush:
Ah, cool. Didn't realize that you intentionally avoided naming it `deal`. Only alternative I can come up with off the top of my head is `allot` but I guess `tip` is fine too

While I agree that it's better to keep the logic minimal, I think practically it's more likely that someone will run into an error when `totalSupply` isn't equal to the sum of balances as compared to an error where the token contract implements some special logic around `totalSupply`

Anyways, I don't have strong opinions on either. Thought I'd share my two cents since I did some similar stuff when testing out contracts with foundry

PR #18, opened 2022-02-23
fix: remove public from constructor

rootulp:
Fixes this warning
```
Warning: Visibility for constructor is ignored. If you want the contract
to be non-deployable, making it "abstract" is sufficient.
   -->
/Users/rootulp/git/rootulp/forge-std/src/test/StdStorage.t.sol:252:5:
    |
252 |     constructor() public {
    |     ^ (Relevant source part starts here and spans across multiple
lines).
```

### DIFF
--- a/src/test/StdStorage.t.sol
+++ b/src/test/StdStorage.t.sol
@@ -217,7 +217,7 @@ contract StdStorageTest is DSTest {
     function testFailStorageNativePack() public {
         stdstore.target(address(test)).sig(test.tA.selector).find();
         stdstore.target(address(test)).sig(test.tB.selector).find();
-        
+
         // these both would fail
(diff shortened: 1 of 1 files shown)


mds1:
@brockelmore the pragma in this file is `pragma solidity >=0.6.0 <0.9.0;`, but 0.6.x requires the constructor to be `public`, so after merging this PR `StdStorage.t.sol` is inconsistent with the pragma.

Could either re-add `public` or bump this file to `pragma solidity >=0.7.0 <0.9.0;`, both of which are fine with me.

rootulp:
Sorry that's my fault. Thanks for flagging @mds1! I'll create a PR to bump the pragma to `pragma solidity >=0.7.0 <0.9.0;`

PR #19, opened 2022-02-23
fix: increase min solidity version to 0.7.0

rootulp:
Previously StdStorage.t.sol claimed to support `pragma solidity >=0.6.0 <0.9.0;`.
However https://github.com/brockelmore/forge-std/pull/18 introduced a bug because it removed `public` from a constructor. This
feature was only added in [solidity 0.7.0](https://docs.soliditylang.org/en/v0.7.0/070-breaking-changes.html#functions-and-events).

> Visibility (`public` / `external`) is not needed for constructors anymore

So this change increases the minimum solidity version needed

### DIFF
--- a/src/test/StdStorage.t.sol
+++ b/src/test/StdStorage.t.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: Unlicense
-pragma solidity >=0.6.0 <0.9.0;
+pragma solidity >=0.7.0 <0.9.0;
 
 import "ds-test/test.sol";
 import "../stdlib.sol";


rootulp:
I choose to increase the min required solidity version instead of reverting https://github.com/brockelmore/forge-std/pull/18 because reverting would re-introduce the warning.

Question for reviewers: I think we could've caught this prior to merge if this repo had a GitHub CI check to compile the code base with different versions of solidity compiler. Is such a GitHub CI check common? Would it be helpful?

brockelmore:
I am just going to revert the previous so that users of 0.6.0 can use the contract

rootulp:
Sounds good. Sorry for the churn.

PR #23, opened 2022-03-03
Streamlined console log functionality with forge

FredCoen:
Just added the console contract from the forge README. Not sure about the licensing here, if it needs change


brockelmore (approved):
lgtm thanks!

PR #28, opened 2022-03-18
fix(stdlib): make variables not meant for end-users `private`

ZeroEkkusu:
Does anyone use these?

```solidity
abstract contract stdCheats {
    Vm constant vm_std_cheats = Vm(address(uint160(uint256(keccak256('hevm cheat code')))));
    StdStorage std_store_std_cheats;
```

Let's make them private so they don't pop up as suggestions (autocomplete). Better UX.


brockelmore (approved):
ty ser

PR #30, opened 2022-03-23
BREAKING CHANGE: deprecate `tip`, add universal `deal` interface

### LINKED ISSUE #33: Unused local variable "bytes memory b"
After running the steps mentioned in the [Working on an Existing Project](https://book.getfoundry.sh/projects/working-on-an-existing-project.html#working-on-an-existing-project) chapter in the Foundry Book, I got the following warnings emitted in the console:

```text
Warning: Visibility for constructor is ignored. If you want the contract to be non-deployable, making it "abstract" is sufficient.
   --> /Users/paulrberg/workspace/pad/forge-std/src/test/StdStorage.t.sol:252:5:
    |
252 |     constructor() public {
    |     ^ (Relevant source part starts here and spans across multiple lines).



Warning: Unused local variable.
   --> /Users/paulrberg/workspace/pad/forge-std/src/test/StdError.t.sol:106:9:
    |
106 |         bytes memory b = someBytes;
    |         ^^^^^^^^^^^^^^
```

As per https://github.com/brockelmore/forge-std/pull/19, it looks like the first one is expected. But it looks like the second one may be not.

Is the `b` variable purposefully left unused, or is this an issue that should be resolved?

ZeroEkkusu:
### Deprecate `tip`

The `tip` stdcheat is now marked as `DEPRECATED`.

```solidity
// DEPRECATED: Use `allot` instead
function tip(address token, address to, uint256 give) public {
```

It still is possible to use but a warning log is emitted to notify the user about the change:

```ignore
├─ emit WARNING_Deprecated(msg: "The `tip` stdcheat has been deprecated. Use `allot` instead.")
```

### Introduce `allot`

First, the new `allot` stdcheat better communicates what this stdcheat does (***sets*** the balance of an account):

```solidity
allot(dai, address(this), 10000e18);
```

Second, the user can now choose whether to update `totalBalance` of the token, by passing an optional 4th argument `bool adjust`:

```solidity
allot(dai, address(this), 10000e18, true);
```

If the new balance is lesser than the previous balance, the difference is subtracted from `totalSupply`, and vice versa.

<br>

New tests are added and the old ones removed.

Also, fixed a few typos.

#### Edit
See the next comment.

Fixes #33

### DIFF
--- a/src/stdlib.sol
+++ b/src/stdlib.sol
@@ -9,8 +9,6 @@ abstract contract stdCheats {
 
     event WARNING_Deprecated(string msg);
 
-    // we use custom names that are unlikely to cause collisions so this contract
-    // can be inherited easily
     Vm private constant vm_std_cheats = Vm(address(uint160(uint256(keccak256('hevm cheat code')))));
     StdStorage private std_store_std_cheats;
 
--- a/src/test/StdCheats.t.sol
+++ b/src/test/StdCheats.t.sol
@@ -109,7 +109,7 @@ contract StdCheatsTest is DSTest, stdCheats {
 
 contract Bar {
     constructor() public {
-        /// `ALLOT` STDCHEAT
+        /// `DEAL` STDCHEAT
         totalSupply = 10000e18;
         balanceOf[address(this)] = totalSupply;
     }
@@ -127,7 +127,7 @@ contract Bar {
         require(tx.origin == expectedOrigin, "!prank");
     }
 
-    /// `ALLOT` STDCHEAT
+    /// `DEAL` STDCHEAT
     mapping (address => uint256) public balanceOf;
     uint256 public totalSupply;
 }


src/stdlib.sol
```diff
@@ -65,19 +67,53 @@ abstract contract stdCheats {
...
+        std_store_std_cheats
+            .target(token)
+            .sig(0x70a08231)
+            .with_key(to)
+            .checked_write(give);
+    }
+
+    // Set the balance of an account for any ERC20 token
+    // Use the alternative signature to update `totalSupply`
+    function allot(address token, address to, uint256 give) public {
```
mds1, line 82: Thoughts on just calling this `deal` like the standard cheatcode for setting ETH balance? Seems unnecessary to have two different method names, plus you can special case the `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` address to fallback to `vm.deal`, that way all ETH/token balance setting has the same interface

mds1: Part of me likes it, part me feels like it clutters the interface because you can accomplish the same by only setting balances directly, e.g. `deal(dai, address(this), dai.balanceOf(address(this)) + 10000e18)`. I think I lean towards keeping the interface simpler and have `deal` just set directly the balance

brockelmore: i like moving to `deal` and i like having token at the front personally

ZeroEkkusu: i'm working on this now

ZeroEkkusu:
**`allot` is now `deal`.**

### Usage

```solidity
deal(address(this), 1 ether);
// give yourself 1 ether

deal(dai, address(this), 10000e18);
// give yourself 10,000 DAI without updating DAI's `totalSupply`

deal(dai, address(this), 10000e18, true);
// give yourself 10,000 DAI and update DAI's `totalSupply` accordingly
```

src/stdlib.sol
brockelmore: how do you guys think we should handle breaking changes like this?

mds1: Right now we have no tags/releases, so my suggestion would be to tag the latest master (d880879) as `v0.1.0`, then once this is merged tag it as `v0.2.0`. Going forwards we'd then bump the patch version on each merge to master (or at least each merge worthy of a release), and bump the minor version on breaking changes like this one. The `v1.0.0` release of forge-std can be roughly synced with foundry's 1.0.0 release

This also helps forge-std prepare for https://github.com/gakonst/foundry/issues/401

ZeroEkkusu: Sounds good.

1. Release `v0.1.0` now
2. Merge this breaking change and release `v0.2.0`
3. Add new feature and release `v0.2.1`
4. Major change something and tag `v0.2.2`
5. Add new feature and release `v0.2.3`
6. Minor change something
7. \* PR with breaking change(s) \*
8. Release `v0.2.4` with work so far, if not released
9. Merge the breaking change(s) and release `v0.3.0`

Would you change something here?

Release notes:

1. new features
2. breaking changes
3. fixes
4. anything else

ZeroEkkusu: These are the release notes for `v0.2.0`

## New features

- New stdcheat: **`deal`** lets you set the ETH or ERC20 token balance of an account from the same interface (no need to use Hevm's `deal` for ETH and a separte stdcheat for ERC20 tokens anymore).
- The `deal` stdcheat (previously `tip`) can now optionally adjust the total supply of an ERC20 token after changes have been made to an account's balance.

## Breaking changes

- The `tip` stdcheat has been deprecated. Use the new `deal` stdcheat instead.
- The `lowLevelError` has been deprecated. Use Hevm's `expectRevert` without any arguments instead.

## Fixes

- Fix a test that was causing the `Unused local variable` compiler warning.

PR #31, opened 2022-03-24
Allow calling expectRevert without any arguments

m1guelpf:
Added in https://github.com/gakonst/foundry/pull/918


ZeroEkkusu:
Thanks!

PR #36, opened 2022-04-14
Improve stdlib import ergonomics

gakonst:
Instead of having to do multiple imports, we now have a canonical `import forge-std/Test.sol` which is inherited from as `contract MyTest is Test`.

Test.sol now contains:
1) Cheatcodes under `vm`
2) all ds-test assert/log emits because it inherits from `DSTest`
3) console.log because it now imports it
4) `stdstore` which is made available by doing `using stdStorage for StdStorage` because libraries are not inherited (https://ethereum.stackexchange.com/a/98923)

This improves testing UX a lot, as you don't need to really remember any special imports or magic variables. You just import `Test` and you're g2g.

```solidity
pragma solidity 0.8.10;

import "forge-std/Test.sol";

    function testExample() public {
        vm.roll(100);
        console.log(1);
        emit log("hi");
        assertTrue(true);
    }
}
```

**This is a BREAKING change.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,14 +19,11 @@ See the contract itself for all error codes.
 
 ```solidity
 
-import "ds-test/test.sol";
-import "forge-std/stdlib.sol";
-import "forge-std/Vm.sol";
+import "forge-std/Test.sol";
 
-contract TestContract is DSTest {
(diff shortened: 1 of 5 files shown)


brockelmore (approved):
nice

PR #2, opened 2022-01-06
feat: add hoax (prank & deal)

transmissions11:

PR #37, opened 2022-04-17
Add base contracts to simplify mocking on tests

bbonanno:
Adds wrappers and conveniece functions and contracts to make mocking easier with foundry.
It also adds a buch of tests, mostly as self-documenting examples.

### DIFF
--- a/src/Mocks.sol
+++ b/src/Mocks.sol
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: Unlicense
+pragma solidity >=0.6.0 <0.9.0;
+
+import "./Vm.sol";
+import "./Test.sol";
+
+contract Mocks is Test {
+    /// @dev creates a strict mock and labels it, it's a good practice to always use strict mocks unless you really, really want leniency.
+    function mock(string memory label) internal returns (address mock_) {
+        mock_ = address(new StrictMock());
+        vm.label(mock_, label);
+    }
+
+    /// @dev same as `mock` but it deploys a lenient mock.
+    function lenientMock(string memory label) internal returns (address mock_) {
+        mock_ = address(new LenientMock());
+        vm.label(mock_, label);
+    }
+
+    /// @dev creates a strict mock (and labels it) on a pre-specified address,
+    /// useful to replace existing contracts on a fork or to test contracts that rely on deterministic addresses like uniswap pools
+    function mockAt(address where, string memory label)
+        internal
+        returns (address)
+    {
+        vm.etch(where, vm.getCode("Mocks.sol:StrictMock"));
+        vm.label(where, label);
+        return where;
+    }
+
+    /// @dev same as `mockAt` but it deploys a lenient mock.
+    function lenientMockAt(address where, string memory label)
+        internal
+        returns (address)
+    {
+        vm.etch(where, vm.getCode("Mocks.sol:LenientMock"));
+        vm.label(where, label);
+        return where;
+    }
+
+    function mockVoidCall(address _mock, bytes memory _calldata) internal {
+        vm.mockCall(
+            _mock,
+            _calldata,
+            abi.encode(0) // Would be nice to have a `vm.mockCall` that didn't require a return value
+        );
+    }
+}
+
+/// @dev Empty contract to deploy and use as a Mock, needed due to the fact that `vm.mockCall()` doesn't with non-contract addresses
+/// @notice this contract will purposely fail if called on an un-stubbed method
+contract StrictMock {
+    fallback() external payable {
+        revert("Not mocked!");
+    }
+}
+
+/// @dev Empty contract to deploy and use as a Mock, needed due to the fact that `vm.mockCall()` doesn't with non-contract addresses
+/// @notice this contract will do nothing if called on a method that wasn't previously stubbed
+contract LenientMock {
+    fallback() external payable {}
+}
--- a/src/test/Mocks.t.sol
+++ b/src/test/Mocks.t.sol
@@ -1,0 +1,158 @@
+// SPDX-License-Identifier: Unlicense
+pragma solidity >=0.6.0 <0.9.0;
+
+import "../Test.sol";
+import "../Mocks.sol";
+
+contract MocksTest is Test, Mocks {
+    function testFailStrictMockWhenUnstubbedFunctionIsCalled() public {
+        address thirdParty = mock("ThirParty");
+        SomeThirdParty(thirdParty).foo();
+    }
+
+    function testStrictMockCanStubAndVerifyParameterlessVoidFunctions() public {
+        address thirdParty = mock("ThirParty");
+
+        mockVoidCall(
+            thirdParty,
+            abi.encodeWithSelector(SomeThirdParty.foo.selector)
+        );
+
+        vm.expectCall(
+            thirdParty,
+            abi.encodeWithSelector(SomeThirdParty.foo.selector)
+        );
+
+        SomeThirdParty(thirdParty).foo();
+    }
+
+    function testStrictMockCanStubAndVerifyVoidFunctions() public {
+        address thirdParty = mock("ThirParty");
+
+        // When making verifications, it's easier to leave the function params out cause the error message
+        // in case there's a mismatch parameter is better
+        mockVoidCall(
+            thirdParty,
+            abi.encodeWithSelector(SomeThirdParty.bar.selector)
+        );
+
+        vm.expectCall(
+            thirdParty,
+            abi.encodeWithSelector(SomeThirdParty.bar.selector, 42)
+        );
+
+        SomeThirdParty(thirdParty).bar(42);
+    }
+
+    function testFailStrictMockOnUnexpectedParams() public {
+        address thirdParty = mock("ThirParty");
+
+        // When making verifications, it's easier to leave the function params out cause the error message
+        // in case there's a mismatch parameter is better
+        mockVoidCall(
+            thirdParty,
+            abi.encodeWithSelector(SomeThirdParty.bar.selector)
+        );
+
+        vm.expectCall(
+            thirdParty,
+            abi.encodeWithSelector(SomeThirdParty.bar.selector, 42)
+        );
+
+        SomeThirdParty(thirdParty).bar(43);
+    }
+
+    function testStrictMockCanStubAndVerifyFunctions() public {
+        address thirdParty = mock("ThirParty");
+
+        // When making verifications, it's easier to leave the function params out cause the error message
+        // in case there's a mismatch parameter is better
+        vm.mockCall(
+            thirdParty,
+            abi.encodeWithSelector(SomeThirdParty.baz.selector),
+            abi.encode(69420)
+        );
+
+        vm.expectCall(
+            thirdParty,
+            abi.encodeWithSelector(SomeThirdParty.baz.selector, 42)
+        );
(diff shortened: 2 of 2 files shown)


brockelmore:
Curious if people will actually use this? most of the time when people mock I think they just add mock functionality to a in-development contract. Open to it but want to make sure its worth the mental overhead for users of this library

bbonanno:
> Curious if people will actually use this? most of the time when people mock I think they just add mock functionality to a in-development contract. Open to it but want to make sure its worth the mental overhead for users of this library

That'd be using a stub if I get what you're saying, which is a valid thing, but different use case imho.
I found myself using this pattern quite often, as I can summon a random dependency in a single line.
I also find the "fail by default" very helpful as what you want the least is accidental, undesired interactions with 3rd parties.

As a general rule, I use stubs for my e2e testing, and mocks for my unit (and sometimes integration) tests.

Not sure how we can collect consensus though :(

mds1:
Hmm, I'm not convinced this should be part of `forge-std`. Main objection is that I'm not sure I see the utility, and wonder if it's something we should improve directly within forge or with a simple helper method or two in `forge-std`. Some thoughts:

1. Mainly: It seems like a single `mockCall` method that just automatically does the required `etch` call would be a lot simpler and add most of the same utility.
2. More of an implementation detail: `forge-std` doesn't add labels for any other utils, such as `deployCode`, so for consistency I don't think we should add them here either: The contract will already be labeled with it's name, and it's simple to add your own label if you need anyway. Doing this makes functions `mock` and `lenientMock` unnecessary as the become one-line methods, and effectively boils things down to just needing a single helper method or two

bbonanno:
> Hmm, I'm not convinced this should be part of `forge-std`. Main objection is that I'm not sure I see the utility, and wonder if it's something we should improve directly within forge or with a simple helper method or two in `forge-std`. Some thoughts:
> 
> 1. Mainly: It seems like a single `mockCall` method that just automatically does the required `etch` call would be a lot simpler and add most of the same utility.
> 2. More of an implementation detail: `forge-std` doesn't add labels for any other utils, such as `deployCode`, so for consistency I don't think we should add them here either: The contract will already be labeled with it's name, and it's simple to add your own label if you need anyway. Doing this makes functions `mock` and `lenientMock` unnecessary as they become one-line methods, and effectively boils things down to just needing a single helper method or two

I like #1, but it'd also require better error output when you call stuff you didn't mock on that address, or at least a way to signal that.

For #2, if you're mocking multiple contracts (as I do) and you don't add labels, it'll be harder to identify each mock in the stacktrace.

mds1:
> I like https://github.com/foundry-rs/forge-std/pull/1, but it'd also require better error output when you call stuff you didn't mock on that address, or at least a way to signal that.

I may be misunderstanding something here, but isn't that solved by the suggestion of doing the `etch` call? I wasn't very clear, but the suggestion is to take the bytecode of `StrictMock` and etch that in the helper method, so the call reverts if not mocked. There might also be a way to improve this at the forge level, but not certain.

Then the UX of `mockCall` is the same as `vm.mockCall`, but with the etching done for you, so it's a much smaller change to `forge-std`

bbonanno:
> > I like #1, but it'd also require better error output when you call stuff you didn't mock on that address, or at least a way to signal that.
> 
> I may be misunderstanding something here, but isn't that solved by the suggestion of doing the `etch` call? I wasn't very clear, but the suggestion is to take the bytecode of `StrictMock` and etch that in the helper method, so the call reverts if not mocked. There might also be a way to improve this at the forge level, but not certain.
> 
> Then the UX of `mockCall` is the same as `vm.mockCall`, but with the etching done for you, so it's a much smaller change to `forge-std`

Ohh, yeah, I see, I love it
OTOH
We will still need the existent `mockCall`, as some times it's useful to just override the behaviour of a single method on an existing contract

mds1:
> We will still need the existent `mockCall`, as some times it's useful to just override the behaviour of a single method on an existing contract

Agreed, and you'd be able to do that with `vm.mockCall` still. But to your point maybe worth renaming this suggested helper method to `mockStrictCall` or `mockCallWithEtch` or something similar, so it's clear it behaves differently than the cheatcode.

bbonanno:
> > We will still need the existent `mockCall`, as some times it's useful to just override the behaviour of a single method on an existing contract
> 
> Agreed, and you'd be able to do that with `vm.mockCall` still. But to your point maybe worth renaming this suggested helper method to `mockStrictCall` or `mockCallWithEtch` or something similar, so it's clear it behaves differently than the cheatcode.

Yeah, 100%

I guess that means this should move to be a feature request in foundry?

brockelmore:
@bbonanno can you replace everything with just a `mockStrictCall` or `mockCallWithEtch` method as discussed above?

bbonanno:
@brockelmore I've been thinking about this and I think we need both
I'm not sure that just offering a method that's conflating the stubbing method with the creation method is enough, reasons being

- Crating a mock with an explicit label is quite valuable imho, if you have several dependencies and they are all mocked, if one of them fail you'll potentially have a hard time finding which one by just looking at the call trace.

- If you need to inject the mock into the contract you're about to test, you need a mock instance/address but not necessarily know what to stub yet (think about doing this on the setup method and then different tests mock different calls and/or the same call with different params)

- Same as above, but to override a 3rd party on a fork, before knowing how each test is going to interact with it

- If you just inject a random address that'll be (maybe) mocked later, you lose the nice message telling you that you called a method that wasn't mocked (in case you have an unexpected call or whatever). Currently if you call a method on an empty address you get more cryptic error

```solidity
    function testStrictMockWhenUnstubbedFunctionIsCalled() public {
        address thirdParty = mock("ThirParty");
        SomeThirdParty(thirdParty).foo();
    }
```
returns 
```shell
Running 1 test for src/test/Mocks.t.sol:MocksTest
[FAIL. Reason: Not mocked!] testStrictMockWhenUnstubbedFunctionIsCalled() (gas: 48378)
Traces:
  [48378] MocksTest::testStrictMockWhenUnstubbedFunctionIsCalled()
    ├─ [12266] → new ThirParty@0xce71065d4017f316ec606fe4422e11eb2c47c246
    │   └─ ← 61 bytes of code
    ├─ [0] VM::label(ThirParty: [0xce71065d4017f316ec606fe4422e11eb2c47c246], "ThirParty")
    │   └─ ← ()
    ├─ [93] ThirParty::c2985578()
    │   └─ ← 0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000b4e6f74206d6f636b656421000000000000000000000000000000000000000000
    └─ ← "Not mocked!"
```

but
```solidity
    function testStrictMockWhenUnstubbedFunctionIsCalled() public {
        address thirdParty = address(0xf00);
        SomeThirdParty(thirdParty).foo();
    }
```
returns
```shell
Running 1 test for src/test/Mocks.t.sol:MocksTest
[FAIL. Reason: Revert] testStrictMockWhenUnstubbedFunctionIsCalled() (gas: 2889)
Traces:
  [2889] MocksTest::testStrictMockWhenUnstubbedFunctionIsCalled()
    └─ ← ()
```

To be honest, I can easily add a `mockCallWithEtch`, but I don't think it'll be used very often... Also, using that by default would be unnecessary overhead in case the mock already exists at that address.

Thoughts?

mds1:
> Crating a mock with an explicit label is quite valuable imho, if you have several dependencies and they are all mocked, if one of them fail you'll potentially have a hard time finding which one by just looking at the call trace.

For labels, I would prefer not to include this as part of the interface because not all users will need/care about labels for their use case:
- If they need labels, it's one extra line, or a simple helper function to batch mock deployment + labeling.
- If they don't need labels, they're either forced them to come up with a label, or we add overloads for each method, bloating the forge-std interface

For consistency we'd then want to add label overloads to `deployCode`, but afaik no one has requested this so I don't think it's a big need, and I worry about forge-std getting bloated with trivial helpers 🙂 (See https://github.com/foundry-rs/forge-std/pull/42 which was closed for a similar reason)

> \< Remaining bullets about mocks \>

Let me know if I'm understanding the use cases/tradeoffs here correctly:  Strict mocks are effectively the default with `mockCall`, but you get an unclear revert reason in traces, and `StrictMock` solves that. If that's correct, I'd propose instead improving the forge revert reason for cases where a matching function selector is not found, as that occurs in places other than mocks and feels like the right soluton.

For lenient mocks, I'm not sure I understand the use case. Could you clarify a bit?

Lastly, if we end up keeping both strict and lenient mocks, I think we can accomplish the same as this PR with a smaller footprint as shown below. And if we only keep one, this is easy to trim down as needed. This feels a bit cleaner and easier to learn IMO (and I've removed the `mockVoidCall` helper since that falls under the "one line helper method bloat").

```solidity
contract StrictMock {
    fallback() external payable {
        revert("Not mocked!");
    }
}

contract LenientMock {
    fallback() external payable {}
}

contract Mocks is Test {
    function deployMock(bool strict) internal returns (address) {
        return strict ? address(new StrictMock()) : address(new LenientMock());
    }

    function deployMock(bool strict, address where) internal returns (address) {
        string memory which = strict ? "Mocks.sol:StrictMock" : "Mocks.sol:LenientMock";
        vm.etch(where, vm.getCode(which));
        return where;
    }
}
```

bbonanno:
Ok, I can keep the label wrappers on my own utils, as I hate that extra line of code :)

The use case for a `LenientMock` is to save on boilerplate, but that only works for void functions.

Let's say you're testing contract `A` that at some point needs to execute a side-effect in contract `B`

in your test, what matters is to see something like
```solidity
vm.expectCall(address(B), abi.encode......);
```

Now, if you do that on an empty address it'll fail as discussed, so you'd need to stub it, so the code would look like 
```solidity
vm.mockCall(address(B), abi.encode......, abi.encode(0));
vm.expectCall(address(B), abi.encode......);
```

if you have to do that enough number times, it gets annoying to have to mock & verify every time, so a lenient mock solves that.

Compared to a StrictMock, it should be only used in very specific scenarios, but I think it's useful nonetheless

I agree with your first statement, if forge can come up with better errors when calling a method that doesn't exist it'd be better, although not sure if enough as the `mockCall` includes input args, so you may be mocked the wrong args (or the call is made more than once with different ones) and you'd get a "method not found" error.

I like your suggestion on the API, but I wonder, given it's just 2 methods, couldn't we just add it to the `Test` contract?

mds1:
As I think about this more, we may want to close this PR and back up to an issue for some discussion: Should we keep forge-std mocking helpers lightweight, or go for something heavier but more feature filled and robust? There is more to a good set of mocking utils that this PR wouldn't cover that's also worth discussing. In other words, instead of just discussing "strict vs. lenient mock" and their interfaces, we can consider more robust mock contracts that let you mock reverts, out of gas, etc. Gnosis, Waffle, and others have good mock contracts that allow things like this, and it's worth considering integrating those and using forge-std as a wrapper around them instead of reinventing the wheel.

@bbonanno If this sounds good to you I can open an issue to get the discussion going 🙂

bbonanno:
@mds1 I'm all up for it :)

bbonanno:
So far I've been trying to hack around with the tools available, but I'd love to take a step back, define what we want/need and then see how can that be achieved in the best way possible

mds1:
Thanks @bbonanno, appreciate your patience! I'll get that issue open today/tomorrow and tag you in it

PR #38, opened 2022-04-18
feat(test): assertApproxEq, assertNotEq, assertFalse

OliverNChalk:
### Brief

This PR adds the following assertions:
- `assertApproxEq(a, b, reason?)` [uint256, int256]
- `assertNotEq(a, b, reason?)` [address, bytes, int, uint, string]
- `assertFalse(a, reason?)` [bool]

### Deliverables

Unsure if this is of interest at this level or should be pitched to `ds-test`. It however seems that ds-test is not being actively extended at the moment, hence the PR here. We use a forked version of forge-std and find these asserts useful but understand if this is out of scope/not going to be merged.

Would be good to get a decision/some discussion on where most appropriate to extend the current set of assertions which have some gaps imo.


bbonanno:
For what is worth, was going to propose/add similar stuff :)

brockelmore:
@transmissions11 curious your thoughts here, I know you were going to maybe do a clean-room rewrite of some existing contracts into `forge-std`

brockelmore:
Can you just yonk the various methods you want from [here](https://github.com/Rari-Capital/solmate/blob/main/src/test/utils/DSTestPlus.sol)? Have permission from t11 to do so

OliverNChalk:
Chucked a few more commits on this, largely driven by the suggestion to use @transmissions11's impls. To summarize, the following should be resolved:

- [ ] assertApproxEq implementation from t11s
- [ ] assertRelApproxEq implementation from t11s
- [ ] assertEq(bool, bool) implementation from t11s (note name is diff)
- [ ] assertEq(bytes, bytes) implementation from t11s (note name is diff)

Areas to cover/discuss:

- [ ] Naming of assertEq variants (solmate impl has things like assertBoolEq, where I've tried to align with ds-test's assertEq argument overloading approach)
- [ ] Maintaining error message support (i did this to align with ds-test but I personally rarely use it and would be happy without)
- [ ] Scope (i have not picked up things like gas metering, fail(), bound) to keep the scope of PR just to enhancing assets)
- [ ] Log alignment style, mainly did this to sanity check spacing but unsure if this is wanted in here

src/Test.sol
```diff
@@ -139,6 +139,256 @@ abstract contract Test is DSTest {
...
+        if (delta > maxDelta) {
+            emit log_named_string   ("Error", err);
+            emit log_named_uint     ("  Expected", b);
+            emit log_named_uint     ("    Actual", a);
+            emit log_named_uint     (" Max Delta", maxDelta);
+            emit log_named_uint     ("     Delta", delta);
+            fail();
+        }
+    }
+    function assertApproxEq(
```
mds1, line 181: Seems you have the `assertApproxEq` methods duplicated?

src/Test.sol
```diff
@@ -139,6 +139,256 @@ abstract contract Test is DSTest {
...
+            emit log            ("Error: a ~= b not satisfied [int]");
+            emit log_named_int  ("  Expected", b);
+            emit log_named_int  ("    Actual", a);
+            emit log_named_int  (" Max Delta", maxDelta);
+            emit log_named_int  ("     Delta", delta);
+            fail();
+        }
+    }
+
+    function assertApproxEq(
```
mds1, line 198: Thoughts on naming the methods `assertApproxEqAbs` and `assertApproxEqRel`?  A bit different from solmate's naming but more clear IMO

src/Test.sol
```diff
@@ -139,6 +139,256 @@ abstract contract Test is DSTest {
...
+            emit log_named_decimal_uint ("     % Delta", percentDelta, 18);
+            fail();
+        }
+    }
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                    NOT EQUAL
+    //////////////////////////////////////////////////////////////////////////*/
+
+    function assertNotEq(address a, address b) internal {
```
mds1, line 263: Personally I'd prefer not to include `assertNotEq` methods in forge-std. It adds a lot of boilerplate to have all the same overloads as `assertEq`, and I don't think "not equal" assertions are very useful / good practice. It's always stronger to assert that two values equal, or that a value has changed from `x` to `y`. Asserting things differ is a weak condition to test

src/Test.sol
```diff
@@ -139,6 +139,256 @@ abstract contract Test is DSTest {
...
+            emit log_named_int      ("    Actual", a);
+            emit log_named_int      (" Max Delta", maxDelta);
+            emit log_named_int      ("     Delta", delta);
+            fail();
+        }
+    }
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                    REL APPROX EQUAL
+    //////////////////////////////////////////////////////////////////////////*/
```
mds1, line 218: Nit, but really just need one section header for all the assertions to differentiate them from the other helper methods. Maybe just called "Assertions"

src/Test.sol
```diff
@@ -139,6 +139,256 @@ abstract contract Test is DSTest {
...
+            assertNotEq(a, b);
+        }
+    }
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                    BOOL ASSERTS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    function assertEq(bool a, bool b) internal virtual {
+        b ? assertTrue(a) : assertFalse(a);
```
mds1, line 356: This will result in a user passing in two values but failing with "Just assertion" failed, instead of differentiating between expected and actual like the rest of the `assertEq` overloads. I'd suggest just changing these to be like the others

```solidity
function assertEq(bool a, bool b) internal {
    if (a != b) {
      emit log("Error: a == b not satisfied [bool]");
      emit log_named_string("  Expected", b ? "true" : "false");
      emit log_named_string("    Actual", a ? "true" : "false");
      fail();
    }
  }
```

mds1 (commented):
Thanks a lot @OliverNChalk!! Looking forward to having these in forge-std

> Naming of assertEq variants (solmate impl has things like assertBoolEq, where I've tried to align with ds-test's assertEq argument overloading approach)

Agree with naming the overloads `assertEq` instead of solmate's `assertBoolEq`

> Maintaining error message support (i did this to align with ds-test but I personally rarely use it and would be happy without)

Supporting error messages for consistency seems like the right approach, though I also rarely use them

> Scope (i have not picked up things like gas metering, fail(), bound) to keep the scope of PR just to enhancing assets)

`bound` is a very popular solmate helper for fuzz tests that is recommended in the docs, so that one is definitely worth upstreaming here. Gas metering and others are more questionable so we can leave them out for now / discuss in a separate issue if people want to consider them

>  Log alignment style, mainly did this to sanity check spacing but unsure if this is wanted in here

Sorry can you clarify what you mean here?

src/Test.sol
OliverNChalk: Four combinations:

- `uint256`
- `uint256` + user err message
- `int256`
- `int256` + user err message

src/Test.sol
OliverNChalk: I prefer this, have made the change against the PR. Can unwind if people disagree

src/Test.sol
OliverNChalk: Okay to drop this, looking through my usage I only have two instances of using notEq and that's for Create2 shenanigans where i could just independently generate the expected address. So agree this is a sign of laziness/code smell

src/Test.sol
OliverNChalk: Due to the scope for a huge amount of overloading, I found code regions helpful for ensuring proper organisation of related but overloaded functions. That said, if people (read maintainers of forge-std) prefer to drop down to just one ASSERTIONS heading then happy for that change to get rolled in prior to this merging

src/Test.sol
OliverNChalk: Fixed by 97f5f77e5ba3c28c2b9440d20202bd954d87da4d

OliverNChalk:
> > Log alignment style, mainly did this to sanity check spacing but unsure if this is wanted in here

> Sorry can you clarify what you mean here?

To help ensure the error messaging comes out aligned on the console, the following indentation style is used:

image

This is somewhat not standard as the `(` is aligned on indentation multiples (4 spaces), so wanted to call it out in case people didn't want it in

src/Test.sol
mds1: Ahh, yes missed that. We probably should add the `int256` versions of `assertApproxEqRel` also then, for consistency

src/Test.sol
mds1: Awesome, thanks 👌

src/Test.sol
mds1: > only hesitation would be if we wanted to properly overload `assertApproxEqRel` and `assertApproxEqAbs`

By "properly overload" do you mean with types other than int256/uint256? If so, I think just supporting int256/uint256 would cover most cases and is sufficient.

Either way, both a single Assertions section or two sections (one for assertEqs, one for approx equals) is good with me. I don't feel strongly about this, but see the thumbs up from @ZeroEkkusu so will leave it up to you both 😛

src/Test.sol
mds1: New ones lgtm, though I think the old one is still there too—there's two functions with the sig `function assertEq(bool a, bool b, string memory err)`

mds1:
> To help ensure the error messaging comes out aligned on the console, the following indentation style is used:
>
>  This is somewhat not standard as the ( is aligned on indentation multiples (4 spaces), so wanted to call it out in case people didn't want it in

Ah I see, this is fine with me, will leave it to anyone else to raise objections! 😅

src/Test.sol
OliverNChalk: Face palm, fixed. Can resolve if happy

src/Test.sol
OliverNChalk: It ends up being 4 sections if you group overloaded functions together. This may be excessive and the argument for would be if we expect more overloads to come down the track - then this makes it blindingly obvious where to put them and whether they comply with existing style/format.

Will defer to @ZeroEkkusu to act as tie breaker and move this forward. For reference here is the commit where I group overloaded assertions together: 924456ad2a19fe4e504639988a856c58244079d2

src/Test.sol
OliverNChalk: Done in 2c2337492baa475fa04bd3a539541d9d778a9c59

src/Test.sol
ZeroEkkusu: The idea I had was to use the big headers to divide the main parts of the contract:
"STD-CHEATS", "STD-ASSERTIONS", "STD-ERRORS", "STD-STORAGE".

I like the order as is, `assertEq` --> `assertApproxEqAbs`... It's logical IMO. Just put `assertFalse` first. I don't think we need any dividers rn; we can always add them later, as that'd be a non-breaking change.

ZeroEkkusu:
If you include `bound` in this PR, add it as a std-cheat below `deployCode`. ^^

src/Test.sol
OliverNChalk: Okay, summary of changes:

- Collapse assertion code regions into one: 99a36d8f522b5b5db867d8633fb8ed887884f6de
- Re-order assertFalse: 3c15901e223e9c2b311d1c67c64c7d2e40e8615d
- Add STD-CHEATS code region: 113b11cc97dbe416db99b1e8dfdc8144a34126d2

Changes not done:

- Did not add STD-ERRORS or STD-STORAGE as these are standalone libraries, not regions within `contract Test`

OliverNChalk:
> If you include `bound` in this PR, add it as a std-cheat below `deployCode`. ^^

Have opted to leave this out as it feels most appropriate in a separate PR. I'll throw that up post merge if no one beats me to the punchb

src/Test.sol
```diff
@@ -139,6 +143,202 @@ abstract contract Test is DSTest {
...
+        if (percentDelta > maxPercentDelta) {
+            emit log_named_string      ("Error", err);
+            emit log_named_int         ("    Expected", b);
+            emit log_named_int         ("      Actual", a);
+            emit log_named_decimal_int (" Max % Delta", maxPercentDelta, 18);
+            emit log_named_decimal_int ("     % Delta", percentDelta, 18);
+            fail();
+        }
+    }
 }
```
ZeroEkkusu, line 344: We should add a header here before `stdError` "STD-ERRORS" to separate them from the assertions.

src/Test.sol
```diff
@@ -159,7 +359,7 @@ library stdError {
 struct StdStorage {
```
ZeroEkkusu, line 359: Same here "STD-STORAGE"

ZeroEkkusu (commented):
Added a few comments. Otherwise, looks good!

src/Test.sol
OliverNChalk: done, can resolve if happy

src/Test.sol
OliverNChalk: done, can resolve if happy

PR #41, opened 2022-04-19
docs: StdStorage Readme Examples Broken

abigger87:
## Overview

There are multiple broken parts of the `StdStorage` README samples.

Implemented and tested changes here: https://github.com/abigger87/stdtests

Example:
Screenshot from 2022-04-18 21-53-47


ZeroEkkusu:
good catch ^^

PR #42, opened 2022-04-19
:gear: Additional Standard Cheats

abigger87:
## Overview

Introduces new Standard Cheats to abstract redundant operations I find myself doing too frequently with foundry testing.

## Checlist

- [x] `reverie`
- [x] `emission`
- [x] `key`

### DIFF
--- a/src/Test.sol
+++ b/src/Test.sol
@@ -23,6 +23,16 @@ abstract contract Test is DSTest {
         vm.warp(block.timestamp - time);
     }
 
+    // Abstract redundant signature encoding
+    function reverie(string memory sig) public {
+        vm.expectRevert(abi.encodeWithSignature(sig));
+    }
+
+    // Expect a full event
+    function emission() public {
+        vm.expectEmit(true, true, true, true);
+    }
+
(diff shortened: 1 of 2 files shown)


brockelmore:
appreciate the effort here, but closing. Two one-line helpers are just kinda bloat (see about adding them to template though?). The pvk helper should likely be upstreamed into rust by taking modulo of priv keys that are above curve order (plz open issue or PR into forge for that)

cc & h/t @mds1

abigger87:
Disagree since `Vm.sol` `addr` function accepts a `uint256` , this seems like a clear footgun.

Screenshot from 2022-04-19 13-29-56

brockelmore:
right the point is that we would allow the full range but for pvks > curve order, forge would modulo down to a valid pvk

abigger87:
sure

mds1:
yea I think we should update the `vm.addr` cheatcode so it doesn't support invalid behavior:
- revert if passed zero
- modulo values above curve order (or revert here too, but then we need a good error message and exposing internals of cryptography to users is prob confusing)

Issue #49, opened 2022-04-22
Add tests for `assertApproxEqRel`, `assertApproxEqAbs`


ZeroEkkusu:
See `assertApproxEqRel` and `assertApproxEqAbs` functions in [`Test.sol`](https://github.com/foundry-rs/forge-std/blob/master/src/Test.sol).

Create a new file in `src/test/`.

OliverNChalk:
Taking this one

PR #51, opened 2022-04-25
refactor(test): assertApproxEq{Rel,Abs} unit tests, improvements. & fixes

OliverNChalk:
### Brief

This PR adds unit tests for the recently added methods:

- assertFalse(bool)
- assertEq(bool,bool)
- assertEq(bytes,bytes)
- assertApproxEqAbs(uint256,uint256,uint256)
- assertApproxEqAbs(int256,int256,uint256)
- assertApproxEqRel(uint256,uint256,uint256)
- assertApproxEqRel(int256,int256,uint256)

In the quest to unit test these components, the following changes were also made:

- Introduced and tested `deltaMaths` library (used in relative and absolute approximation)
- Aligned visibility of asserts on `internal virtual` (2 funcs didn't have `virtual`)
- BREAKING: `assertApproxEq{Abs,Rel}(int256,int265,int256)` -> `assertApproxEq{Abs,Rel}(int256,int265,uint256)`, this is because a delta cannot be negative (i think :P)

### Next Steps

Hopefully there isn't anything too controversial in here, so should just be a matter of code reviewing & merging once happy. @ZeroEkkusu I did wanna call out that for the code blocks I use the following thought process (excerpt from a personal repo), let me know if you agree/disagree (saw you re-align some of the previous blocks):

```solidity
    // NB: code blocks extend from 4 to col 80, in-line with markdown standards
    // NB: code block name starts at 37:
    //
    //     length = (80 - 4)
    //     offset = 4
    //     text_len = 12
    //
    //     start_index = offset + length / 2 - text_len / 2
    //     start_index = 36
    //
    //     due to how editors count columns (starting at 1), this means the
    //     first character will be placed at col 37, i.e. col 36 will still be
    //     whitespace.
    //
    //     if your title is less than 12 chars, it will be left of center
    //     if your title is more than 12 chars, it will be right of center
    //     this is fine, do not attempt to center your title, just start typing
    //     at col 37 and beyond

    /*//////////////////////////////////////////////////////////////////////////
                                    CONSTANTS
```

### Resolves

Closes: #49


OliverNChalk:
CC: @ZeroEkkusu @mds1

src/Test.sol
```diff
@@ -159,7 +159,7 @@ abstract contract Test is DSTest {
         assertTrue(!data, err);
     }
 
-    function assertEq(bool a, bool b) internal {
+    function assertEq(bool a, bool b) internal virtual {
```
mds1, line 162: Agree with keeping the `assertApproxEq` methods virtual in case users want different approximation logic. The `assertEq` probably don't need to be virtual, and existing ones in DSTest's `test.sol` are also not virtual, so we should probably follow that convention for now

src/Test.sol
```diff
@@ -580,3 +580,49 @@ library stdStorage {
         return result;
     }
 }
+
+/*//////////////////////////////////////////////////////////////////////////
+                                DELTA MATHS
+//////////////////////////////////////////////////////////////////////////*/
+
+library deltaMaths {
```
mds1, line 588: About to have some nits on naming/style, probably should make sure @ZeroEkkusu is ok with these suggestions before implementing.

First is I'd suggest renaming this library to `mathHelpers` to be a bit more generic, as maybe some other utils will get added here down the road (e.g. I can see things like `max`, `min`, etc. getting added here).

src/Test.sol
```diff
@@ -580,3 +580,49 @@ library stdStorage {
...
+    function abs(int256 a) internal pure returns (uint256) {
+        if (a < 0) {
+            // lifted from: @OpenZeppelin/utils/math/SignedMath.sol
+            unchecked {
+                return uint256(-a);
+            }
+        }
+
+        return uint256(a);
```
mds1, line 597: I think the style OZ uses is a bit more readable due to its brevity:
```suggestion
         unchecked {
             // must be unchecked in order to support `n = type(int256).min`
            return uint256(n >= 0 ? n : -n);
        }
```

src/Test.sol
```diff
@@ -580,3 +580,49 @@ library stdStorage {
...
+    }
+
+    function getDelta(uint256 a, uint256 b) internal pure returns (uint256) {
```
mds1, line 600: Consistent with conventions like naming functions `max` instead of `getMax`
```suggestion
    function delta(uint256 a, uint256 b) internal pure returns (uint256) {
```

src/Test.sol
```diff
@@ -580,3 +580,49 @@ library stdStorage {
...
+        // a and b are of the same sign
+        if (a >= 0 && b >= 0 || a < 0 && b < 0) {
+            return getDelta(abs(a), abs(b));
+        }
+
+        // a and b are of opposite signs
+        return abs(a) + abs(b);
+    }
+
+    function getPercentDelta(uint256 a, uint256 b) internal pure returns (uint256) {
```
mds1, line 616: Similar comment on naming
```suggestion
    function percentDelta(uint256 a, uint256 b) internal pure returns (uint256) {
```

src/test/StdAssertions.t.sol
```diff
@@ -0,0 +1,368 @@
+// SPDX-License-Identifier: Unlicense
+pragma solidity 0.8.10;
+
+import "../Test.sol";
+
+contract StdAssertionsTest is Test
+{
+    // todo: figure out how to expectEmit multiple events
```
mds1, line 8: I think the `assertEq` overloads are simple enough that we don't really need tests for them as they may add more clutter than value—even the original ones don't have tests 😅

But you can stack expect emits like this:

```solidity
vm.expectEmit(true, true, true, true);
emit MyFirstEvent(val1);
vm.expectEmit(true, true, true, true);
emit MySecondEvent(val2);

// This will look for both events in the below call
myContract.myCall();
```

mds1 (commented):
Thanks a lot, looks great! Most comments are nits on style, and one where I realized we're checking relative tolerance incorrectly

> BREAKING: assertApproxEq{Abs,Rel}(int256,int265,int256) -> assertApproxEq{Abs,Rel}(int256,int265,uint256), this is because a delta cannot be negative (i think :P)

Yes good call here, `uint256` is the right choice 👌

src/Test.sol
```diff
@@ -580,3 +580,49 @@ library stdStorage {
...
+        uint256 delta = getDelta(a, b);
+
+        return delta * 1e18 / b;
```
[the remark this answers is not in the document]
mds1: Worth noting there are exceptions here: If a value starts at `a`, then at some point changes to `b`, you'd calculate the percent change with `a` in the denominator.

But I suspect most uses cases of this method will be "given `a` and `b`, what is the percent error between them", in which case the denominator should be the smaller of the two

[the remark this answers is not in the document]
transmissions11: interesting, frankie wanted the behavior of delta between a and b, not min()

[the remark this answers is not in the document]
OliverNChalk: +1 for assuming expected to be the source of truth and the percent delta to measured against that

src/Test.sol
OliverNChalk: Done, can resolve if happy

src/Test.sol
OliverNChalk: Done, can resolve if happy

src/test/StdAssertions.t.sol
OliverNChalk: Thanks for the expectEmit answer! Re testing assertEq im flexible, it was quite easy to just template all these tests and fill them in (some of it was even macro-able). If you think their value as regression guards is outweighed by the clutter (given it would be impressive to regress assertEq()), then okay for them to be dropped

mds1: Hmm, paging @ZeroEkkusu for a second opinion here since I don't feel too strongly

src/Test.sol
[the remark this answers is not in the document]
mds1: Sounds good, marking this resolved then!

ZeroEkkusu (commented):
Hey @OliverNChalk, great additions! I think Forge Std should provide things that are often needed when testing but not natively available in Solidity. I'd do these manually, so a math lib is great.

Agree with Matt on following the DSTest conventions for now.
That being said, we can also reduce the amount of code if we do assertions overriding like [this](https://github.com/dapphub/ds-test/blob/2c7dbcc8586b33f358e3307a443e524490c17666/src/test.sol#L157-L170). Let me know what you think.

The test are pretty comprehensive, love it.

I'll leave more comments in conversations.

Edit: I don't mind the code block style! Didn't know it was intentional.

src/Test.sol
ZeroEkkusu: `StdMath` is prettier than `MathHelpers`.

```solidity
a = MathHelpers.abs(x);

a = StdMath.abs(x);
```

In either case, let's pick a good, descriptive name now so we don't have to break users' code later :)

mds1: Oh good call I like `StdMath`, consistent with the other parts of forge-std too

OliverNChalk: Done, can resolve if happy

src/test/StdAssertions.t.sol
OliverNChalk: DSTest's convention being to not have a corresponding test file for any assertions?

OliverNChalk:
> That being said, we can also reduce the amount of code if we do assertions overriding like [this](https://github.com/dapphub/ds-test/blob/2c7dbcc8586b33f358e3307a443e524490c17666/src/test.sol#L157-L170). Let me know what you think.

Do we want to follow DSTest convention here? I feel like we should, I can call the inner `assertEq()` from `assertEq(string memory err)`

src/Test.sol
```diff
@@ -585,16 +565,12 @@ library stdStorage {
                                 DELTA MATHS
```
ZeroEkkusu, line 565: `STD-MATH`

ZeroEkkusu (commented):
@OliverNChalk, I'll leave other comments in the convos.

src/Test.sol
OliverNChalk: Done

ZeroEkkusu:
Oops looks like there's a bug in my GH UI and the comments were not posted.

Regarding the assertions test, I wrote:
"I meant the simple ones, `assertFalse` and `assertEq`. But totally don't mind if you leave it as is (for more coverage 😀)."

If you don't feel to strongly about either, let's leave them and cover everything in Forge Std from now on. It's the standard testing library for Foundry, after all.

ZeroEkkusu:
Otherwise, lgtm; @mds1 can take a final look and merge once ready.

mds1:
LGTM, thanks a lot @OliverNChalk!

src/Test.sol
```diff
@@ -580,3 +560,45 @@ library stdStorage {
...
     }
 }
+
+/*//////////////////////////////////////////////////////////////////////////
+                                STD-MATH
+//////////////////////////////////////////////////////////////////////////*/
+
+library stdMath {
+    function abs(int256 a) internal pure returns (uint256) {
+         unchecked {
```
ernestognw, line 570: Just letting you guys know: https://github.com/foundry-rs/forge-std/issues/53

PR #68, opened 2022-05-10
feat(storage): add `read` functions

### LINKED ISSUE #65: feature: syntactic sugar for reading storage variables
Currently, in order to access a value from storage in a test you need to:

1. get its storage slot, which is a `uint256`
2. convert the slot variable to `bytes32`
3. load the data from the address' storage
4. convert back to the variable's type

```
// get slot
uint256 varSlot = stdstore
            .target(address(contractName))
            .sig(contractName.storageVar.selector)
            .depth(0)
            .find();
// convert to bytes32 and load from slot, then convert to desired variable type (uint256 in this case)
uint256 storageVar = uint256(vm.load(address(collatz), bytes32(delay_slot)));
```

Would be great to have some syntactic sugar to make this process more straightforward as well as hopefully getting rid of the back and forth with type conversions. Perhaps `...find().load()` or something along those lines?

ZeroEkkusu:
## Motivation

Resolve #65 

## Solution

Add the following terminator functions to `StdStorage`:

- `read_bytes32`
- `read_bool`
- `read_address`
- `read_uint`
- `read_int`

The naming style follows the DSTest convention (`_<type>`, such as `log_address`).

### Example

```solidity
// BEFORE

uint256 slot = stdstore
    .target(address(c))
    .sig(c.a.selector)
    .find();

bytes32 val = vm.load(address(c), bytes32(slot));

address a = address(uint160(uint256(val)));


// NOW

address a = stdstore
    .target(address(c))
    .sig(c.a.selector)
    .read_address();
```

Cc @itamarreif
And also let's discuss whether this is the right solution.

### DIFF
--- a/src/Test.sol
+++ b/src/Test.sol
@@ -564,32 +564,31 @@ library stdStorage {
         delete self._depth;
     }
 
-    function read_bytes32(StdStorage storage self) internal returns (bytes32) {
+    function read(StdStorage storage self) private returns (bytes memory) {
         address t = self._target;
         uint256 s = find(self);
-        return vm_std_store.load(t, bytes32(s));
+        return abi.encode(vm_std_store.load(t, bytes32(s)));
+    }
+
+    function read_bytes32(StdStorage storage self) internal returns (bytes32) {
+        return abi.decode(read(self), (bytes32));
     }
 
 
     function read_bool(StdStorage storage self) internal returns (bool) {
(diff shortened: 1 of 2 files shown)


brockelmore (approved):
awesome, lgtm :)

gakonst (approved):
love it

Issue #13, opened 2022-02-05
CapWords naming sytle


ZeroEkkusu:
Is there a reason why the names of the contracts and libraries don't begin with a capital letter?

E.g. `stdCheats` --> `StdCheats`

ZeroEkkusu:
Whoa, this issue is old; I barely knew how to use Foundry back then.

But, yeah, I'm still for this.

gakonst:
Closing as #wontfix per conversation [here](https://github.com/foundry-rs/forge-std/pull/69#issuecomment-1129408485). LMK if anyone disagrees / feels strongly about re-opening.

mds1:
Reopening this as something we may implement as part of breaking changes as discussed in https://github.com/foundry-rs/forge-std/pull/126#discussion_r924699853

PR #69, opened 2022-05-13
fix: CapWords naming style

shawnharmsen:
issue: https://github.com/foundry-rs/forge-std/issues/13

Not sure if the issue is still relevant haha

Regarding test
- I saw that ci does forge test -vvv
https://github.com/foundry-rs/forge-std/blob/37a3fe48c3a4d8239cda93445f0b5e76b1507436/.github/workflows/tests.yml#L20-L21

Are there other things I should be doing for testing?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 cache/
 out/
+.idea
--- a/README.md
+++ b/README.md
@@ -222,25 +222,25 @@ Expand upon the assertion functions from the `DSTest` library.
 ### `console.log`
 
 Usage follows the same format as [Hardhat](https://hardhat.org/hardhat-network/reference/#console-log).
-It's recommended to use `console2.sol` as shown below, as this will show the decoded logs in Forge traces.
+It's recommended to use `Console2.sol` as shown below, as this will show the decoded logs in Forge traces.
 
 ```solidity
 // import it indirectly via Test.sol
 import "forge-std/Test.sol";
 // or directly import it
-import "forge-std/console2.sol";
+import "forge-std/Console2.sol";
 ...
 console2.log(someValue);
 ```
 
-If you need compatibility with Hardhat, you must use the standard `console.sol` instead.
-Due to a bug in `console.sol`, logs that use `uint256` or `int256` types will not be properly decoded in Forge traces.
+If you need compatibility with Hardhat, you must use the standard `Console.sol` instead.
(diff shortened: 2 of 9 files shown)


ZeroEkkusu (commented):
Thanks!

This is a breaking change so let's see how we should approach it. Cc @mds1.

> Are there other things I should be doing for testing?

`forge test -vvv` was enough in this case. But when you change code or add new features in `Test`, you also want to check compatibility with older solc versions, as done in the CI.

### Edit

We should rename `stdStorage` (the library) to `StdStorage` as well.

Thoughts on renaming `StdStorage` (the struct) to `StdStore` then?

That way it would match with the name of the variable. We'd then use the library like so:

```solidity
using StdStorage for StdStore;

function f() public {
    stdstore.target(address(c)). //...
```

We can also rename `stdstore` to `stdStore`.

mds1:
We should probably sync this with the next planned foundry breaking change. @gakonst @onbjerg is there anywhere that tracks a list of breaking changes planned for a future release? Wondering the best way logistically to keep these in sync

gakonst:
This large naming change seems unnecessary for now..the `Console.sol` change I'm against because it means I'd need to do `Console.log("foo")` whereas `console.log` is nicer. 

I'd vote to close #13 as #wontfix given that this is a stylistic change on a testing lib which doesn't seem to be exposed to the user in many ways?

ZeroEkkusu:
#13 is a relatively old issue (`console` wasn't even part of Forge Std).

As I said recently, I still like the idea of following the CapWords convention, but don't like breaking things.

Best to decide now, and if we're not going to do it, close as wontfix as Georios proposed.

mds1:
Agree that `console` should stay lowercase (sorry, hadn't looked at the diff yet). 

IIUC the reason this is a breaking change is because users need to access e.g. `stdError.arithmeticError` and `using stdStorage for StdStorage`? It'd be nice if that was `StdError` and `StdStorage`, because that's the convention used pretty much everywhere else, but obviously this isn't a huge deal so I'm fine leaving as-is. Plus then we also have to rename the struct anyway to avoid a name conflict. So I think I've talked myself into agreeing we close this PR + #13 and leave as-is for now

gakonst:
@shawnharmsen thanks for the contribution, I think we'll close this for now given the arguments brought up above. LMK if you think this is unreasonable and we can discuss ways we can incorporate in future releases while being sensitive of breakages!

PR #72, opened 2022-05-15
feat(tests): add check for successful `deployCode`.

### LINKED ISSUE #35: feat: check for successful deploy in `deployCode`
`create` (and `create2`, though it's not used) return the zero address if a deploy fails. The `deployCode` helpers should have a check such as:

```solidity
require(addr != address(0), "Deployment failed. Ensure the artifact path and file contents are correct.")
```

dsam82:
# Motivation

Resolves #35

Added checks for `zero` address reverts.

cc: @mds1

### DIFF
--- a/src/Test.sol
+++ b/src/Test.sol
@@ -128,7 +128,7 @@ abstract contract Test is DSTest {
     }
 
     function bound(uint256 x, uint256 min, uint256 max) public returns (uint256 result) {
-        require(min <= max, "StdCheats bound(uint256,uint256,uint256): MAX_LESS_THAN_MIN");
+        require(min <= max, "Test bound(uint256,uint256,uint256): Max is less than min.");
 
         uint256 size = max - min;
 
@@ -165,7 +165,7 @@ abstract contract Test is DSTest {
 
         require(
             addr != address(0),
-            "StdCheats deployCode(string memory,bytes memory): Deployment failed"
+            "Test deployCode(string,bytes): Deployment failed."
         );
     }
 
@@ -181,7 +181,7 @@ abstract contract Test is DSTest {
 
         require(
             addr != address(0),
-            "StdCheats deployCode(string memory): Deployment failed"
+            "Test deployCode(string): Deployment failed."
         );
     }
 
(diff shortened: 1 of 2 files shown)


src/test/StdCheats.t.sol
```diff
@@ -138,6 +138,11 @@ contract StdCheatsTest is Test {
         assertEq(string(getCode(deployed)), string(getCode(address(this))));
     }
 
+    function testDeployCodeFail() public {
+        vm.expectRevert(bytes("StdCheats deployCode(string memory): Deployment failed"));
```
dsam82, line 142: @ZeroEkkusu this `expectRevert` is not working as expected, even though error is being thrown.

ZeroEkkusu: This is because when `expectRevert` is used before `deployCode`, the value will not get assigned to `addr` in `deployCode`, regardless of the success:

https://github.com/foundry-rs/forge-std/blob/4d36e3f7e2168c8155c641eb0f80e85cd584bd1c/src/Test.sol#L163

Would you please take a look at this, @mds1?

src/Test.sol
```diff
@@ -468,10 +478,10 @@ library stdStorage {
                 vm_std_store.store(who, reads[i], prev);
             }
         } else {
-            require(false, "No storage use detected for target");
+            require(false, "StdCheats find(StdStorage storage): No storage use detected for target");
```
ZeroEkkusu, line 481: The name of this library is `stdStorage`, so let's refer to it that way.
```suggestion
            require(false, "stdStorage find(StdStorage): No storage use detected for target.");
```

src/Test.sol
```diff
@@ -468,10 +478,10 @@ library stdStorage {
...
         }
 
-        require(self.finds[who][fsig][keccak256(abi.encodePacked(ins, field_depth))], "NotFound");
+        require(self.finds[who][fsig][keccak256(abi.encodePacked(ins, field_depth))], "StdCheats find(StdStorage storage): NotFound");
```
ZeroEkkusu, line 484: Same as above.

```suggestion
        require(self.finds[who][fsig][keccak256(abi.encodePacked(ins, field_depth))], "stdStorage find(StdStorage): Slot(s) not found.");
```

ZeroEkkusu (commented):
Small nits. There are also two more require statements in `stdStorage`, for when the slots are packed.

src/test/StdCheats.t.sol
mds1: Sorry not sure I'm following—don't we want `addr` to be the zero address, so the require statement is hit and it reverts?

Or is the issue that expectRevert isn't working here because we're not actually making a call? If so, you can try making it an external call with `this.deployCode` or using a try/catch and verifying the revert reason

ZeroEkkusu: Thanks @mds1, `this.deployCode` solves the issue.

@dsam82, don't forget to update the `bound` tests to use the new messages too!

dsam82: Wow! this works. Can you make my dumbass understand why? 😅

gakonst (approved):

src/test/StdCheats.t.sol
gakonst: `this.deployCode` works because it makes a `CALL`, whereas `deployCode` is making a JUMP which `expectRevert` does not hook into

PR #83, opened 2022-06-01
feat(test.sol): hint for `setUp` function

wiasliaw:
A hint for `setUp` function signature. The only break change is to add `override` in `setUp` function.

image

```sol
contract T is Test {
  function setUp() public override {
    ...
  }
}
```


ZeroEkkusu:
I think you are looking for snippets; check out [Forge Snippets](https://github.com/crisgarner/VSCodeForgeSnippets#setup). `setUp` is supported.

Let me know if that was helpful.

Issue #78, opened 2022-05-23
refactor: add an abstract `Script` contract


mds1:
With solidity scripting and deploys on the way (https://github.com/foundry-rs/foundry/pull/1208), a a series of new cheatcodes for sending transactions will be added to the `Vm` interface. To access these, you'd currently need to have your scripts inherit from `Test` which is a bit awkward because (1) this is not a test file, and (2) even if you just renamed `Test`, you still inherit lots of cheatcodes and forge-std methods which can be dangerous for live transactions, and therefore are not desirable to have present in a script. 

As a result, I think we should refactor to have `abstract contract Script` and `abstract contract Test is Script`, this way scripts can just inherit from `Script`, and tests can continue to inherit from `Test` with no breaking changes. 

This abstract `Script` contract will contain:
- A minimal `Vm` interface with just the new `broadcast` cheatcodes, as well as "view" cheatcodes that don't change state. I may be missing some, but I think this includes: `load`, `addr`, `getNonce`, `records`, `accesses`, `getCode`, and `label`.
- The `stdMath` library.
- The `stdStorage` library, but with the `checked_write` methods disabled.

Libraries don't get inherited by default, so we may need to split the last one into `stdStorageRead` and `stdStorage is stdStorageRead` (assuming libraries can inherit? I'm not actually sure). Alternatively perhaps we make it into a contract and use a storage var to conditionally disable `checked_write` based on that, though this would be a breaking change.

PR #85, opened 2022-06-09
✨ Add assertEq for uint and int arrays

paulrberg:
I didn't add an overloaded function with an `err` argument because I couldn't decide where to emit the custom error string: in the array length comparison check, in the per-element comparison check, or as a third argument to the `assertEq` function.

### DIFF
--- a/src/Test.sol
+++ b/src/Test.sol
@@ -241,20 +241,16 @@ abstract contract Test is DSTest, Script {
 
     function assertEq(int256[] memory a, int256[] memory b) internal {
         if (a.length != b.length) {
-            emit log("Error: a.length == b.length not satisifed");
+            emit log("Error: a.length == b.length not satisfied");
             fail();
-        }
-
-        for (uint256 i = 0; i < a.length;) {
-            if (a[i] != b[i]) {
-                emit log("Error: a[] == b[] not satisifed");
-                assertEq(a[i], b[i]);
-            }
-            unchecked {
-                i += 1;
+        } else {
+            for (uint256 i = 0; i < a.length; ++i) {
+                if (a[i] != b[i]) {
+                    emit log("Error: a[] == b[] not satisfied");
+                    assertEq(a[i], b[i]);
+                }
             }
         }
-    }
 
     function assertApproxEqAbs(
         uint256 a,
--- a/src/Test.sol
+++ b/src/Test.sol
@@ -253,6 +253,32 @@ abstract contract Test is DSTest, Script {
         }
     }
 
+    function assertEq(uint256[] memory a, uint256[] memory b, string memory err) internal {
+        if (a.length != b.length) {
+            emit log_named_string("Error", err);
+        } else {
+            for (uint256 i = 0; i < a.length; ++i) {
+                if (a[i] != b[i]) {
+                    emit log_named_string("Error", err);
+                }
+            }
+        }
+        assertEq(a, b);
+    }
+
+    function assertEq(int256[] memory a, int256[] memory b, string memory err) internal {
+        if (a.length != b.length) {
+            emit log_named_string("Error", err);
+        } else {
+            for (uint256 i = 0; i < a.length; ++i) {
+                if (a[i] != b[i]) {
+                    emit log_named_string("Error", err);
+                }
+            }
+        }
+        assertEq(a, b);
+    }
+
     function assertApproxEqAbs(
         uint256 a,
         uint256 b,
--- a/src/Test.sol
+++ b/src/Test.sol
@@ -256,27 +256,29 @@ abstract contract Test is DSTest, Script {
     function assertEq(uint256[] memory a, uint256[] memory b, string memory err) internal {
         if (a.length != b.length) {
             emit log_named_string("Error", err);
+            assertEq(a, b);
         } else {
             for (uint256 i = 0; i < a.length; ++i) {
                 if (a[i] != b[i]) {
                     emit log_named_string("Error", err);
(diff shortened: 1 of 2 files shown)


ZeroEkkusu (changes requested):
This is a good!

> I couldn't decide where to emit the custom error string

Generally, dev's logs are emitted before the standard logs.

<br>

We need to fix a couple of things.

ZeroEkkusu:
Additionally, we should include tests. I already wrote some to review the PR, so feel free to use them!

```solidity
    /*//////////////////////////////////////////////////////////////////////////
                                    ASSERT_EQ(ARRAY)

    function testAssertEq_UintArr_Pass(uint256 e0, uint256 e1, uint256 e2) public {
        uint256[] memory a = new uint256[](3);
        a[0] = e0;
        a[1] = e1;
        a[2] = e2;
        uint256[] memory b = new uint256[](3);
        b[0] = e0;
        b[1] = e1;
        b[2] = e2;

        t._assertEq(a, b, EXPECT_PASS);
    }

    function testAssertEq_IntArr_Pass(int256 e0, int256 e1, int256 e2) public {
        int256[] memory a = new int256[](3);
        a[0] = e0;
        a[1] = e1;
        a[2] = e2;
        int256[] memory b = new int256[](3);
        b[0] = e0;
        b[1] = e1;
        b[2] = e2;

        t._assertEq(a, b, EXPECT_PASS);
    }

    function testAssertEq_UintArr_FailEl(uint256 e1) public {
        vm.assume(e1 != 0);
        uint256[] memory a = new uint256[](3);
        uint256[] memory b = new uint256[](3);
        b[1] = e1;

        vm.expectEmit(false, false, false, true);
        emit log("Error: a[] == b[] not satisifed");
        t._assertEq(a, b, EXPECT_FAIL);
    }

    function testAssertEq_IntArr_FailEl(int256 e1) public {
        vm.assume(e1 != 0);
        int256[] memory a = new int256[](3);
        int256[] memory b = new int256[](3);
        b[1] = e1;

        vm.expectEmit(false, false, false, true);
        emit log("Error: a[] == b[] not satisifed");
        t._assertEq(a, b, EXPECT_FAIL);
    }

    function testAssertEq_UintArr_FailLen(uint256 lenA, uint256 lenB) public {
        vm.assume(lenA != lenB);
        vm.assume(lenA <= 10000);
        vm.assume(lenB <= 10000);
        uint256[] memory a = new uint256[](lenA);
        uint256[] memory b = new uint256[](lenB);


        vm.expectEmit(false, false, false, true);
        emit log("Error: a.length == b.length not satisifed");
        t._assertEq(a, b, EXPECT_FAIL);
    }

    function testAssertEq_IntArr_FailLen(uint256 lenA, uint256 lenB) public {
        vm.assume(lenA != lenB);
        vm.assume(lenA <= 10000);
        vm.assume(lenB <= 10000);
        int256[] memory a = new int256[](lenA);
        int256[] memory b = new int256[](lenB);


        vm.expectEmit(false, false, false, true);
        emit log("Error: a.length == b.length not satisifed");
        t._assertEq(a, b, EXPECT_FAIL);
    }

```

These two go inside the `TestTest` contract, which is at the end of the file:

```solidity
    function _assertEq(uint256[] memory a, uint256[] memory b, bool expectFail) external expectFailure(expectFail) {
        assertEq(a, b);
    }    

    function _assertEq(int256[] memory a, int256[] memory b, bool expectFail) external expectFailure(expectFail) {
        assertEq(a, b);
    }    

```

paulrberg:
Oh hey @ZeroEkkusu thanks for your feedback! And sorry, I should have written the tests myself.

> The oldest supported Solidity version is `0.6.0`, so we can't use `unchecked`.

Any plans to bump this to v0.8? That would encourage devs to upgrade. Anyone still using v0.6 and v0.7 could just use the older versions of `forge-std`.

ZeroEkkusu:
I like your thinking direction, but think we still need to support projects that use those. Although that would also allows us to implement some things easily (#78).

Interested to hear Matt's thoughts on this anyway.

ZeroEkkusu (approved):
Alright! I added custom errors as well.

Thank you for contributing!

src/Test.sol
```diff
@@ -225,6 +225,62 @@ abstract contract Test is DSTest, Script {
         assertEq0(a, b, err);
     }
 
+    function assertEq(uint256[] memory a, uint256[] memory b) internal {
```
mds1, line 228: This PRis already merged, but curious what you think of just verifying array equality like this instead? Emitting the full arrays on failure seems like a nice way to report mismatched arrays (at least it was helpful for me when writing that lib), and the code is simpler: https://github.com/mds1/solidity-generators/blob/dcb3eeeabbc8c6fb6684617fe2344a8442f8f17e/test/Generators.t.sol#L27-L34

mds1: Won't this approach also result in duplicate logs? First you'll get `emit log("Error: a[] == b[] not satisfied");` emitted, then you'll also get `emit log("Error: a == b not satisfied [address]");` emitted

ZeroEkkusu: Yeah, I was aware of how the logs worked.

Your solution is cleaner and we should indeed go with the expected/actual approach. I overlooked it.

That being said, it is only a matter of refactoring the code and changing the logs, no breaking changes. Do you want to reimplement it @mds1?

paulrberg: I also agree that @mds1's solution is more elegant. I didn't even think about using `keccak256`!

PR #86, opened 2022-06-10
♻️ Simplify assertEq for arrays

mds1:
@ZeroEkkusu likely won't be able to get to the tests today so if wanted to fix those that would be appreciated!

cc @paulrberg

### DIFF
--- a/src/Test.sol
+++ b/src/Test.sol
@@ -249,12 +249,14 @@ abstract contract Test is DSTest, Script {
 
     function assertEq(uint256[] memory a, uint256[] memory b, string memory err) internal {
         if (keccak256(abi.encode(a)) != keccak256(abi.encode(b))) {
+            emit log_named_string("Error", err);
             assertEq(a, b);
         }
     }
 
     function assertEq(int256[] memory a, int256[] memory b, string memory err) internal {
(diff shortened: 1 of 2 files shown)


src/Test.sol
```diff
@@ -226,58 +230,32 @@ abstract contract Test is DSTest, Script {
     }
 
     function assertEq(uint256[] memory a, uint256[] memory b) internal {
-        if (a.length != b.length) {
-            emit log("Error: a.length == b.length not satisfied");
+        if (keccak256(abi.encode(a)) != keccak256(abi.encode(b))) {
+            emit log("Error: a == b not satisfied [uint[]]");
+            emit log_named_array("  Expected", b);
+            emit log_named_array("    Actual", a);
```
ZeroEkkusu, line 236: Forge doesn't support printing these, does it? @mds1 

They show up only in traces.

ZeroEkkusu (approved):
Ready to merge once we resolve the convo above.

src/Test.sol
mds1: Thanks for handling the tests!

Good point, it doesn't currently, but it shouldn't be too hard to support so we can create an issue upstream to track

paulrberg:
This is awesome guys! Just found the time to review the code changes, and it LGTM.

Issue #32, opened 2022-03-25
feature: support writing to packed slots


mds1:
For local development, I currently use uint256 everywhere so I can use the forge-std helpers to easily set slot values. However, eventually I want to start packing variables without these helpers and the tests breaking.

I think the easiest way to do this is to leverage the solc storage layout to get the info we need. From @brockelmore:

> ye so this actually gets interesting now that i have a dynamic mapping cuz we can actually represent json directly in solidity (kek). we could add a cheatcode to get the storage layout into memory then you could pass in the layout with relative ease for a slot and *know* youre doing it correctly. in the short term if this is something u want and can easily provide the positioning + left and right padding we can support it
>
> something like `.packed(left, right)`
> 
> where each is number of bits on the left and right
> 
> which eventually would ideally be grabbable in code via cheatcode

Since storage layout is not included by default, we either need to throw an error if the storage layout isn't present, or recompile the specified contract with that flag on the fly

For contracts that aren't local (e.g. if testing against a forked network), I think we'd be limited to only setting vars in packed slots if the contract is verified on Etherscan so we can pull the source and compile with storage layout enabled

japui-coder:
feature: support writing to packed slots #32

Yonkoo11:
Looks like this landed.

- `stdstore.enable_packed_slots()` is on main in `StdStorage.sol`
- landed via #148 (`feat(stdStorage): Allow writing to a packed storage slot`)
- tests in `StdStorage.t.sol` exercise packed writes
- #804 fixed related packed-slot docs

Unless something is still missing from the original ask (layout-from-solc / etherscan path), this can probably close.

PR #103, opened 2022-07-03
🐛 Storage: Fix finding bool

### LINKED ISSUE #90: bug(StdStorage): unable to find bool slot within map struct
Hey guys,

Seems like `stdStorage` unable to find `bool` slot within a map struct variable. I debug and found out that the `rdata` not return hex"1337" if `address.staticcall` to the bool slot. The whole bytes32 just become `0000000000000000000000000000000000000000000000000000000000000000` as you can see below:
telegram-cloud-photo-size-5-6145377720268337323-y

You can reproduce with this contract: https://github.com/spicysquid168/solidity-playground/blob/main/src/TryStdStorageBool.sol
Run by `forge script ./src/TryStdStorageBool.sol -vvv`

Thanks,

ZeroEkkusu:
## Motivation

Fix #90 

Edit: Needs more work for other types that are not 32 bytes long and are not packed in the struct.

### DIFF
--- a/src/Test.sol
+++ b/src/Test.sol
@@ -486,6 +486,14 @@ library stdStorage {
         uint256 field_depth = self._depth;
         bytes32[] memory ins = self._keys;
 
+        // Helpful message for some reverts
+        string memory tips = who == address(0) || fsig == bytes4(0) ? string(abi.encodePacked(
+            " Did you forget to set ",
+            who == address(0) ? "target" : "",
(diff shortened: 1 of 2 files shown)


ZeroEkkusu:
Waiting for `parseJson` (https://github.com/foundry-rs/forge-std/issues/32).

Theghost256 (approved):

PR #109, opened 2022-07-10
feat: add computeCreateAddress helper function in Script

devanonon:


transmissions11:
ik forge-std doesn't place a large emphasis on comments, but i think given the amount of magic constants in this code its worth using a commented impl like `LibRLP` https://github.com/Rari-Capital/solmate/blob/v7/src/utils/LibRLP.sol

src/Script.sol
```diff
@@ -11,4 +11,50 @@ abstract contract Script {
...
+                bytes1(0xda),
+                bytes1(0x94),
+                origin,
+                bytes1(0x84),
+                uint32(nonce)
+            );
+        bytes32 hash = keccak256(data);
+        assembly {
+            mstore(0, hash)
+            creation := mload(0)
```
brockelmore, line 57: why do you mstore then mload here? also memslot 0 is unstable so if you want it around, use address at free mem pointer and return it. but more likely feels like it should just be removed?

brockelmore (changes requested):

src/Script.sol
```diff
@@ -11,4 +11,50 @@ abstract contract Script {
         address(bytes20(uint160(uint256(keccak256('hevm cheat code')))));
 
     Vm public constant vm = Vm(VM_ADDRESS);
+
+    // Calcuate the corresponding contract creation address for a given address and nonce
+    function addressFrom(address origin, uint256 nonce) internal pure returns (address creation) {
```
mds1, line 16: Thoughts on renaming to something like `computeCreateAddress`, that way adding another `computeCreate2Address` fits nicely?

src/Script.sol
devanonon: I think it was to avoid all the casting necessary to go from bytes32 to address but obviously the casting is neater / more explicit. It's very old code and it's been replaced by the solmate impl in my latest commit anyway as recommended by @transmissions11.

devanonon:
> ik forge-std doesn't place a large emphasis on comments, but i think given the amount of magic constants in this code its worth using a commented impl like `LibRLP` https://github.com/Rari-Capital/solmate/blob/v7/src/utils/LibRLP.sol

Nice point, I've updated it to use an impl based on that one, thanks!

brockelmore (approved):
lgtm

devanonon:
@brockelmore would you prefer `addressFrom` or something else?

brockelmore:
`computeCreateAddress` re @mds1's suggestion

devanonon:
@brockelmore agreed, already changed to that in my last commit ser

brockelmore:
lmfao sorry, will merge

brockelmore:
just realized this broke solidity 0.6.x compiler version cuz CI was tightly configured to only run for certain contributors :(. @devanonon can you try to fix, sorry

devanonon:
@brockelmore Just realized that too and am on it, no worries :)

PR #111, opened 2022-07-10
fix: remove type(x).max in favor of powers of 2

devanonon:
Fix after failing CI in #109 due to older solc versions not allowing `type(x).max`.


brockelmore:
ty :)

PR #114, opened 2022-07-13
Optimize same sign check

paulrberg:
See [this](https://twitter.com/PaulRBerg/status/1546957951579062272).

I was also thinking to add a comment to explain why `(a^b)>-1` works - WDYT?


paulrberg:
Done

PR #120, opened 2022-07-14
Mark "bound" cheat as "public"

paulrberg:
Just like https://github.com/foundry-rs/forge-std/pull/119 but for the `bound` cheat.

### DIFF
--- a/src/Test.sol
+++ b/src/Test.sol
@@ -136,7 +136,7 @@ abstract contract Test is DSTest, Script {
         }
     }
 
-    function bound(uint256 x, uint256 min, uint256 max) internal virtual returns (uint256 result) {
+    function bound(uint256 x, uint256 min, uint256 max) public virtual returns (uint256 result) {
         require(min <= max, "Test bound(uint256,uint256,uint256): Max is less than min.");
 
         uint256 size = max - min;


ZeroEkkusu (commented):
We made it `internal virtual` here to resolve conflicts in this particular case: https://github.com/foundry-rs/forge-std/pull/92

paulrberg:
Interesting - I double-checked and indeed it looks like when two methods have different visibility it doesn't matter whether they both have the `virtual` modifier, the compiler will throw this error:

> TypeError: Overriding function visibility differs.

I tested using this code:

```solidity
pragma solidity >=0.8.15;

contract Foo {
    function method() public virtual view returns (uint256 time) {
        time = block.timestamp;
    }
}

contract Bar {
    function method() internal virtual view returns (uint256 time) {
        time = block.timestamp;
    }
}

contract Baz is Foo, Bar {
    function method() public override(Foo, Bar) view returns (uint256 time) {
        time = block.timestamp;
    }
}
```

There's something I don't like about keeping some cheats `internal` while most are `public` .. but I may expound on this in another issue at another time. Going to close this now, thanks for clarifying.

Issue #122, opened 2022-07-15
Meta: Unbundling the cheats from the assertions (leg 1)


paulrberg:
## Description

The way the [Test](https://github.com/foundry-rs/forge-std/blob/c19dfd2f2a88a461216b0dd1f4961e1a85dcad46/src/Test.sol) contract is currently architected is not in line with the principle of separation of concerns.

The most problematic thing IMHO is the fact that the cheats and assertions live under the same roof. This makes it difficult for Forge Std to be composed with other testing assertion contracts, such as [PRBTest](https://github.com/paulrberg/prb-test).

Of course, I'm a bit biased because of PRBTest, but even if we nevermind my intentions, there are many reasons for why people may wish to write their own testing assertion contracts. If anything, maybe they simply don't like either DSTest or PRBTest. Or perhaps they're working with [user-defined value types](https://blog.soliditylang.org/2021/09/27/user-defined-value-types/), which need certain overloads for `assertEq`. Or perhaps they're using a fork of the EVM on which DSTest doesn't compile or doesn't work well. Bottom line is: Forge Std is not as modular as it could be.

Additionally, separating the cheats from the assertions would also tidy up the contracts a bit, since there would be fewer lines of code in each contract.

## Solution

Create a `Cheats` contract and move all cheats there. `Test` would then inherit from `Cheats`.

Doing this would make Forge Std more modular. Users would be able to selectively use just the cheats or just the assertions.

paulrberg:
Closing since #126 has been merged on the `v0.3` branch.

Issue #124, opened 2022-07-15
Meta: Unbundling Test from Script


paulrberg:
## Description

Commit [f18682b2][1] made by @devanonon added two new utility functions in the `Script` contract:

1. `computeCreateAddress`
2. `addressFromLast20Bytes`

Which are of course very useful in the context of deploying contracts (the main feature of Solidity Scripting).

But now this makes it confusing that the `Test` contract [inherits](https://github.com/foundry-rs/forge-std/blob/c19dfd2f2a88a461216b0dd1f4961e1a85dcad46/src/Test.sol#L8) from `Script`. Why should it? They seem to be two separate concerns. If a user just so happens to define a function called `run` in their tests, that would be treated as a script entrypoint by Forge.

If the goal was to add the `vm` property to `Test`, we could simply redeclare the `VM_ADDRESS` and the `vm` variables in `Test`. Redundancy is not intrinsically bad. 

Actually, even before commit [f18682b2][1], I was a bit confused by the fact that the `Test` contract inherits from`Script`. Even in the Foundry Book, there's an implicit distinction made between [Writing Tests](https://book.getfoundry.sh/forge/writing-tests) and [Solidity Scripting](https://book.getfoundry.sh/tutorials/solidity-scripting). So even if this proposal isn't accepted, at the very least we should add a warning in the Foundry Book that contracts inheriting from `Test` also inherit from `Script`.

## Solution

1. Remove the inheritance from `Script` in `Test`.
2. Redeclare `VM_ADDRESS` and `vm` in `Test`.
3. Rewrite the methods in `Script` to have public visibility.
4. Rewrite the tests in `Script.t.sol` to instantiate an internal test contract `script`.

## Caveats

1. If this proposal passes, a child contract couldn't inherit from both at the same time, but if I understand the goals of `Test` and `Script` correctly, users are not expected to do this anyway.
2. As I understand it, Foundry excludes all contracts with the `IS_SCRIPT` flag set to `true` from the gas reports. But does it also exclude `IS_TEST` contracts? If yes, then this isn't a problem. But if not, why? If it does that for scripts it should also it for pure tests contracts. Happy to open an issue in Foundry to track this.

[1]: (https://github.com/foundry-rs/forge-std/commit/f18682b2874fc57d7c80a511fed0b35ec4201ffa)

paulrberg:
Closing as completed since https://github.com/foundry-rs/forge-std/pull/126 has been merged on the v0.3 branch.

[...]

Issue #123, opened 2022-07-15
Meta: New category StdUtils


paulrberg:
## Description

When I was first told about the [bound](https://github.com/foundry-rs/forge-std/blob/c19dfd2f2a88a461216b0dd1f4961e1a85dcad46/src/Test.sol#L139-L160) function, I had immediately assumed that it's some flavor of `assume`. So I had thought that I can use it like this:

```solidity
vm.bound
```

Of course, I had then learned what its purpose is and how it should be used.

Now, looking at its source code, I can see that there is neither a call to `vm`, nor a call to any other cheat defined in `Test`. So it doesn't seem to me like this is a "cheat" - it's more of like a utility function.

### 1st Proposal

Add a `STD-UTILS` delineator at the bottom of the `Test` contract:

```solidity
/*//////////////////////////////////////////////////////////////////////////
                                STD-UTILS
//////////////////////////////////////////////////////////////////////////*
```

Move the `bound` function underneath it, and write a new test file `StdUtils.t.sol` in which to move the following tests:

- `testBound`
- `testBoundUint256Max`
- `testCannotBoundMaxLessThanMin`

### 2nd Proposal

Create a new `StdUtils` contract and move `bound` there. `Test` would then inherit from `StdUtils.`

paulrberg:
Closing since #126 has been merged on the `v0.3` branch.

PR #126, opened 2022-07-15
Modularize forge-std

paulrberg:
Implements https://github.com/foundry-rs/forge-std/issues/122, https://github.com/foundry-rs/forge-std/issues/123, and https://github.com/foundry-rs/forge-std/issues/124.

## Modularization

- [x] Assertions and logs moved to `Assertions.sol`
- [x] Cheatcodes moved to `Cheats.sol`
- [x] Errors lib `stdErrors` moved to `Errors.sol`
- [x] Math lib `stdMath` moved to `Math.sol`
- [x] Utilities (such as `bound`)  moved to `Utils.sol` - also moved the utilities from `Script.sol` to `Utils.sol`
- [x] `Test` now inherits from `Assertions`, `Cheats` and `Utils`

## New Features

- [x] `createCompute2Address` utility

## Fixes

- [x] Drop the pragma to `0.8.0` in `StdErrors.t.sol`

## Breaking Changes

- [x] Removed the now-deprecated `lowLevelError ` custom error
- [x] Removed the now-deprecated `tip` cheat
- [x] `Test` doesn't inherit from `Script` anymore

I'm very much open to any feedback. Feel free to destroy this.


paulrberg:
> if we're going to a big refactor like this there's some other things we should consider including.

Yep, happy to move this to another bigger PR if appropriate.

Alternatively, happy to split this PR into its constituent components so reviewing each proposal is easier.

> splitting `Vm` into something like `VmSafe` and `Vm is VmSafe`, where `VmSafe` is a subset that's safe to use on scripts intended for a live network

That would be cool!

> As a side note, thanks for all the recent issues/PRs to forge-std! 🙌

Very glad that I can help.

paulrberg:
> it seemed monolithic was generally better

My proposal would still be "monolithic" for the end user - it's just that we internally would maintain the `Cheats` as a separate entity. The primary advantage (as I said in #122) would be to make Forge Std more modular and composable with other testing frameworks than DSTest.

> single inherit, no `MyLib.funcCall`

Yeah that wouldn't be cool.

ZeroEkkusu (commented):
I'm very supportive of making Forge Std modular to allow for maximal interoperability. It'd be great to standardize more things about testing in Solidity / frameworks, but that's out of scope of this PR, of course.

I just took a look at the changes and have a few questions.

One suggestion I have is to modularize every part of Forge Std (Std Errors, etc) and then just import them in `Test`.

Notes:

- We should make a new release before implementing these changes
- The `tip` std-cheat, which was deprecated, is removed

+1 for keeping naming simple with `Cheats`, `Storage`, etc!

src/Cheats.sol
```diff
@@ -0,0 +1,147 @@
...
+
+import "./Storage.sol";
+import "./Vm.sol";
+
+// Wrappers around Cheats to avoid footguns
+contract Cheats {
+    using stdStorage for StdStorage;
+
+    StdStorage private stdstore;
+    Vm private constant vm_cheats = Vm(address(uint160(uint256(keccak256("hevm cheat code")))));
```
gakonst, line 12: this should be inherited

```suggestion
    Vm internal constant vm_cheats = Vm(address(uint160(uint256(keccak256("hevm cheat code")))));
```

gakonst (changes requested):

src/Script.sol
```diff
@@ -1,17 +1,13 @@
...
 import "./console.sol";
 import "./console2.sol";
 
-abstract contract Script {
+abstract contract Script is Cheats {
     bool public IS_SCRIPT = true;
-    address constant private VM_ADDRESS =
-        address(bytes20(uint160(uint256(keccak256('hevm cheat code')))));
-
-    Vm public constant vm = Vm(VM_ADDRESS);
```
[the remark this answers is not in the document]
paulrberg: I have explained this in [Meta: Unbundling Test from Script](https://github.com/foundry-rs/forge-std/issues/124).

Happy to extract this change and move it to another PR, so we can discuss it there in isolation from the other changes made here.

src/Storage.sol
```diff
@@ -0,0 +1,252 @@
...
+    address _target;
+    bytes32 _set;
+}
+
+library stdStorage {
+    event SlotFound(address who, bytes4 fsig, bytes32 keysHash, uint slot);
+    event WARNING_UninitedSlot(address who, uint slot);
+
+    uint256 private constant UINT256_MAX = 115792089237316195423570985008687907853269984665640564039457584007913129639935;
+    int256 private constant INT256_MAX = 57896044618658097711785492504343953926634992332820282019728792003956564819967;
```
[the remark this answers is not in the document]
paulrberg: Good.

src/Storage.sol
```diff
@@ -0,0 +1,252 @@
...
+
+    Vm private constant vm_std_store = Vm(address(uint160(uint256(keccak256("hevm cheat code")))));
```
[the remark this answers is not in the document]
paulrberg: Okay.

src/Cheats.sol
paulrberg: You mean inherited in `Test`?

The problem with marking it as `internal` is that there would be collisions with the third-party testing frameworks which need to also define `vm` such that that framework can used independently of `forge-std` (for example, [I define `vm` in `PRBTest`](https://github.com/paulrberg/prb-test/blob/d781e55cc8eaf923b3eb4edb9a2a808a93c39a6a/src/PRBTest.sol#L65)).

I suggest leaving it as `private`. Forge Std already redeclares the `vm` in `Test`.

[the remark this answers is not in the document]
paulrberg: Okay, I thought that there would be a namespace collision, but it looks like it's fine even with just `vm`.

ZeroEkkusu (commented):
I left a couple of suggestions.

> [Modularize] I'll work on this and push another commit to this PR.

We should move out the assertions, as well?
The logs can stay inside `Test` because they are specific to Foundry. No need for a header.

As for `bound`, I thought we can add it and the utils from `Script` to `Utils.sol`, but haven't thought about this too much.

src/Cheats.sol
ZeroEkkusu: I think a good approach to enabling composability long-term may be for individual components (`Cheats`, etc) to provide only the functionality they are made for.
E.g. `Cheats` should expose std-cheats; not `vm` because it's a Forge thing and not a feature of `Cheats`. This will also prevent collisions when composing, and some devs might come up with interesting combinations we cannot foresee.

Additionally, every component should be a self-contained unit, with its own private `vm` and `stdstore` if it needs it, so it does not depend on another component. I don't think that would be too much hassle.

Specifically made end-user contracts, such as `Test` and `Script` can expose `vm` and `stdstore`.

paulrberg: Very much agree with @ZeroEkkusu's resolution here.

> every component should be a self-contained unit, with its own private `vm` and `stdstore`

+1

paulrberg:
> Should we move out the assertions, as well?

Just did it!

> The logs can stay inside `Test` because they are specific to Foundry, no need for a header.

Actually, no. I had to move the logs to the newly created `Assertions` file because the assertions depended upon some logs, e.g. `log_array`.

> As for bound, I thought we could add it and the utils from `Script` to `Utils.sol`, but haven't thought about this much.

Hah, this was my 2nd proposal from https://github.com/foundry-rs/forge-std/issues/123.

I just did this in the last commit. The only cost was the inability to use `log_named_uint` in `bound` anymore - though this should be fine? I don't see why users would want Forge Std to log the bounded value.

src/Utils.sol
```diff
@@ -0,0 +1,48 @@
...
+        if (size == 0) {
+            result = min;
+        } else if (size == UINT256_MAX) {
+            result = x;
+        } else {
+            ++size; // make `max` inclusive
+            uint256 mod = x % size;
+            result = min + mod;
+        }
+    }
```
ZeroEkkusu, line 22: > The only cost was the inability to use log_named_uint in bound anymore - though this should be fine? I don't see why users would want Forge Std to log the bounded value.

I think it's useful when a fuzz run reverts, because you can read what the value was, which helps with debugging. `bound` is an internal function, so Forge doesn't show the result in traces.

We can avoid defining an event with a workaround to log with `console` if the framework supports it. If it doesn't, nothing will break.

```suggestion
    function bound(uint256 x, uint256 min, uint256 max) internal virtual returns (uint256 result) {
        require(min <= max, "Test bound(uint256,uint256,uint256): Max is less than min.");

        uint256 size = max - min;

        if (size == 0) {
            result = min;
        } else if (size == UINT256_MAX) {
            result = x;
        } else {
            ++size; // make `max` inclusive
            uint256 mod = x % size;
            result = min + mod;
        }
        
        address(0x000000000000000000636F6e736F6c652e6c6f67).staticcall(
            abi.encodeWithSignature("log(string,uint)", "Bound result", result)
        );
    }
```

ZeroEkkusu (commented):
> I had to move the logs to the newly created Assertions file because the assertions depended upon some logs, e.g. log_array.

Right, I forgot about this!

Here's an idea for logging the `bound` result:

src/Script.sol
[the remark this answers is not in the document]
ZeroEkkusu: Maybe I'm misunderstanding you? From #78:

> This abstract Script contract will contain:
> - A minimal Vm interface with just the new broadcast cheatcodes, as well as "view" cheatcodes that don't change state. I may be missing some, but I think this includes: load, addr, getNonce, records, accesses, getCode, and label.

How would users access `vm` from their scripts?

src/Utils.sol
paulrberg: Love this! In the future we might want to create a private function for this so we can avoid repeating the `address(0x0000...)` syntax, but it works very well for the purposes of this PR.

src/Script.sol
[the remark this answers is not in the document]
paulrberg: Oh, shoot, yeah I was talking about something else. I have just re-added `vm` to `Script`.

paulrberg:
Okay, this is starting to look good! @ZeroEkkusu, I have made the following final touches:

1. Fix the compiler warning in `bound` - Solidity complained that the return value from the `staticcall` wasn't used.
2. Re-add `vm` to `Script` to fix the mistake you spotted in one of the comments above.
3. Define `console2`'s address as a constant in `Utils.sol`, for better readability.

I will work on fixing the git conflicts tomorrow if this also LGT everyone else?

src/Utils.sol
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.6.0 <0.9.0;
+
+contract Utils {
+    address internal constant CONSOLE2_ADDRESS = address(0x000000000000000000636F6e736F6c652e6c6f67);
```
ZeroEkkusu, line 5: Let’s keep it private for now?

```suggestion
    address private constant CONSOLE_ADDRESS = address(0x000000000000000000636F6e736F6c652e6c6f67);
```

ZeroEkkusu (commented):
It is starting to lg. Thank you for all the effort and patience.

> I will work on fixing the git conflicts

I checked today, the conflicts are minimal.

src/Utils.sol
```diff
@@ -0,0 +1,56 @@
...
+            result = x;
+        } else {
+            ++size; // make `max` inclusive
+            uint256 mod = x % size;
+            result = min + mod;
+        }
+        
+        // Log the result with console2.log if supported
+        (bool status, ) = address(CONSOLE2_ADDRESS).staticcall(
+            abi.encodeWithSignature("log(string,uint)", "Bound result", result)
```
mds1, line 26: This is the same mistake hardhat's `console.sol` made which is why `console2.sol` was needed 🙂

```suggestion
            abi.encodeWithSignature("log(string,uint256)", "Bound result", result)
```
Though, I'm not sure I fully understand why we wouldn't just import the console and call it, or declare `event log_named_uint` at the top of this file? There's lot of related issues, where's the best place to get an explanation of the new architecture and UX being implemented in this PR?

src/Cheats.sol
```diff
@@ -0,0 +1,147 @@
...
+            addr := create(0, add(bytecode, 0x20), mload(bytecode))
+        }
+
+        require(
+            addr != address(0),
+            "Test deployCode(string,bytes): Deployment failed."
+        );
+    }
+
+    function deployCode(string memory what)
```
mds1, line 132: Looks like the two `deployCode` overloads with value that were added in #133 are missing

src/Errors.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.6.0 <0.9.0;
```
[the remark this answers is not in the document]
mds1: I don't understand this comment. Is this trying to document the solidity version in which `Panic` errors were introduced? If so, where is there both ` >= 0.8.0` and `officially >=0.8.10`? This is worth clarifying, and potentially linking to the solidity docs for more info

src/Script.sol
```diff
@@ -1,39 +1,15 @@
...
-        // 0x84 = 0x80 + 0x04 (0x04 = the bytes length of the nonce, 4 bytes, in hex)
-        // We assume nobody can have a nonce large enough to require more than 32 bytes.
-        return addressFromLast20Bytes(keccak256(abi.encodePacked(bytes1(0xda), bytes1(0x94), deployer, bytes1(0x84), uint32(nonce))));
-    }
-
-    function addressFromLast20Bytes(bytes32 bytesValue) internal pure returns (address) {
-        return address(uint160(uint256(bytesValue)));
-    }
+    address internal constant VM_ADDRESS =
+        address(bytes20(uint160(uint256(keccak256("hevm cheat code")))));
```
mds1, line 13: Nit on removing unnecessary cast and putting this on one line

```suggestion
    address internal constant VM_ADDRESS = address(uint160(uint256(keccak256("hevm cheat code"))));
```

src/Test.sol
```diff
@@ -1,734 +1,19 @@
...
+import "./console.sol";
+import "./console2.sol";
+import "./Errors.sol";
+import "./Math.sol";
+import "./Storage.sol";
+import "./Utils.sol";
+
+abstract contract Test is Assertions, Cheats, Utils {
+    address internal constant VM_ADDRESS =
+        address(bytes20(uint160(uint256(keccak256("hevm cheat code")))));
```
mds1, line 15: Same nit on one line + unnecessary cast that way all declarations are consistent

src/Utils.sol
```diff
@@ -0,0 +1,56 @@
...
+        } else if (size == UINT256_MAX) {
+            result = x;
+        } else {
+            ++size; // make `max` inclusive
+            uint256 mod = x % size;
+            result = min + mod;
+        }
+        
+        // Log the result with console2.log if supported
+        (bool status, ) = address(CONSOLE2_ADDRESS).staticcall(
```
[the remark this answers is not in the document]
mds1: Or even
```suggestion
        (bool /* status */, ) = address(CONSOLE_ADDRESS).staticcall(
```
To remove the `// Silence compiler warning` section below

src/Utils.sol
```diff
@@ -0,0 +1,56 @@
...
+            abi.encodeWithSignature("log(string,uint)", "Bound result", result)
+        );
+        // Silence compiler warning
+        status; 
+    }
+
+    /// @dev Compute the address a contract will be deployed at for a given deployer address and nonce
+    /// @notice adapated from Solmate implementation (https://github.com/Rari-Capital/solmate/blob/main/src/utils/LibRLP.sol)
+    function computeCreateAddress(address deployer, uint256 nonce) internal pure virtual returns (address) {
```
mds1, line 34: Thoughts on throwing in a `computeCreate2Address` method in this PR?

```solidity
function computeCreate2Address(bytes32 salt, bytes32 initcodeHash, address deployer) internal pure returns (address) {
}
```

src/Utils.sol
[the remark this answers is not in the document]
paulrberg: @mds1 setting `bool /* status */` doesn't work:

> Type tuple(bool,bytes memory) is not implicitly convertible to expected type tuple(type(bool),).

src/Utils.sol
paulrberg: Re redeclaring the `log_named_uint` event - we cannot do that because there would be namespace collisions with the events defined in `Assertions`.

But yeah, importing `console2.sol` should be fine. Thoughs @ZeroEkkusu?

paulrberg:
Hey @mds1, thanks for your review.

> There's lot of related issues, where's the best place to get an explanation of the new architecture and UX being implemented in this PR?

@ZeroEkkusu and I have worked our way up to modularize the code base commit by commit .. besides the original proposals that I made in https://github.com/foundry-rs/forge-std/issues/122, https://github.com/foundry-rs/forge-std/issues/123, and https://github.com/foundry-rs/forge-std/issues/124, there's no central place where all changes are documented. What I will do now though is write a brief spec in my PR comment body at the top - let me know what you think.

src/Cheats.sol
paulrberg: That's because I haven't rebased yet. Will do so now.

src/Utils.sol
paulrberg: Well, we've gone so far already. Why not? 😅

paulrberg:
Okay guys, I have:

1. Rebased from `master`.
2. Imported `console2` in `Utils` so we can ditch the low-level `staticcall`.
3. Added `computeCreate2Address` utility.

Also, I have renamed the title of this PR and added a changelog in the original comment above.

src/Errors.sol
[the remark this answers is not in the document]
paulrberg: `Test` imports `Errors`, so we need to keep the `>=0.6.0 <0.9.0` pragma in `Errors.sol`. But panic errors have been introduced in v0.8.0.

Though I am also not sure what @ZeroEkkusu meant by "officially >=0.8.10" - perhaps panics had a bug up until that version?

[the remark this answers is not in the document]
paulrberg: Yeah I think what Brock said there is wrong. Panics have been added in v0.8.0.

I'll tentatively drop the pragma to `>=0.8.0` in all files and remove your comment about `v0.8.10` @ZeroEkkusu.

ZeroEkkusu:
@mds1 

> There's lot of related issues, where's the best place to get an explanation of the new architecture and UX being implemented in this PR?

In addition to @paulrberg's [answer](https://github.com/foundry-rs/forge-std/pull/126#issuecomment-1188726514):

- [the exchange about composability](https://github.com/foundry-rs/forge-std/pull/126#discussion_r922906437)
- additionally, we can make Assertions `virtual`, since everything else is `virtual`

This creates a distinction between _users_ (those who test with Forge Std) and _devs_ (those who build on top of Forge Std),
As well as _components_ (e.g. `Cheats`) and _UX contracts_ (e.g. `Test`).

For example,

A dev wants to create their own fork of `Cheats`? No problem:

```solidity
import {Cheats} from "forge-std/Cheats.sol";

contract ZECheats is Cheats {
    function hoax(address who) internal virtual override {
       // different
    }
    
    function newCheat() internal virtual {
       // new
    }
}
```

Push it to GH and users can use it as a submodule to swap out components:

```solidity
import "forge-std/Test.sol";
import "ze-cheats/ZECheats.sol";

contract CustomTest is TestBase, Assertions, ZECheats, Utils {
    // anything else
}
```

What is `TestBase`? I just realized we should create a base with the standard content `vm` etc. in the same file (`Test.sol` and `Script.sol`) and inherit it in the UX contract. This is better UX if we agree what I'm proposing makes sense, as the user would not have to write the boilerplate themselves.

Completely new UX contracts can be created, even not backward-compatible, because we have modularized everything:

```solidity
pragma solidity >=0.8.0 <=0.9.0;

import "forge-std/Test.sol";
import {PRBAssertions} from "prb-test/PRBAssertions.sol"; // new assertions, capital letter logs
import {Helpers} from "prb-test/Helpers.sol"; // free functions, unchecked

contract Test is TestBase, PRBAssertions, Cheats, Utils {
    // anything else
}
```

A dev wants to improve UX? No problem, ship it as a Forge Std fork:

```solidity
import "xyz-test/Test.sol"; // Forge Std but loads .env automatically

    // test
}
```

Users and devs will still have the modularity of Forge Std while enjoying the added features:

```solidity
import "xyz-test/Test.sol"; // exports upgraded TestBase
import "ze-cheats/ZECheats.sol";

contract CustomTest is TestBase, Assertions, ZECheats, Utils {
    // anything else
}
```

<br>

These are some ideas I've had. As Paul says, I'm open to any feed back; feel free to destroy this.
Code snippets may not be accurate.

src/Assertions.sol
```diff
@@ -0,0 +1,232 @@
+// SPDX-License-Identifier: MIT
```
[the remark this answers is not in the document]
paulrberg: Agree #1 and #3, and like #2, though to get the ball rolling with the current PR I'd suggest first merging this into the new `v0.3` branch and working on that separately.

ZeroEkkusu:
> I'm open about TestBase, but I'm a bit hesitant because I don't know how useful it would be

Let me clarify why I proposed this:

`TestBase` is more of a necessity than a convenience.

If you want to swap out `Cheats` for `ZECheats`, you can't always do this:

```solidity
contract CustomTest is Test, ZECheats {}
```

This is because if `ZECheats` does not only add new cheats but also overrides existing ones, you will run into a conflict and have to do a manual override in your `CustomTest`:

```solidity
contract CustomTest is Test, ZECheats {
    function hoax(address who) internal virtual override (Test, ZECheats) {
       ZECheats.hoax(who);
    }
}
```

Which brings us to this pattern:

```solidity
contract CustomTest is Assertions, ZECheats, Utils {
    address internal constant VM_ADDRESS = address(uint160(uint256(keccak256("hevm cheat code"))));

    StdStorage internal stdstore;
    Vm internal constant vm = Vm(VM_ADDRESS);
}
```

While this is not a problem for a dev, or they may want to do things completely deifferently, as you said, why would a user have to write the same standard boilerplate when it can be replaced with a single word:

```solidity
contract CustomTest is TestBase, Assertions, ZECheats, Utils {}
```

Moreover, bases allow Forge Std forks to export their upgraded `TestBase`, which in turn allows the modularity cycle to continue:

`xyz/Test.sol` (Forge Std but automatically loads .env)

```solidity
import "forge-std/Assertions.sol";
import "forge-std/Cheats.sol";
import "forge-std/console.sol";
import "forge-std/console2.sol";
import "forge-std/Errors.sol";
import "forge-std/Math.sol";
import "forge-std/Storage.sol";
import "forge-std/Utils.sol";

import {TestBase as SuperBase} from "forge-std/Test.sol";

contract TestBase is SuperBase {
    constructor() {
        // load .env
    }
}

contract Test is TestBase, Assertions, Cheats, Utils {}
```

Now, if you're using `xyz/Test.sol` fork, your `.env` file also loads automatically. This is not the best example, as loading `.env` automatically may be better as a plugin and not a completely new fork of Forge Std, but you get the idea.

This is where the cycle continues: if another dev want to build on top of `xyz/Test`, they need only inherits its upgraded `TestBase` to get the functionality.

A good pattern for UX contract may be like this:

`fork/Test.sol`

```solidity
import "forge-std/Components.sol"; // inherit all components with a single import
import {TestBase as SuperBase} from "forge-std/Test.sol"; // import the standard boilerplate if you need it

contract ExampleTask0 {
    constructor() {
        // functionality
    }
}

contract ExampleTask1 {
    constructor() {
        // functionality
    }
}

contract TestBase is SuperBase, ExampleTask0, ExampleTask1 {}

contract Test is TestBase, Assertions, Cheats, Utils {}
```

*Note: `ExampleTask0` and `ExampleTask0` can be either contracts or free functions (`>= 0.7.1`).*

If a fork of this fork wants to import its `Test` to add more on top, it can. If it wants to import only its `TestBase` to keep the functionality but use custom components, it can. If it wants to import only its Task1, it can. This would allow for maximum modularity while keeping everything clean.
The point is, if a dev wants to do something, they can. We lose nothing by doing this (and we're already here with the changes).

<br>

As a side not, a "fork" would be something like how Hardhat uses its Ethers fork to add helpers for HH, or who knows what devs would come up with.
Again, feel free to absolutely destroy this.

paulrberg:
Ok @ZeroEkkusu, I'm sold. I have implemented `TestBase` and `Components` in my latest commit!

> If you want to swap out Cheats for ZECheats

Ah, ok, I haven't considered this use case.

> you will run into a conflict and have to do a manual override

Yep, just like it happened with the `bound` function in https://github.com/foundry-rs/forge-std/pull/92.

> bases allow Forge Std forks to export their upgraded TestBase, which in turn allows the modularity cycle to continue:

Oh, this would be cool indeed!

> This would allow for maximum modularity while keeping everything clean.

Turtles all the way down!

ZeroEkkusu:
Alright, we'll see what others say.

I've made proof-of-concept repos to try it myself:

- [Proof-of-concept modular Forge Std](https://github.com/ZeroEkkusu/poc-forge-std)
- [Proof-of-concept Forge Std fork](https://github.com/ZeroEkkusu/poc-forge-std-fork)
- [Example project with Forge Std fork](https://github.com/ZeroEkkusu/ex-forge-std-fork-project)

One thing I noticed is that every fork should have its `Components.sol` (a DX contract, you could say), which exports the Forge Std's components and any new components. It should also always have the UX contracts (`Test.sol` and `Script.sol`).

src/Assertions.sol
```diff
@@ -0,0 +1,232 @@
...
+pragma solidity >=0.6.0 <0.9.0;
+
+import "ds-test/test.sol";
```
ZeroEkkusu, line 4: We have a problem.

I've noticed `IS_TEST` is missing from the Test Base. I looked at DS Test and it's `bool public IS_TEST` (not overridable). Moreover, DS Test contains other logic that the base must contain. See [`DSTest`](https://github.com/dapphub/ds-test/blob/master/src/test.sol).

However, `Test` doesn't inherit from `DSTest` directly anymore. `Assertions` inherit `DSTest` and `Test` inherits `Assertions`, which means `Test` is directly depended on `Assertions` for this functionality.

I propose we merge `TestBase` and `Assertions` into `TestBase` because `DSTest` is not flexible. Suggestions welcome.

paulrberg: Good catch. This got me thinking about the extensibility of PRBTest, so I went ahead and [refactored `IS_TEST` into a virtual function](https://github.com/paulrberg/prb-test/commit/834ded25410f7ff40273932f681a39f19b746b52).

> I propose we merge `TestBase` and `Assertions` into `TestBase`

This would gnaw away at the original goals of this PR (maximum modularization).

> for `prb-test` to provide `Legacy.sol` with unbundled `DSTest`

Oh, I would prefer PRBTest to not touch DSTest at all so that I don't have to deal with any GPL licensing issue. There wouldn't be much value in this anyway since as you pointed out you'd have to import `ds-test/test.sol` anyway.

> inherit `DSTest` in both `TestBase` and `Assertions`

I like this.

ZeroEkkusu: @paulrberg 

> This would gnaw away at the original goals of this PR

100%

<br>

I suggest we put `TestBase` in `Test.sol`, like I did in the [proof-of-concept](https://github.com/ZeroEkkusu/poc-forge-std/blob/master/src/Test.sol).
And create `ScriptBase`, in the same way. It should behave the same as `TestBase`, so `IS_SCRIPT` should be in `Script` instead:

```solidity
abstract contract ScriptBase {
    address internal constant VM_ADDRESS = address(uint160(uint256(keccak256("hevm cheat code"))));
    Vm internal constant vm = Vm(VM_ADDRESS);
}

abstract contract Script is ScriptBase {
    bool public IS_SCRIPT = true;
}
```

<br>

I think we can review the PR once again after that and consider merging into `v0.3`.

paulrberg: Yeah I'm okay with this. Users needing just `TestBase` can import it as a specific symbol.

Implemented below!

ZeroEkkusu (approved):
Team, this is the recap of the PR:

- No breaking changes besides the ones we've made deliberately
- The library is now modular
- One gotcha: `DSTest` is not flexible and we could not separate `IS_TEST` from the assertions; we inherited `DSTest` in both `Test` and `Assertions`

Lgtm.
@mds1, we should consider merging into `v0.3`.

I will open an issue to track proposals for `v0.3`.

paulrberg:
Thanks very much @ZeroEkkusu for your continuous guidance on this PR!

> No breaking changes besides the ones we've made deliberately

In case you're looking for a list of breaking changes, see my [top comment](https://github.com/foundry-rs/forge-std/pull/126#issue-1305916561).

gakonst (approved):
Went through the code and the refactor LGTM. Trust you guys above that the user testing makes sense and that there's no critical breaking changes. Would be helpful to document if any.

ZeroEkkusu:
Alright, we're merging this into `v0.3`! Thank you @paulrberg for taking the initiative and bearing with us!

If Matt finds something is missing, we can make changes in `v0.3`.

paulrberg:
Whoop whoop!

I went ahead and closed issues #122, #123 and #124.

PR #138, opened 2022-07-26
feat: add parseJson helper functions

odyslam:
Helper functions that leverage the new `parseJson`  cheat code and enable the user to easily parse `forge script` artifacts.

The Structs will probably change a bit, due to the artifacts requiring some changes to avoid `Solidity keyword` keys and make the artifacts be consistent (produce keys with null values instead of omitting them)


mds1:
Now that `vm.parseJson` is merged (thanks a lot @odyslam!) let's revisit this PR and get the required Vm interface updates, types, and any helpers merged

odyslam:
Blocked on https://github.com/foundry-rs/foundry/pull/2914

Currently, for some reason, a particular Receipt on the broadcast fixture is not decoded properly for unknown reasons.

odyslam:
All issues fixed and `--via-ir` is not required any more

odyslam:
The ci fail is propably because they fix PR that was just merged has not be yet propagated

mds1:
If I rebase locally against main all tests pass, seems this branch is missing the `projectRoot` cheatcode definition from https://github.com/foundry-rs/forge-std/pull/158, so let's try that and see if it fixes CI? Though the CI failure does seem different than missing `projectRoot`

src/Test.sol
```diff
@@ -2,6 +2,7 @@
 pragma solidity >=0.6.0 <0.9.0;
 
 import "./Script.sol";
+import "./Vm.sol";
```
mds1, line 5: Script imports `Vm` so we can delete this

src/Test.sol
```diff
@@ -448,8 +449,471 @@ abstract contract Test is DSTest, Script {
...
+        string[] arguments;
+        address contractAddress;
+        string contractName;
+        // json value name = function
+        string functionName;
+        bytes32 hash;
+        // json value name = tx
+        RawEIP1559TransactionDetail detail;
+        // json value name = type
+        string contractType;
```
mds1, line 467: Maybe an issue for foundry repo, but I think we should change the JSON names. `function` and `type` are keywords in JS/TS too so having them as field names doesn't seem ideal.

- I like `functionName`
- I'd propose `txDetail` or `txInfo` instead of just `detail`
- `contractType` should be `txType` (e.g. type 0, 1, or 2) IIUC

src/Test.sol
```diff
@@ -448,8 +449,471 @@ abstract contract Test is DSTest, Script {
...
+        bytes data;
+        address from;
+        bytes gas;
+        bytes nonce;
+        address to;
+        bytes txType;
+        bytes value;
+    }
+
+    struct EIP1559Transaction {
```
mds1, line 481: Do these struct names match anything in the foundry repo? If not, wdyt of calling them `RawEIP1559Tx` and `ParsedEIP1559Tx` (a little more brief + descriptive). Similar change for legacy transaction struct.

Alternatively, `RawType0Tx`, `ParsedType1Tx`, etc. is another idea

src/Test.sol
```diff
@@ -448,8 +449,471 @@ abstract contract Test is DSTest, Script {
...
+        uint256 blockNumber;
+        bytes data;
+        uint256 logIndex;
+        bytes32[] topics;
+        uint256 transactionIndex;
+        uint256 transactionLogIndex;
+        bool removed;
+    }
+
+    struct TransactionReturn {
```
mds1, line 595: When is this type used? In general might be helpful to help a short comment above each struct/function explaining where it's from / when to use it

src/Test.sol
```diff
@@ -448,8 +449,471 @@ abstract contract Test is DSTest, Script {
...
+        artifact.path = rawArtifact.path;
+        artifact.timestamp = rawArtifact.timestamp;
+        artifact.pending = rawArtifact.pending;
+        artifact.transactionReturns = rawArtifact.transactionReturns;
+        artifact.receipts = rawToConvertedReceipts(rawArtifact.receipts);
+        artifact.transactions = rawToConvertedEIP1559Txs(rawArtifact.transactions);
+        return artifact;
+    }
+
+    // break into two functions
```
mds1, line 638: Is this a TODO or outdated comment?

src/Test.sol
```diff
@@ -448,8 +449,471 @@ abstract contract Test is DSTest, Script {
...
+    function rawToConvertedEIP1559Txs(RawEIP1559Transaction[] memory rawTxs)
+        internal
+        returns (EIP1559Transaction[] memory)
+    {
+        EIP1559Transaction[] memory txs = new EIP1559Transaction[](rawTxs.length);
+        for(uint i;i<rawTxs.length;i++){
```
mds1, line 644: Style nit (applies to all `for` loops in this PR)
```suggestion
        for (uint i; i < rawTxs.length; i++) {
```

src/Test.sol
```diff
@@ -448,8 +449,471 @@ abstract contract Test is DSTest, Script {
...
+        internal
+        returns (EIP1559Transaction memory)
+    {
+            EIP1559Transaction memory transaction;
+            transaction.arguments = rawTx.arguments;
+            transaction.contractName = rawTx.contractName;
+            transaction.functionName = rawTx.functionName;
+            transaction.hash= rawTx.hash;
+            transaction.detail = rawToConvertedEIP1559Detail(rawTx.detail);
+    }
```
mds1, line 660: whitespace nit
```suggestion
    {
        EIP1559Transaction memory transaction;
        transaction.arguments = rawTx.arguments;
        transaction.contractName = rawTx.contractName;
        transaction.functionName = rawTx.functionName;
        transaction.hash= rawTx.hash;
        transaction.detail = rawToConvertedEIP1559Detail(rawTx.detail);
    }
```

src/Test.sol
```diff
@@ -448,8 +449,471 @@ abstract contract Test is DSTest, Script {
...
+        string memory key = string(abi.encodePacked(".transactions[",vm.toString(index), "]"));
+        bytes memory parsedDeployData =
+            vm.parseJson(deployData, key);
+        RawEIP1559Transaction memory rawTx = abi.decode(parsedDeployData, (RawEIP1559Transaction));
+        return rawToConvertedEIP1559Tx(rawTx);
+    }
+
+
+    // Analogous to readTransactions, but for receipts.
+    // It requires --via-ir to avoid `stack too deep errors`
```
mds1, line 707: This comment is now outdated right?

src/Test.sol
```diff
@@ -448,8 +449,471 @@ abstract contract Test is DSTest, Script {
...
+            logs[i].topics = rawLogs[i].topics;
+            logs[i].transactionIndex = bytesToUint(rawLogs[i].transactionIndex);
+            logs[i].transactionLogIndex = bytesToUint(rawLogs[i].transactionLogIndex);
+            logs[i].removed = rawLogs[i].removed;
+        }
+        return logs;
+
+    }
+
+    function bytesToUint(bytes memory b) internal pure returns (uint256){
```
mds1, line 781: This function is used a lot but we don't have any tests for it. Alternatively, you can use assembly to convert to bytes32 and then just cast to uint

src/Test.sol
```diff
@@ -448,8 +449,471 @@ abstract contract Test is DSTest, Script {
...
+            for(uint i=0;i<b.length;i++){
+                number = number + uint(uint8(b[i]))*(2**(8*(b.length-(i+1))));
+            }
+        return number;
+    }
+
 }
 
+// Helpers for parsing keys into types. We'd include these for all value types
+contract JsonParser is Test {
```
mds1, line 792: Wdyt about making this a library with all internal methods and having `Script` import it? That would be more consistent with how existing libs like `console` are used

mds1 (commented):
This is looking awesome, thanks a lot for all the work!

src/Test.sol
odyslam: I don't think that would work, cause the parser has state, it's not stateless

odyslam:
I addressed all your comments, except for the library. The parser needs the state to have a nicer UX. It also needs console2, vm, so the imports would get a bit funky. 

What I can't reason about is the `projectRoot` thing and why locally it seems to be OK but the file in GH is not.

src/Json.sol
```diff
@@ -0,0 +1,139 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.6.0 <0.9.0;
+
+import "console2.sol";
+import "Vm.sol";
+
+// Helpers for parsing keys into types.
+contract JsonParser {
```
ZeroEkkusu, line 8: What do you think about this?

```solidity
using JsonParser for string;

function test() public {
    string memory someJson = vm.readFile("...");
    address someAddr = someJson.readAddress("...");
    // ...
}
```

Also, we should choose that is in line with https://github.com/foundry-rs/forge-std/pull/157. I suggest `stdJson` if it is going to be a library, and `StdJson` otherwise. We can improve the function names as necessary.

mds1: Yea I like this, making it a lib definitely feels like the right approach and should be doable:
- The current `path` storage var is just used to alias `readFile` to `readJson`. I'd argue this isn't necessary, but could add a helper method to the lib if we like the explicit naming.
- By making `string memory json` the first arg for all methods, we don't need the `json` storage var and can use the `using for` syntax @ZeroEkkusu has above.
- I think we can remove the `console2` import/usage, as that's easy for users to implement themselves and doesn't add much value IMO

src/Test.sol
mds1: @odyslam saw you marked this resolved—did we change the JSON outputs upstream already or should we create an issue to track?

src/Test.sol
mds1: Saw you marked this one resolved but the comment is still there, so just want to follow up

src/Test.sol
mds1: On second thought, naming them after the actual type number such as `RawType2Tx` might not work because I'm not sure if other networks follow the same conventions. For example Polygon has EIP-1559 support, but is it still considered a "type 2" tx there? Not sure. So I like the current naming convention we now have, marking this resolved, thanks @odyslam

mds1:
I'd say three things left before merging this, all easy things at this point

- [ ] Refactor into a library using this approach: https://github.com/foundry-rs/forge-std/pull/138#discussion_r956749279
- [ ] Make sure we track potential JSON output renames in the foundry repo (cc @mattsse @joshieDo for any thoughts): https://github.com/foundry-rs/forge-std/pull/138#discussion_r956333364
- [ ] Potential outdated comment? https://github.com/foundry-rs/forge-std/pull/138#discussion_r956340962

src/Test.sol
odyslam: So, we don't **have** to follow the naming convention of foundry. I propose to have the best UX here and then work out to upstream to foundry, but we shouldn't block it until it's upstreamed.

Moreover, a lot of naming changes originate from ethers-rs. At some point I need to change some things there, including the `omit if none` behaviour, which could potentially break the integration here.

mds1: Ah, yea totally agree we shouldn't block on that—just wanted to make sure we don't lose track of the potential upstream change. I'll create a foundry issue to track and move discussion of potential name changes there, thanks!

mds1: As I'm creating this, I just noticed it seems the field is called `transaction` in the JSON, not `tx`

mds1: Also, looks like `function` is the full signature, e.g. `transfer(address,uint256)`, not just the name, so we should rename that to `functionSig` (or `functionSignature`) here

mds1: Added both of these to the pre-merge tracker comment in https://github.com/foundry-rs/forge-std/pull/138#issuecomment-1229496911, lmk what you think

odyslam:
@mds1 I like txDetail, cause we call the outer struct Transaction. It feels very intuitive.

odyslam:
@mds1 what's up with `projectRoot`? Locally, I have only one definition in the interface.

mds1:
Yea I don't know why the CI is failing, I'm gonna take a closer look later today so don't worry about it

odyslam:
Thank you guys for helping me push this through the finish line. Your suggestions were really good.

src/stdJson.sol
```diff
@@ -0,0 +1,120 @@
...
+
+import "Vm.sol";
+
+// Helpers for parsing keys into types.
+library JsonParser {
+
+    address constant private VM_ADDRESS =
+        address(bytes20(uint160(uint256(keccak256('hevm cheat code')))));
+
+    Vm internal constant vm = Vm(VM_ADDRESS);
```
ZeroEkkusu, line 12: This is to make the library `v0.3`-ready. We use this approach for all components on that brach.

```suggestion
    Vm private constant vm = Vm(address(uint160(uint256(keccak256("hevm cheat code")))));
```

src/stdJson.sol
```diff
@@ -0,0 +1,120 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.6.0 <0.9.0;
+
+import "Vm.sol";
+
+// Helpers for parsing keys into types.
+library JsonParser {
```
ZeroEkkusu, line 7:
```suggestion
library stdJson {
```

src/stdJson.sol
```diff
@@ -0,0 +1,120 @@
...
+    Vm internal constant vm = Vm(VM_ADDRESS);
+
+    function parseRaw(string memory json, string memory key)
+        internal
+        returns (bytes memory)
+    {
+        return vm.parseJson(json, key);
+    }
+
+    function readUint256(string memory json, string memory key)
```
ZeroEkkusu, line 21: Should we simplify the name?
```suggestion
    function readUint(string memory json, string memory key)
```

src/test/StdCheats.t.sol
```diff
@@ -2,10 +2,13 @@
 pragma solidity >=0.7.0 <0.9.0;
 
 import "../Test.sol";
+import "../Json.sol";
```
ZeroEkkusu, line 5:
```suggestion
import "../StdJson.sol";
```

src/Script.sol
```diff
@@ -1,15 +1,16 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.0 <0.9.0;
 
-import "./Vm.sol";
 import "./console.sol";
-import "./console2.sol";
+import "./Json.sol";
```
ZeroEkkusu, line 5:
```suggestion
import "./StdJson.sol";
```

src/Script.sol
```diff
@@ -1,15 +1,16 @@
...
 
 abstract contract Script {
     bool public IS_SCRIPT = true;
     address constant private VM_ADDRESS =
         address(bytes20(uint160(uint256(keccak256('hevm cheat code')))));
 
+    using stdJson for string;
+
```
ZeroEkkusu, line 13: Library usage is not inherited, so this has no effect.

```suggestion
```

src/Script.sol
```diff
@@ -1,15 +1,16 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.0 <0.9.0;
 
-import "./Vm.sol";
 import "./console.sol";
```
ZeroEkkusu, line 4: Restore the imports.

```suggestion
import "./Vm.sol";
import "./console.sol";
import "./console2.sol";
```

ZeroEkkusu (changes requested):
We need to fix a few things.

Also:

- remove `Json.sol`
- rename `stdJson.sol` to `StdJson.sol`

src/Vm.sol
```diff
@@ -172,4 +172,33 @@ interface Vm {
     function deriveKey(string calldata, uint32) external returns (uint256);
     // Derive a private key from a provided mnenomic string (or mnenomic file path) at the derivation path {path}{index}
     function deriveKey(string calldata, string calldata, uint32) external returns (uint256);
+        // Returns the path to the project root, () => (path)
+    function projectRoot() external returns (string memory);
```
ZeroEkkusu, line 176: The CI is failing because this PR adds `projectRoot`, but we already have it in `Vm`.

```suggestion
```

odyslam: But locally I only have a single `projectRoot` in my `vm.sol` ?

ZeroEkkusu: We merged it in https://github.com/foundry-rs/forge-std/pull/158, so your `Vm` became outdated and Git resolved your `projectRoot` as a new addition to `Vm` here.

src/stdJson.sol
mds1: We currently have `StdJson`, but should be `stdJson` to be consistent with e.g. `stdStorage`. @ZeroEkkusu just double check me on that

odyslam:
wrt the tests not passing. Shouldn't solc 8.0 support struct decoding?

ZeroEkkusu:
Those are tests for backward compatibility with solc `0.8.0`, `0.7.0`, `0.6.0`.

One option is to decode the structs manually. I'm not sure how struct arrays are encoded.

src/stdJson.sol
ZeroEkkusu: @mds1 Yes, the filename should be `StdJson.sol` and the library name `stdJson`.

mds1:
I'd be ok with only supporting 0.8.x for now (and updating the `StdJson.sol` pragmas + CI as needed) and opening a separate issue to track supporting the `stdJson` lib for older versions, to avoid blocking this PR. Since this is a new feature no one is currently relying on it so only supporting 0.8.x won't break any backwards compatibility, and they'll still be able to use `parseJson`, just not the helper lib.

ZeroEkkusu:
We can use "predeploys". Here's proof-of-concept:

We write a contract with the problematic std-cheats in the latest Solidity. All functions should be `external pure`. We note down its runtime bytecode.

Then, in Forge Std, we "predeploy" this contract with `vm.etch`. Making a static call to a desired function will decode the data into a struct.

E.g. `decoder.decodeReceiptArray(parsedDeployData)`:

```solidity
function readReceipts(string memory path)
         internal
         returns (Receipt[] memory)
     {
         string memory deployData = vm.readFile(path);
         bytes memory parsedDeployData = vm.parseJson(deployData, ".receipts");
         RawReceipt[] memory rawReceipts = decoder.decodeReceiptArray(parsedDeployData);
         return rawToConvertedReceipts(rawReceipts);
     }
```

Proof:

image

mds1:
Ah nice, that is a great idea! Some thoughts:

1. I think it still will require some changes to CI anyway? (since we'll have a contract in this repo that only compiles with 0.8)
2. I'm worried this might get a bit messy, so we probably want a simple script to automate the process of compiling + injecting the bytecode into the contracts + ensuring the 0.8-specific code is isolated.
4. I'm imagining the lib now calls out to the etched predeploy, so UX-wise it's still the same as a native library?
5. One downside is that because these are external calls they'll now show up in traces, but I don't think there's anything we can do about that if we go this route.

Also, regardless of which approach we use, will the existence of a contract with a 0.8 pragma break builds for users with an 0.6/0.7 pragma, or will that file just get skipped during compilation? I'm not sure offhand.

I do like this idea, but FWIW I'd still be ok only supporting 0.8 to start, then making this change afterwards if requested, since it does add some tech debt/maintenance burden/overhead that we'll need to maintain.

@gakonst lmk if you have any thoughts here, not sure if you have a better grasp of how many 0.6/0.7 users there are + if they'd need this feature

ZeroEkkusu:
If only solc `>= 0.8.0` is going to be supported, we must remove the import in `Script` and let users import the library manually. There would be no compilation problems on their end, and CI here would continue to work.

gakonst:
Could the tests not be 0.8.0 and the contract itself be >=0.6.0 or something?

mds1:
@gakonst I don't think that helps since the decoding is done in the main `Test.sol` contract, not in the tests, e.g. check the failed CI here: https://github.com/foundry-rs/forge-std/runs/8059709854?check_suite_focus=true#step:6:38

mds1:
Got it, so IIUC correctly the suggestion here is that:
- Users might be writing contracts with solidity 0.6 or 0.7
- But, they can still write their tests in 0.8 and therefore forge-std only needs to support 0.8?

One limitation there is I don't think the toml file currently is flexible enough to support various solc configs? So you're stuck with the defaults for the test version, which maybe is ok. (The config file definitely should support solc profiles per-contract/file though, not sure if that's being tracked somewhere)

odyslam:
I think that forge will automatically use the appropriate solc version? W

ZeroEkkusu:
To summarize the discussion:

- The new std-cheats use struct decoding, which is available in Solidity `>=0.8.0`
- Forge Std supports Solidity versions as low as `0.6.0`
- A common practice is to import a contract you want to test into a `.t.sol` file with a test contract
- If the contract is using a Solidity version below `0.8.0`, the project **will not** compile

```solidity
pragma solidity >=0.6.0 <0.8.0;

contract Contract {}
```

```solidity
// any pragma

import "src/Contract.sol";

import "forge-std/Test.sol";

    Contract c;

    function setUp() public {
        c = new Contract();
    }
}
```

mds1:
I [searched sourcegraph](https://sourcegraph.com/search?q=context:global+solc%28-%7C_%29version%5Cs*%3D%5Cs*%28%5C%22%7C%5C%27%290.%5Cd.%5Cd*%28%5C%22%7C%5C%27%29+file:foundry.toml&patternType=regexp) and found 112 repos that specify a solc version in `foundry.toml`. Of those 110 are 0.8.x and 2 are 0.7.x (nomad and beanstalk). 

I know Maker uses <0.8 in a few repos but they are migrating away from that, and I've gotten confirmation they'd be ok with new forge-std features being >= 0.8 only.

@odyslam Do you know if nomad would mind if this parseJson feature is only >= 0.8?

If not, I think the simplest path forward is to make the library >= 0.8 but not import it by default to avoid breaking existing builds. Then, we can import by default whenever (1) we drop support for < 0.8, or (2) we modify the lib to decode manually / use the predeploy trick to add support for < 0.8

odyslam:
@mds1 would it help if we defined `pragma experimental ABIEncoderV2;`?

That being said, since there is no other way around it, I think it's fair to
a) move the helper functions from the Test contract to the library
b) Make the library compatible with >=0.8

Personally, the pre-deploy seems too complicated and not worth the effort for backward compatibility. 

If we are aligned, I will do (a) and let's merge this gigantic PR

ZeroEkkusu:
Sgtm; let's make it an `abstract contract` so we can merge it with StdCheats later without breaking changes:

```solidity
import "forge-std/JsonStdCheats.sol";

contract ContractTest is Test, JsonStdCheats {}
```

If we did a library, it would require this syntax (can't merge seamlessly in the future):

```solidity
lib.stdCheat();
```

<br>

We should address https://github.com/foundry-rs/forge-std/pull/138#discussion_r956757666 as well.

odyslam:
@ZeroEkkusu  -- done. 

For some reason, it compiles with 0.6.0 and 0.7.0, but not with 0.7.6 🙃

odyslam:
@ZeroEkkusu @mds1 can I get an ack on alignment for the design before going into implementation

mds1:
Sorry, catching up now. What's the proposed design here? Is it a >=0.8 lib, or an abstract contract? 

Using `pragma experimental ABIEncoderV2;` was a good idea and I support that if it brings backwards compatibility

odyslam:
@mds1 abstract contract so it can be compatible (with what I assume) a future refactoring of how cheats works in the std lib

mds1:
If we change to an abstract contract, we lose the `myJson.readUint(key)` syntax and instead need `readUint(myJson, key)` which is ok, though personally I prefer the lib syntax, curious what you each prefer?

@ZeroEkkusu Not sure I follow why using a lib is harder for future merging? Couldn't you just have it imported into `Script` and `Test` by default, declare `using stdJson for string`, then use it?

Really weird that it works with other 0.6/0.7 solc versions but not 0.7.6, was there a bug in that solc version?

odyslam:
1. It doesn't work for 0.7.5 either
2. I am alright with `JsonParser parser = new JsonParser(json); parser.readUint(key)`, but the lib syntax is also sweet.

I want to merge this so that I can work then on:
1. LegacyTx definitions
2. Artifact decoding (it still can't decode whole artifact, but can it's receipts/transactions)
3. writeJson

mds1:
> That being said, since there is no other way around it, I think it's fair to
> a) move the helper functions from the Test contract to the library
> b) Make the library compatible with >=0.8
> 
> Personally, the pre-deploy seems too complicated and not worth the effort for backward compatibility.
> 
> If we are aligned, I will do (a) and let's merge this gigantic PR

Personally I think both a and b here is the best approach. From everything I gathered, not supporting below 0.8 is totally fine so I don't think we should worry about that. And since the lib syntax is nice, and makes it easy to use, let's go with it (unless @ZeroEkkusu confirms there's some reason we can't do lib). 

+1 on getting this merged soon, we're really close to the finish line

mds1:
Commit [7daae38](https://github.com/foundry-rs/forge-std/pull/138/commits/7daae38aa5261e8a50f5ce5cace7670c0577b5ab) had CI passing and I believe used the lib format, should we just drop the last commit, fix https://github.com/foundry-rs/forge-std/pull/138#discussion_r956757666, and merge?

ZeroEkkusu:
Catching up 👀

ZeroEkkusu:
Let me clarify, `StdJson.sol` should contain the `stdJson` library as before, and `StdCheatsJson.sol` should contain the `StdCheatsJson` `abstract contract` so we can merge it with StdCheats in the future without breaking changes.

mds1:
> Let me clarify, `StdJson.sol` should contain the `stdJson` library as before

Makes sense.

> and `StdCheatsJson.sol` should contain the `StdCheatsJson` `abstract contract` so we can merge it with StdCheats in the future without breaking changes.

But what's this? Duplicate functionality as an abstract contract instead of a library?

ZeroEkkusu:
> But what's this? Duplicate functionality as an abstract contract instead of a library?

No, we are going to move the problematic std-cheats there.

ZeroEkkusu:
Essentially, we can get `Test` to compile because the new std-cheats use struct decoding. We are going to move them to a separate component, `StdCheatsJson` and let users import it manually if they want to use it. Does that make sense? @mds1 @odyslam

odyslam:
@ZeroEkkusu to be very clear. We **can** compile with the experimental pragma. The issue we seem to be hitting on **some** solc is decoding `string[]` and `bytes[]`

mds1:
Ok, I think I follow:

```solidity
// Library
function parseRaw(string memory json, string memory key)
function readUint(string memory json, string memory key)
function readUintArray(string memory json, string memory key)
function readInt(string memory json, string memory key)
function readIntArray(string memory json, string memory key)
function readBytes32(string memory json, string memory key)
function readBytes32Array(string memory json, string memory key)
function readString(string memory json, string memory key)
function readStringArray(string memory json, string memory key)
function readAddress(string memory json, string memory key)
function readAddressArray(string memory json, string memory key)
function readBool(string memory json, string memory key)
function readBoolArray(string memory json, string memory key)
function readBytes(string memory json, string memory key)
function readBytesArray(string memory json, string memory key)

// StdCheatsJson

// --- snip: all the custom structs ---
function readEIP1559ScriptArtifact(string memory path)
function rawToConvertedEIPTx1559s(RawTx1559[] memory rawTxs)
function rawToConvertedEIPTx1559(RawTx1559 memory rawTx)
function rawToConvertedEIP1559Detail(RawTx1559Detail memory rawDetail)
function readTx1559s(string memory path)
function readTx1559(string memory path, uint256 index)
function readReceipts(string memory path)
function readReceipt(string memory path, uint index)
function rawToConvertedReceipts(RawReceipt[] memory rawReceipts)
function rawToConvertedReceipt(RawReceipt memory rawReceipt)
function rawToConvertedReceiptLogs(RawReceiptLog[] memory rawLogs)
function bytesToUint(bytes memory b) internal pure returns (uint256){
```

> and let users import it manually if they want to use it. Does that make sense

This is the part I was missing, I didn't realize we still had this requirement

odyslam:
@mds1 even with the flag, we seem not being able to compile for 0.7.6 due to the types I mentioned above. Didn't go into why.

ZeroEkkusu:
Lemme check.

mds1:
Ahh I thought it was a test/runtime error, not a compilation error

ZeroEkkusu:
So, it works with `ABIEncoderV2`. I just tried decoding `string[]` inside a struct with `0.7.6` and it works as well. @odyslam

odyslam:
My solc failed to compile. What the actual  F

ZeroEkkusu:
I was experimenting manually. Going to check the PR now and compile it myself.

odyslam:
Try to build from 7daae38aa5261e8a50f5ce5cace7670c0577b5ab

ZeroEkkusu:
Ok, I see with `0.7.6`. Looking into it.

ZeroEkkusu:
@odyslam It's because it missing `pragma experimental ABIEncoderV2;`.

ZeroEkkusu:
So, let's:

- revert the last commit
- add `pragma experimental ABIEncoderV2;` to `StdJson.sol`
- rename the library to `stdJson` and correct the name in `StdCheatsTest`
- address https://github.com/foundry-rs/forge-std/pull/138#discussion_r956757666
- merge

odyslam:
@ZeroEkkusu I can't merge

mds1:
Love to see it
image


Thanks so much @odyslam for getting this done, thanks @ZeroEkkusu for your help with review. @ZeroEkkusu this LGTM, feel free to squash merge if you wanted to do a final check

odyslam:
Next steps:
- Address Artifact parsing
- Address LegacyTxs (transaction artifacts produced by non 1559 networks)

src/StdJson.sol
```diff
@@ -0,0 +1,117 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.6.0 <0.9.0;
```
ZeroEkkusu, line 2:
```suggestion
pragma solidity >=0.6.0 <0.9.0;
pragma experimental ABIEncoderV2;
```

ZeroEkkusu:
The reason CI didn't fail was because it is configured to use `0.7.0`. Now, everything's compiling.

Great work, team!

Issue #139, opened 2022-07-26
`v0.3` changes


ZeroEkkusu:
#### Required changes

- [ ] Check https://github.com/foundry-rs/forge-std/commit/2c7cbfc6fbede6d7c9e6b17afe997e3fdfe22fef and later commits for any necessary changes to make them `v0.3`-ready
- [x] Relicense **(decided)**
    - https://github.com/foundry-rs/forge-std/issues/132

#### Proposed changes

*See source for description*

- [x] Modularize Forge Std
    - PR: #126 
- [x] Allow only "safe" cheatcodes in scripts
    - [source](https://github.com/foundry-rs/forge-std/pull/126#issuecomment-1185526976)
    - https://github.com/foundry-rs/forge-std/issues/78
- [x] ~~CapWords and matching filenames~~
    - [source](https://github.com/foundry-rs/forge-std/pull/126#discussion_r924699853)
    - https://github.com/foundry-rs/forge-std/issues/13
- [ ] Bump pragma to `0.6.2`
    - [source](https://github.com/foundry-rs/forge-std/pull/147#discussion_r944974402)
- [x] Rename components to `Std<Component>`
    - [source](https://github.com/foundry-rs/forge-std/pull/147#issuecomment-1214184147)

#### Related issues

- https://github.com/foundry-rs/forge-std/issues/62

#### Should these issues be resolved?

- https://github.com/foundry-rs/forge-std/issues/118
- https://github.com/foundry-rs/forge-std/issues/127

<br>

Cc @mds1 @paulrberg
Feel free to edit this post.

paulrberg:
> Feel free to edit this post.

It looks like I don't have permission to edit it (perhaps because I'm not part of the `foundry-rs` org). I would suggest editing the top of the post to be a [task list](https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists), and add #126 as a checked off item.

> Should these issues be resolved?

I don't have strong opinions on #127 (it would be nice to have, but not necessary), but I think #118 has become even more topical now in the context of the recent modularization we've built into Forge Std.

If Forge's future will be glorious, there will a large developer ecosystem developed around it. There will be many forks and remixes of the `Assertions ,`Cheats`, etc. components offered by this repo, and similarly people will experiment with different flavors of `Test`.

[...]

PR #147, opened 2022-08-07
feat: make `Script` safer

### LINKED ISSUE #78: refactor: add an abstract `Script` contract

ZeroEkkusu:
## Motivation

Fix #78 

**Note**: Merging into `v0.3`

### Changes

- Add `stdStorageSafe` without `checked_write`
- Add `CheatsSafe` with only `deployCode`
- Add `VmSafe` with only read-only cheatcodes
- Update `Script`
- Make all components `abstract`
- Update `deployCode` tests

### Notes

- `stdStorage` delegates some functions to `stdStorageSafe`. This solution works for Solidity 0.6.0 and requires minimal code duplication:

```solidity
stdStorageSafe {
    safeFun() {
        implementation
    }
}

stdStorage {
    safeFun() {
        return stdStorageSafe.safeFun()
    }

    unsafeFun() {
        implementation
    }
}
```

- Interfaces cannot inherit in Solidity 0.6.0, so, `VmSafe` contains duplicate code. Someone should check if I removed any necessary cheatcodes in `VmSafe`.

## Breaking changes

- Only Cheats from `CheatsSafe` will be accessible in `Script` by default
- Only cheatcodes from `VmSafe` will be accessible from `vm` in `Script` by default

<br>

Link #139

### DIFF
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -2,6 +2,108 @@
 pragma solidity >=0.6.0 <0.9.0;
 pragma experimental ABIEncoderV2;
 
+interface VmSafe {
+    struct Log {
+        bytes32[] topics;
+        bytes data;
+    }
+
+    // Loads a storage slot from an address (who, slot)
+    function load(address,bytes32) external returns (bytes32);
+    // Signs data, (privateKey, digest) => (v, r, s)
+    function sign(uint256,bytes32) external returns (uint8,bytes32,bytes32);
+    // Gets the address for a given private key, (privateKey) => (address)
+    function addr(uint256) external returns (address);
+    // Gets the nonce of an account
+    function getNonce(address) external returns (uint64);
+    // Performs a foreign function call via the terminal, (stringInputs) => (result)
+    function ffi(string[] calldata) external returns (bytes memory);
+    // Sets environment variables, (name, value)
+    function setEnv(string calldata, string calldata) external;
+    // Reads environment variables, (name) => (value)
+    function envBool(string calldata) external returns (bool);
+    function envUint(string calldata) external returns (uint256);
+    function envInt(string calldata) external returns (int256);
+    function envAddress(string calldata) external returns (address);
+    function envBytes32(string calldata) external returns (bytes32);
+    function envString(string calldata) external returns (string memory);
+    function envBytes(string calldata) external returns (bytes memory);
+    // Reads environment variables as arrays, (name, delim) => (value[])
+    function envBool(string calldata, string calldata) external returns (bool[] memory);
+    function envUint(string calldata, string calldata) external returns (uint256[] memory);
+    function envInt(string calldata, string calldata) external returns (int256[] memory);
+    function envAddress(string calldata, string calldata) external returns (address[] memory);
+    function envBytes32(string calldata, string calldata) external returns (bytes32[] memory);
+    function envString(string calldata, string calldata) external returns (string[] memory);
+    function envBytes(string calldata, string calldata) external returns (bytes[] memory);
+    // Expects an error on next call
+    function expectRevert(bytes calldata) external;
+    function expectRevert(bytes4) external;
+    function expectRevert() external;
+    // Records all storage reads and writes
+    function record() external;
+    // Gets all accessed reads and write slot from a recording session, for a given address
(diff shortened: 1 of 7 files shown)


src/Vm.sol
```diff
@@ -2,6 +2,108 @@
...
+    function recordLogs() external;
+    // Gets all the recorded logs, () => (logs)
+    function getRecordedLogs() external returns (Log[] memory);
+    // Derive a private key from a provided mnenomic string (or mnenomic file path) at the derivation path m/44'/60'/0'/0/{index}
+    function deriveKey(string calldata, uint32) external returns (uint256);
+    // Derive a private key from a provided mnenomic string (or mnenomic file path) at the derivation path {path}{index}
+    function deriveKey(string calldata, string calldata, uint32) external returns (uint256);
+}
+
 interface Vm {
```
gakonst, line 107: Should this inherit from `VmSafe`? Or it cannot because of older solidity?

gakonst (approved):

src/Vm.sol
mds1: Looks like interface inheritance was introduced in 0.6.2 https://github.com/ethereum/solidity/blob/develop/Changelog.md#062-2020-01-27

mds1:
PR looks good to me overall 👌

> Only Cheats from `CheatsSafe will be accessible in `Script` by default
> Only cheatcodes from `VmSafe` will be accessible from `vm` in `Script` by default

Where do you inherit from if you want all cheats in a script?

src/Vm.sol
mds1: Looks like the pragma on master is `pragma solidity >=0.6.0 <0.9.0;`. Based on the changelog, I'd be surprised if anyone needed 0.6.0 or 0.6.1 and can't bump to 0.6.2, so I think bumping that pragma to enable interface inheritance is safe.

I know some Maker repos are on 0.6.x, but from a quick search it looks like they use `^0.6.12` so no issues there:
- https://github.com/makerdao/dss-psm/search?q=pragma
- https://github.com/makerdao/dss/search?q=pragma

ZeroEkkusu:
> Where do you inherit from if you want all cheats in a script?

For Cheats:

```solidity
contract MyScript is Script, Cheats {
```

For cheatcodes:

```solidity
contract MyScript is Script {
      Vm internal constant VM = Vm(VM_ADDRESS);
```

You cannot name it `vm`, so this is up for discussion.

mds1:
Got it, seems reasonable. Also should we rename to `StdCheats[Safe]` and `StdAssertions`? That naming fits with stdStorage, stdMath, etc., and makes it easier to different StdCheats (forge-std helpers) from cheatcodes (native forge cheatcodes), whereas IMO "Cheats" vs "cheatcodes" is an easy-to-confuse distinction. The files could all be renamed accordingly too for clarity (open to doing this in a separate PR, just thought of it since the above was a bit confusing)

ZeroEkkusu:
I'll add those ideas to our todo.

Theghost256 (approved):

PR #152, opened 2022-08-14
assertApproxEqRel error message on zero delta

aminlatifi:
`err` message was not passed to `assertEq` in `assertApproxEqRel` method. These two methods are equivalent when `maxPercentDelta` is zero.


ZeroEkkusu (approved):
Thanks!

PR #154, opened 2022-08-17
feat(stdCheats): Add runScript utility function

devanonon:
Finally getting around to writing a little blog post on the importance of testing your scripts and as part of it I had a lot of:
```solidity
Script script = new Script();
script.setUp();
script.run();
```
in my tests. I figure this utility function in `forge-std` will be helpful for others too. It means we can just write:
```solidity
runScript("Script.sol");
```

### DIFF
--- a/src/Test.sol
+++ b/src/Test.sol
@@ -243,10 +243,10 @@ abstract contract Test is DSTest, Script {
         address scriptAddress = deployCode(what);
     
         // Allow setUp() to fail in case it does not exist
-        scriptAddress.call(abi.encodeWithSignature("setUp()"));
+        (bool success, ) = scriptAddress.call(abi.encodeWithSignature("setUp()"));
 
         // Ensure that run() succeeds
-        (bool success, ) = scriptAddress.call(abi.encodeWithSignature("run()"));
+        (success, ) = scriptAddress.call(abi.encodeWithSignature("run()"));
         require(success, "Test runScript(string): call to run() failed in script.");
     }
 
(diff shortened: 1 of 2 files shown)


mds1:
I haven't written any complex standalone scripts that need tests, so my approach has usually been to have my tests inherit from my deploy script and run my protocol's test suite against it. But of course that doesn't always apply, so I can see the use case here. 

One question I have is: Why `new Script()` instead of `MyTest is MyScript`, i.e. why not just inherit from the script and test that way? (Not implying inheritance is better, just curious on why you took this approach)

Also this only supports `run()` with no args, but scripts can be run with any sig and args, so we'd probably want some overloads to account for that

devanonon:
Hey @mds1, thanks for the review!

The use case for `new Script()` over `MyTest is MyScript` is that I like to test multiple script invocations to get to a known state. i.e. I have more than one "deploy" script as part of a pipeline of on-chain actions which can be used. I also have one script which can get a local node to a known state with one command.

It's probably overkill 😄 

Anyway, I settled on:
```solidity
Contract contract = new DeployScript().setUp().run();
new UpdateParametersScript().setUp().run(contract);
```
as I get type safety and it's still quite succinct on one line per script.

I basically get `setUp` to return the contract instance so that I can chain the method calls and I don't think there's now any need for `runScript`.

mds1:
Ah, I like what you settled on. I'll close this since it's no longer needed and arguably too specific towards `run()` with no args in its current form.

You should submit a PR to the [scripting](https://book.getfoundry.sh/tutorials/solidity-scripting) page on the foundry book about how to test your scripts, I don't think there's any content on that currently 😁

PR #157, opened 2022-08-20
feat: rebrand components

ZeroEkkusu:
## Motivation

Rename Forge Std **components** as suggested in https://github.com/foundry-rs/forge-std/pull/147#issuecomment-1214184147

## Solution

Rename the contracts and files to `Std<Component>`


.gitignore
```diff
@@ -1,4 +1,5 @@
 cache/
 out/
 .vscode
-.idea
\ No newline at end of file
+.idea
+.DS_Store
```
mds1, line 5: Recommend adding `.DS_Store` to your global `.gitignore` instead, since this is commonly generated on macOS

src/StdStorage.sol
```diff
@@ -212,12 +212,6 @@ library stdStorage {
         return stdStorageSafe.sigs(sigStr);
     }
 
-    /// @notice find an arbitrary storage slot given a function sig, input data, address of the contract and a value to check against
-    // slot complexity:
-    //  if flat, will be bytes32(uint256(uint));
-    //  if map, will be keccak256(abi.encode(key, uint(slot)));
-    //  if deep map, will be keccak256(abi.encode(key1, keccak256(abi.encode(key0, uint(slot)))));
-    //  if map struct, will be bytes32(uint256(keccak256(abi.encode(key1, keccak256(abi.encode(key0, uint(slot)))))) + structFieldDepth);
```
mds1, line 220: Why delete these comments? They seem useful to keep

ZeroEkkusu: This `find` only routes the `JUMP` to the `find` in `stdStorageSafe`. There are already comments on line 33.

.gitignore
ZeroEkkusu: Thanks!

src/StdStorage.sol
mds1: Ah fair enough, thanks!

mds1 (approved):

PR #168, opened 2022-09-01
feat: add more uint assertions

### LINKED ISSUE #165: support more uint types in assert
I would like to PR some `assert` functions for `uint32`, which I use all the time, but I have to cast to uint256 to be asserted.

- We can have `assert` overloading
- We can have new functions that do the casting, as in [DSTestPlus](https://github.com/transmissions11/solmate/blob/main/src/test/utils/DSTestPlus.sol)

Opening this issue to align first, before diving into the PR

odyslam:
- Fixes #165 
- Adds `foundry.toml` to address latest `allowed_paths` error that originates from https://github.com/foundry-rs/foundry/pull/3007

### DIFF
--- a/src/Test.sol
+++ b/src/Test.sol
@@ -330,19 +330,7 @@ abstract contract Test is DSTest, Script {
         }
     }
 
-    function assertUint128Eq(uint128 a, uint128 b) internal {
-        assertEq(uint256(a), uint256(b));
-    }
-
-    function assertUint64Eq(uint64 a, uint64 b) internal {
-        assertEq(uint256(a), uint256(b));
-    }
-
-    function assertUint96Eq(uint96 a, uint96 b) internal {
-        assertEq(uint256(a), uint256(b));
-    }
-
-    function assertUint32Eq(uint32 a, uint32 b) internal {
+    function assertEqUint(uint256 a, uint256 b) internal {
         assertEq(uint256(a), uint256(b));
     }
 
--- a/src/Test.sol
+++ b/src/Test.sol
@@ -330,6 +330,22 @@ abstract contract Test is DSTest, Script {
         }
     }
 
+    function assertUint128Eq(uint128 a, uint128 b) internal {
+        assertEq(uint256(a), uint256(b));
+    }
+
+    function assertUint64Eq(uint64 a, uint64 b) internal {
+        assertEq(uint256(a), uint256(b));
+    }
+
+    function assertUint96Eq(uint96 a, uint96 b) internal {
+        assertEq(uint256(a), uint256(b));
+    }
+
+    function assertUint32Eq(uint32 a, uint32 b) internal {
+        assertEq(uint256(a), uint256(b));
+    }
+
     function assertApproxEqAbs(
         uint256 a,
         uint256 b,
--- a/src/test/StdCheats.t.sol
+++ b/src/test/StdCheats.t.sol
@@ -248,10 +248,10 @@ contract StdCheatsTest is Test {
     }
 
     function testAssertions() public {
-        assertUint32Eq(uint32(1), uint32(1));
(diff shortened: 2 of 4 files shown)


src/test/StdCheats.t.sol
```diff
@@ -247,6 +247,13 @@ contract StdCheatsTest is Test {
         Receipt[] memory receipts = readReceipts(path);
     }
 
+    function testAssertions() public {
+        assertUint32Eq(uint32(1), uint32(1));
+        assertUint64Eq(uint64(1), uint64(1));
+        assertUint96Eq(uint96(1), uint96(1));
+        assertUint128Eq(uint128(1), uint128(1));
```
mds1, line 254: I was going to comment on changing all the names to `assertEq`, then realized solc errors due to ambiguity. So then I deleted the 4 new assertions and changed these 4 tests to just use `assertEq`, and as expected solc upcasts them to uint256 using the existing assertions. So now I'm struggling to remember why we need these 😅

To clarify, if you remove the changes in `Test.sol`, all of these pass:

```solidity
function testAssertions() public {
    assertEq(uint32(1), uint32(1));
    assertEq(uint64(1), uint64(1));
    assertEq(uint96(1), uint96(1));
    assertEq(uint128(1), uint128(1));
}
```

odyslam: While coding in 0.7.6, I have to explicitly cast the uint32 to uint256 to use assert.

odyslam: Didn't know that they are automatically upcasted in >0.8 . It's been so long in 0.7.0 land

mds1: Heh sorry to hear that 😛

mds1 (approved):
This LGTM, @ZeroEkkusu would love a quick review to make sure there's nothing that affect v0.3 branch here / anything else blocking merge

src/Test.sol
```diff
@@ -330,6 +330,22 @@ abstract contract Test is DSTest, Script {
...
+    }
+
+    function assertUint96Eq(uint96 a, uint96 b) internal {
+        assertEq(uint256(a), uint256(b));
+    }
+
+    function assertUint32Eq(uint32 a, uint32 b) internal {
+        assertEq(uint256(a), uint256(b));
+    }
+
```
ZeroEkkusu, line 348:
```suggestion
    function assertEqUint(uint256 a, uint256 b) internal {
        assertEq(a, b);
    }
```

src/test/StdCheats.t.sol
```diff
@@ -247,6 +247,13 @@ contract StdCheatsTest is Test {
...
+    }
+
```
ZeroEkkusu, line 256:
```suggestion
```

src/Test.sol
odyslam: Not sure what you suggest @ZeroEkkusu

ZeroEkkusu: Sorry, @odyslam, accidentally clicked `Add single comment`. I'll write up a review real quick.

ZeroEkkusu (changes requested):
Solidity versions `<=0.8.0` have a problem differentiating `uint` and `int` functions when an argument is of type `uintX`, where `X` is less than `256`. This does not apply to `int` arguments.

I suggest we add a single function, `assertEqUint` that can be universally used for these types. Let me know what you think.

odyslam:
sounds good

ZeroEkkusu (approved):
LGTM. Thank you!

It is worth noting that we can also mix the types; e.g. this works:

```solidity
assertEqUint(uint8(5), uint128(5));
```

<br>

One nit:

Tests should be in `StdAssertions.t.sol` (right now they're in `StdCheats.t.sol`).

If you want, I'll move the tests.

PR #171, opened 2022-09-06
feat: add rememberKey cheat code

devanonon:
Re https://github.com/foundry-rs/foundry/pull/3047/

### DIFF
--- a/src/Script.sol
+++ b/src/Script.sol
@@ -37,7 +37,8 @@ abstract contract Script {
         return address(uint160(uint256(bytesValue)));
     }
 
-    function deriveRememberKey(string memory mnemonic, uint32 index) internal returns (address) {
-        return vm.rememberKey(vm.deriveKey(mnemonic, index));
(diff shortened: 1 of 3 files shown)


src/Script.sol
```diff
@@ -36,4 +36,8 @@ abstract contract Script {
     function addressFromLast20Bytes(bytes32 bytesValue) internal pure returns (address) {
         return address(uint160(uint256(bytesValue)));
     }
+
+    function deriveRememberKey(string memory mnemonic, uint32 index) internal returns (address) {
+        return vm.rememberKey(vm.deriveKey(mnemonic, index));
+    }
```
mds1, line 42: Wdyt about also returning the private key as the second arg? That way if people want it for e.g. signatures or other reasons it's available

devanonon: Yeah I was really 50/50 on this one, but I guess a tuple is nicer because if you don't want the private key you can easily just ignore it. Will have examples in the book for this too (PR nearly ready)!

mds1: Yep that was my thinking too, thanks a lot!

PR #176, opened 2022-09-14
feat: add package.json

smartcontracts:
Adds a package.json file so forge-std can be properly imported as a git dependency.


mds1:
Thanks!

PR #178, opened 2022-09-18
feat: add getDeployedCode cheatcode

mattsse:
Add new `getDeployedCode` cheatcode

book https://github.com/foundry-rs/book/pull/624

foundry: https://github.com/foundry-rs/foundry/pull/3248


mds1 (approved):
Thanks!

PR #180, opened 2022-09-20
feat: add new fork cheatcodes

mds1:
Ref https://github.com/foundry-rs/foundry/pull/3262


mds1:
@mattsse I copied this straight from your PR so should be good, but lmk if anything is missing/copied wrong

PR #183, opened 2022-09-21
ci: add forge fmt job

mattsse:
integrate `forge fmt --check` in CI

ran `forge fmt`

PR #184, opened 2022-09-21
feat: forge-std v1.0.0

ZeroEkkusu:
# Forge Standard Library V1

Forge-Std is now modular - enabling composability, and paving the way for new libraries for testing in Solidity.

## Features

- #126 (includes `createCompute2Address` and standards)
- [Smart `bound`](https://github.com/foundry-rs/forge-std/pull/184/commits/7d14c38b8db4ab578af347e779dddbb6fa02735e)
- #193
- https://github.com/foundry-rs/forge-std/pull/195

### Breaking changes

- Bump pragma to `>=0.6.2 <0.9.0`
- Remove deprecated features (`tip` std-cheat, `lowLevelError` std-error)
- Disallow "unsafe" std-cheats in scripts by default https://github.com/foundry-rs/forge-std/pull/147
- `using stdStorage for StdStorage` will break in scripts and will need to be changed to `stdStorageSafe` (or will require a manual import of `stdStorage`)

## Before merging

- [x] Bump pragma to `0.6.2` https://github.com/foundry-rs/forge-std/pull/147#discussion_r944974402
- [x] Check https://github.com/foundry-rs/forge-std/commit/2c7cbfc6fbede6d7c9e6b17afe997e3fdfe22fef and later commits for any necessary changes (e.g. std-cheats should be `internal virtual`)


src/Vm.sol
```diff
@@ -2,6 +2,108 @@
 pragma solidity >=0.6.0 <0.9.0;
 pragma experimental ABIEncoderV2;
 
+interface VmSafe {
```
mds1, line 5: Looks like some cheats from master are missing in this file, such as `allowCheatcodes`

mds1: Ah sorry didn't realize that wasn't done yet, was going to say `makeAddr` is also missing too. Cool let me know once that's ready and I'll take a closer look + ping some people for testing

ZeroEkkusu: I can do it tomorrow (also open for anyone to take).

mds1: New cheatcode was added so needs another merge from main—I'll start testing this tomorrow!

src/Components.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.6.2 <0.9.0;
+
+import "./console.sol";
```
mds1, line 4: _(arbitrary comment location to start a thread)_

I think we might need to fix some input ordering, or use explicit import syntax instead? Just dropped this branch into a repo that works with the current version and got a lot of errors
image

mds1: Seems the issue is that I have scripts that are `is Script`, tests that are `is Test`, and those tests import and run the script which results in duplicate declarations. For example:

```solidity
// script/MyScript.s.sol
import "forge-std/Script.sol";
contract MyScript is Script {
  function run() public {
    // --- snip ---
  }
}

// test/MyTest.t.sol
import "forge-std/Test.sol";
import "script/MyScript.s.sol";

contract MyTest is MyScript, Test {
  function setUp() public {
    // Deploys the system and saves the deploy addrs to storage
    MyScript.run();
  }
}
```

ZeroEkkusu: Is this a common pattern? We can fix this with a common abstract contract, for `TestBase` and `ScriptBase`, but `vm` cannot be `Vm` and `VmSafe` at the same time.

ZeroEkkusu: Thanks for the insight.

What do you think about `vms`? Maybe even as an alias.

mds1: I could be convinced, but initial reaction is that `vm` and `vms` are very similar so it's too easy to mix one up for the other without noticing

ZeroEkkusu: Yeah. `vmSafe` is good for now.

You can add a new file with a common abstract contract locally and inherit it in `TestBase` and `ScriptBase` if you want to continue testing (I'm not on my computer, or you can push the fix if you want).

mds1:
@ZeroEkkusu I tried moving test dir to the root in f21ef1a to match the convention of `forge init` and most projects these days, any idea why that broke CI? Don't have time to check at the moment, so pushed another commit to revert that for now

ZeroEkkusu:
It's as if `--contracts src/Test.sol` in `forge build --contracts src/Test.sol --use solc:0.8.0` has no effect, and it compiles all contracts.

mds1:
Ah makes sense. There is now a `--skip` flag (https://github.com/foundry-rs/foundry/pull/3370) so we should be able to `forge build --skip test --use solc:0.8.0` instead now. I should be able to get to this later today

mds1:
@ZeroEkkusu wdyt of also running `forge fmt` + adding to CI (per https://github.com/foundry-rs/forge-std/pull/183) in this PR? That way once it's merged v1.0.0 is done

ZeroEkkusu:
Sure @mds1.

mds1:
- Took the default `fmt` values and defined them explicitly in `foundry.toml` (if left implicit, a contributor with their own config in `~/.foundry/foundry.toml` would end up applying their config instead of the default)
- Ran `forge fmt`
- Split CI into separate jobs since now there's a few things going on

I think all that's left before merge is some more testing. Just tested the latest commit on this branch in some repos and it worked great. Would love for @nventuro to test since he has an 0.7.6 repo

nventuro:
Thanks for the ping @mds1! Just run our tests using the v0.3 branch and everything seems to work correctly.

nventuro:
Actually, after further testing I'm seeing some changes in behavior when using commit [dc8d830](https://github.com/foundry-rs/forge-std/pull/184/commits/dc8d830d3af6f67758f793ce42572270fb0f41fb) when compared to 9310e879db8ba3ea6d5c6489a579118fd264a3f5. 

[`testExitSwaps`](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/pool-weighted/test/foundry/WeightedMath.t.sol#L105) fails with too many `vm.assume` rejects. I imagine that could be due to `bound` behaving differently, reuslting in a different random variable distribution.

[`testUpdateCachedRatios`](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol#L120) also fails, though this one is more bizarre. The assertion that fails is a [comparison of gas costs](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol#L166), which seems to be unrelated to `forge-std` entirely as we only call our own code inside the `gasleft()` blocks.

I've switched back and forth between commits dc8d830d3af6f67758f793ce42572270fb0f41fb (this branch) and 9310e879db8ba3ea6d5c6489a579118fd264a3f5, and consistently get errors only on the updated commit.

CarterAppleton:
tested in our 0.8.16 and 0.6.12 project, works great 👍 looking forward to this

src/StdUtils.sol
```diff
@@ -10,17 +10,29 @@ abstract contract StdUtils {
...
         require(min <= max, "Test bound(uint256,uint256,uint256): Max is less than min.");
 
+        // If x is between min and max, return x directly. This is to ensure that dictionary values
+        // do not get shifted if the min is nonzero. More info: https://github.com/foundry-rs/forge-std/issues/188
+        if (x >= min && x <= max) return x;
+
+        // Handle degenerate cases.
         uint256 size = max - min;
+        if (size == 0) return min;
+        if (size == UINT256_MAX) return x;
```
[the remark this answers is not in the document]
ZeroEkkusu: Line 20 only.

ZeroEkkusu:
It’s a very good solution @mds1! You should probably PR this to Solmate. Cc @transmissions11.

src/StdUtils.sol
```diff
@@ -10,17 +10,29 @@ abstract contract StdUtils {
...
+
+        // If the value is 0,1,2 wrap that to min, min+1, min+2. Similarly, for the MAX_UINT side.
+        // This helps ensure coverage of the min/max values. It does introduce a small bias, however
+        // uint256 is so large that this is insignificant.
+        if (x == 0) return min;
+        if (x == 1 && size >= 1) return min + 1;
+        if (x == 2 && size >= 2) return min + 2;
+        if (x == UINT256_MAX) return max;
+        if (x == UINT256_MAX - 1 && size >= 1) return max - 1;
+        if (x == UINT256_MAX - 2 && size >= 2) return max - 2;
```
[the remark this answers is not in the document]
mds1: Good catch, fixed in 1959a93

src/StdUtils.sol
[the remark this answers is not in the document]
mds1: Should be fixed in 1959a93, though @ZeroEkkusu not sure what you mean by "Line 20 only"?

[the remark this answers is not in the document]
ZeroEkkusu: @mds1 Line 19 was included in the preview, so I wanted to point out it applied only to line 20.

mds1:
> Actually, after further testing I'm seeing some changes in behavior when using commit [dc8d830](https://github.com/foundry-rs/forge-std/pull/184/commits/dc8d830d3af6f67758f793ce42572270fb0f41fb) when compared to f36dab2 
> 
> [`testExitSwaps`](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/pool-weighted/test/foundry/WeightedMath.t.sol#L105) fails with too many `vm.assume` rejects. I imagine that could be due to `bound` behaving differently, reuslting in a different random variable distribution.
> 
> [`testUpdateCachedRatios`](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol#L120) also fails, though this one is more bizarre. The assertion that fails is a [comparison of gas costs](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/pool-weighted/test/foundry/CircuitBreakerLib.t.sol#L166), which seems to be unrelated to `forge-std` entirely as we only call our own code inside the `gasleft()` blocks.
> 
> I've switched back and forth between commits dc8d830d3af6f67758f793ce42572270fb0f41fb (this branch) and f36dab2, and consistently get errors only on the updated commit.

@nventuro I was able to reproduce those failures. Some new failures due to the different/improved bound behavior is a good thing since it confirms this is now working better. However, if I keep forge-std on this branch but overwrite `bound` to use the old version, I still see the failures, so it seems `bound` isn't the cause (or maybe is partly the cause) 🤔

mds1:
@nventuro Had some time on a flight without wifi so figured this was a good offline debug task. Here's what I found:

### testUpdateCachedRatios

Below is a concrete test that will pass with forge-std on master but fail with v0.3 (well, v1.0.0 technically but calling it v0.3 since that's this branch name).

The difference is in the `newWeight = bound(newWeight, _MINIMUM_BOUND_PERCENTAGE, FixedPoint.ONE);` line.
The first four parameters that we bound all have the same output value on master and v0.3. However, the last parameter, `newWeight`, differs between v0.3 and master. Since `newWeight = type(uint256).max` in v0.3 it now gets wrapped to the upper bound of the range, which is `FixedPoint.ONE`. On master it was wrapped to 477686011984099564. Presumably this results in a different code path or different storage values since the dynamic costs vary significantly as shown below:

```
cachedCost, master:  2534
cachedCost, v0.3:    2526

dynamicCost, master: 7702
dynamicCost, v0.3:   2892
```

If I change the upper bound for `newWeight` from `FixedPoint.ONE` to `FixedPoint.ONE-1` the test (including the fuzz variant) always passes on v0.3. I'm not sure if this is logical/correct in this test, however based on these findings I don't think there is an issue with `bound` as the results are valid, the test probably just needs to be updated to account for the new bound behavior.


```solidity
function testUpdateCachedRatiosFailure() public {
    // Hardcode inputs from the failure case found by the fuzzer on the v0.3 branch.
    uint256 bptPrice = 0;
    uint256 referenceWeight = 0;
    uint256 newWeight = type(uint256).max;
    uint256 lowerBound = 0;
    uint256 upperBound = 0;

    // Pass them through bound.
    bptPrice = bound(bptPrice, _MIN_BPT_PRICE, _MAX_BPT_PRICE);
    lowerBound = bound(lowerBound, _MINIMUM_BOUND_PERCENTAGE, FixedPoint.ONE);
    upperBound = bound(upperBound, lowerBound, _MAX_BOUND_PERCENTAGE);
    referenceWeight = bound(referenceWeight, _MINIMUM_TOKEN_WEIGHT, _MAXIMUM_TOKEN_WEIGHT);
    newWeight = bound(newWeight, _MINIMUM_BOUND_PERCENTAGE, FixedPoint.ONE);

    // Log them for convenience.
    console2.log("bptPrice", bptPrice);
    console2.log("lowerBound", lowerBound);
    console2.log("upperBound", upperBound);
    console2.log("referenceWeight", referenceWeight);
    console2.log("newWeight", newWeight);

    // Convert the bound statements in the fuzz test to assumes as a sanity check that 
    // the bound outputs are valid.
    vm.assume(bptPrice >= _MIN_BPT_PRICE && bptPrice <= _MAX_BPT_PRICE);
    vm.assume(lowerBound >= _MINIMUM_BOUND_PERCENTAGE && lowerBound <= FixedPoint.ONE);
    vm.assume(upperBound >= lowerBound && upperBound <= _MAX_BOUND_PERCENTAGE);
    vm.assume(referenceWeight >= _MINIMUM_TOKEN_WEIGHT && referenceWeight <= _MAXIMUM_TOKEN_WEIGHT);
    vm.assume(newWeight >= _MINIMUM_BOUND_PERCENTAGE && newWeight <= FixedPoint.ONE);

    // Call the fuzz test with our parameters.
    testUpdateCachedRatios(bptPrice, referenceWeight, newWeight, lowerBound, upperBound);
}
```

### testExitSwaps

Similar to above, the new `bound` behavior just results in more test rejections.

On average I get around 70% of the way through the 10k runs before erroring with too many rejects. (The 60k reject limit you currently have is slightly less than the default of 65536). So one solution is to simply bump the max rejects limit and deal with longer test times. I tried setting `max_test_rejects = 100000` and it seems to consistently pass. On master this test passes in about 10 seconds, on v0.3 it'll fail after ~35s, and with the 100k rejects it passes after ~50s.

An alternative idea is that you may be able to get a bit more clever with how you use `bound` to ensure the requirement of `amountOut.divUp(FixedPoint.ONE - swapFeePercentage) <= balances[tokenIndex]` always holds. Modifying the `swapFeePercentage` bound seems like the easiest way to do this, e.g. something like this (not certain it's correct, make sure to verify if you go this route):

```solidity
amountOut = bound(amountOut, 0, balances[tokenIndex]);
bptTotalSupply = bound(bptTotalSupply, _DEFAULT_MINIMUM_BPT, type(uint112).max);
swapFeePercentage = bound(swapFeePercentage, 0, 0.99e18);

// This exit type is special in that fees are charged on the amount out. This creates scenarios in which the
// total amount out (including fees) exceeds the Pool's balance, which will lead to reverts. We reject any runs
// that result in this edge case.
bool isValid = amountOut.divUp(FixedPoint.ONE - swapFeePercentage) <= balances[tokenIndex];
if (!isValid) {
    // To speed up tests by minimizing rejections, modify the swapFeePercentage to ensure
    // the validity check holds.
    swapFeePercentage = balances[tokenIndex] - amountOut - 1;
    isValid = amountOut.divUp(FixedPoint.ONE - swapFeePercentage) <= balances[tokenIndex];
}
vm.assume(isValid);
```

### TLDR

I think the branch is behaving as expected and some tests need to be updated to account for the new behavior

ZeroEkkusu:
Just dropped a new `bound` implementation, @mds1. This one has even number distribution.

For `bound(x, 15, 18)`:

**Previously**

```
15 16 17 18 15 16 17 18 15 16 17 18 15 16 17 15 16 17 18 18 15
--------------------------------------------------------------
 0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18 19 20
```

**Now**

```
15 16 17 18 16 17 18 15 16 17 18 15 16 17 18 15 16 17 18 15 16
--------------------------------------------------------------
 0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18 19 20
```

> 🔎 Notice how `max + 1` is `min` and `min - 1` is `max` with the new implementation:
>
> ```
> 18 15 16 17 18 15
> -----------------
> 14 15 16 17 18 19
> ```

We should test it more in practice to make sure there are no errors.

mds1:
I'll check this out and put a simple script together to test the distributions before and after to make sure things are good / the new complexity (new conditionals) are worth it.

Going forward let's make these changes as PRs into this branch instead of committing directly to it—I think the separation is cleaner and makes it easier to discuss changes without resulting in one giant PR, plus if we decide to not implement a change we just close the PR instead of force pushing or adding commits to revert something

mds1:
Ok @ZeroEkkusu I think this implementation looks good. 

> We should test it more in practice to make sure there are no errors.

Did you plan on writing these? Wondering what additional tests we should add. A bit tedious, but perhaps we hardcode an array of `[0, 1, 2, 3, ..., 48, 49, 50]`, and loop over that array for a few different min/max ranges (or a smaller array so it's more manageable)

ZeroEkkusu:
Good idea, I'll add one.

Thoughts on the merge, after #195?

mds1:
Awesome, thanks.

Once we get #195 merged I want to do one more quick round of testing by pinging a few people again (since we made some additional changes), then good to merge if that goes smoothly

mds1:
Ok, this is _finally_ just about ready to be merged, and just updated the description to link the relevant new PRs. I'm going to test this once more on some repos, @hexonaut @nventuro if you want to test again that would be appreciated

mds1:
Ok just tested in two repos:
- First one had tests all passing
- One had some failures, but these seem ok since they're due to `bound`, and some failures are expected there

hexonaut:
Just tested on one of our larger repos. Works fine.

src/StdCheats.sol
```diff
@@ -0,0 +1,619 @@
...
+        Chain Polygon;
+        Chain PolygonMumbai;
+        Chain Avalanche;
+        Chain AvalancheFuji;
+        Chain BnbSmartChain;
+        Chain BnbSmartChainTestnet;
+        Chain GnosisChain;
+    }
+
+    Chains stdChains;
```
ZeroEkkusu, line 49:
```suggestion
    Chains private stdChains;
```

Additionally, we should consider naming the structs `StdChain(s)`, to not pollute the namespace. Some structs for JSON std-cheats are also problematic (e.g. `Receipt`), but those breaking changes may be unwelcome.

mds1: Renaming to `StdChain(s)` is fine with me if you want to open a PR into this branch for it

But I think we should leave the JSON struct names to avoid breaking changes, and also because I think those are core-enough to Ethereum itself that it's ok they don't have the `Std` prefix

ZeroEkkusu: Resolved; no changes will be made.

src/StdUtils.sol
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.6.2 <0.9.0;
+
+import "./console2.sol";
+
+abstract contract StdUtils {
+    uint256 internal constant UINT256_MAX =
```
ZeroEkkusu, line 7: I think this should be available from `CommonBase` instead from a specific Component.

ZeroEkkusu (commented):
Left one suggestion, otherwise LGTM (Let's Get This Merged)!

PR #187, opened 2022-09-28
Add assertion error for different types

Valentin-Seehausen:
When accidentally mixing uint and int in tests, the console throws a very long output. This PR corrects this by throwing a simple error.

```
error[9322]: TypeError: No matching declaration found after argument-dependent lookup.
  --> test/lib/FundingFee/Curve.t.sol:14:9:
   |
14 |         assertEq(FundingFee.curve(ONE * 1 / 10), 2 * PERCENT, "0.1");
   |         ^^^^^^^^
Note: Candidate:
   --> lib/forge-std/src/Test.sol:212:5:
    |
212 |     function assertEq(bool a, bool b) internal {
    |     ^ (Relevant source part starts here and spans across multiple lines).
Note: Candidate:
   --> lib/forge-std/src/Test.sol:221:5:
    |
221 |     function assertEq(bool a, bool b, string memory err) internal {
    |     ^ (Relevant source part starts here and spans across multiple lines).
```

image

### DIFF
--- a/src/Test.sol
+++ b/src/Test.sol
@@ -329,6 +329,20 @@ abstract contract Test is DSTest, Script {
             assertEq(a, b);
         }
     }
+    
+    function assertEq(uint a, int b) internal {
+        emit log                ("Error: variables have different types [type]");
+        emit log_named_string   ("  Expected int");
+        emit log_named_string   ("    Actual uint");
+        fail();
+    }    
+    
+    function assertEq(int a, uint b) internal {
+        emit log                ("Error: variables have different types [type]");
+        emit log_named_string   ("  Expected uint");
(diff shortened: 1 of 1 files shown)


mds1:
Hmm, I'm not sure we want to merge this for a few reasons:
- That error you're seeing is from solc, so I'm hesitant to suppress it to begin with. It sounds like what you really want is for solc to give a more concise error message. There's lots of other places this error can show up aside from forge-std assertions
- This is a bit contrived to handle the uint/int case by adding fake assertion helpers that intentionally revert. There's lots of other mismatched cases this can happen for, and solc already handles them all for us.

Valentin-Seehausen:
Ok, thanks for the feedback. Closing then.

PR #191, opened 2022-10-07
feat: add assertApproxEqAbsDecimal assertions

aaoferreira:
Adding decimal formatting support for the `assertApproxEqAbs` assertions. Since it's a small one I figured I'd just PR instead of opening an issue, let me know if I need to create one

### DIFF
--- a/src/StdAssertions.sol
+++ b/src/StdAssertions.sol
@@ -124,6 +124,31 @@ abstract contract StdAssertions is DSTest {
         }
     }
 
+    function assertApproxEqAbsDecimal(uint256 a, uint256 b, uint256 maxDelta, uint256 decimals) internal virtual {
+        uint256 delta = stdMath.delta(a, b);
+
+        if (delta > maxDelta) {
+            emit log("Error: a ~= b not satisfied [uint]");
+            emit log_named_decimal_uint("  Expected", b, decimals);
+            emit log_named_decimal_uint("    Actual", a, decimals);
+            emit log_named_decimal_uint(" Max Delta", maxDelta, decimals);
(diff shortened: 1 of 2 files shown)


mds1:
Sorry for the delay here, this slipped by. Do you mind updating this against the latest master? Then I'll review in a more timely manner 😅

Given that we already have `assertEqDecimal`, adding the equivalents for the approxEq analogs seems reasonable, so supportive of this PR

aaoferreira:
@mds1 just rebased it

src/StdAssertions.sol
```diff
@@ -146,6 +171,31 @@ abstract contract StdAssertions is DSTest {
...
+            emit log("Error: a ~= b not satisfied [int]");
+            emit log_named_decimal_int("  Expected", b, decimals);
+            emit log_named_decimal_int("    Actual", a, decimals);
+            emit log_named_decimal_uint(" Max Delta", maxDelta, decimals);
+            emit log_named_decimal_uint("     Delta", delta, decimals);
+            fail();
+        }
+    }
+
+    function assertApproxEqAbsDecimal(int256 a, int256 b, uint256 maxDelta, uint256 decimals, string memory err)
```
mds1, line 187: Do you mind also adding the equivalents for `assertApproxEqRel` too?

mds1 (commented):
This looks great, thanks! Failing CI test is unrelated

src/StdAssertions.sol
aaoferreira: done ✅

mds1 (approved):
Thanks!

PR #193, opened 2022-10-15
feat: add interfaces

### LINKED ISSUE #192: feat: add common interfaces to forge-std
Add ERC-20, 165, 721, 1155, and 4626 interfaces so devs don't need to install OZ or other libs to use them. This matches the set of default interfaces included by vyper: https://github.com/vyperlang/vyper/tree/master/vyper/builtin_interfaces

Note that name and symbol are technically optional in ERC-20, so vyper and OZ both split those into separate files. But I don't think I've seen a token that doesn't have name/symbol, so perhaps we just include that by default in a single IERC20 interface?

h/t @brockelmore

mds1:
Closes #192 

When an EIP had a solidity interface with comments I pasted that directly, otherwise I took the OZ interfaces and truncated them a bit. In all cases I made the natspec style consistent

Includes an unrelated visibility change for bound

Merging into the v0.3 branch since that should get merged soon

### DIFF
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -7,7 +7,7 @@ abstract contract StdUtils {
     uint256 internal constant UINT256_MAX =
         115792089237316195423570985008687907853269984665640564039457584007913129639935;
 
-    function _bound(uint256 x, uint256 min, uint256 max) internal view virtual returns (uint256 result) {
+    function _bound(uint256 x, uint256 min, uint256 max) internal pure virtual returns (uint256 result) {
         require(min <= max, "StdUtils bound(uint256,uint256,uint256): Max is less than min.");
 
         // If x is between min and max, return x directly. This is to ensure that dictionary values
--- a/src/interfaces/IERC1155.sol
+++ b/src/interfaces/IERC1155.sol
@@ -6,26 +6,28 @@ import "src/interfaces/IERC165.sol";
 /// @dev See https://eips.ethereum.org/EIPS/eip-1155
 /// Note: The ERC-165 identifier for this interface is 0xd9b67a26.
 interface ERC1155 is IERC165 {
-    /// @dev Either `TransferSingle` or `TransferBatch` MUST emit when tokens are transferred, including zero value transfers as well as minting or burning (see "Safe Transfer Rules" section of the standard).
-    /// The `_operator` argument MUST be the address of an account/contract that is approved to make the transfer (SHOULD be msg.sender).
-    /// The `_from` argument MUST be the address of the holder whose balance is decreased.
-    /// The `_to` argument MUST be the address of the recipient whose balance is increased.
-    /// The `_id` argument MUST be the token type being transferred.
-    /// The `_value` argument MUST be the number of tokens the holder balance is decreased by and match what the recipient balance is increased by.
-    /// When minting/creating tokens, the `_from` argument MUST be set to `0x0` (i.e. zero address).
-    /// When burning/destroying tokens, the `_to` argument MUST be set to `0x0` (i.e. zero address).
+    /// @dev
+    /// - Either `TransferSingle` or `TransferBatch` MUST emit when tokens are transferred, including zero value transfers as well as minting or burning (see "Safe Transfer Rules" section of the standard).
+    /// - The `_operator` argument MUST be the address of an account/contract that is approved to make the transfer (SHOULD be msg.sender).
+    /// - The `_from` argument MUST be the address of the holder whose balance is decreased.
+    /// - The `_to` argument MUST be the address of the recipient whose balance is increased.
+    /// - The `_id` argument MUST be the token type being transferred.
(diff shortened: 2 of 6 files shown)


src/StdUtils.sol
```diff
@@ -7,7 +7,7 @@ abstract contract StdUtils {
     uint256 internal constant UINT256_MAX =
         115792089237316195423570985008687907853269984665640564039457584007913129639935;
 
-    function _bound(uint256 x, uint256 min, uint256 max) internal view virtual returns (uint256 result) {
+    function _bound(uint256 x, uint256 min, uint256 max) internal pure virtual returns (uint256 result) {
```
ZeroEkkusu, line 10:
```suggestion
    function _bound(uint256 x, uint256 min, uint256 max) private pure virtual returns (uint256 result) {
```

mds1: I figured internal is preferable cause it also enables using `bound` without logging the result

ZeroEkkusu (approved):

src/interfaces/IERC1155.sol
```diff
@@ -0,0 +1,103 @@
...
+    event ApprovalForAll(address indexed _owner, address indexed _operator, bool _approved);
+
+    /// @dev MUST emit when the URI is updated for a token ID.
+    /// URIs are defined in RFC 3986.
+    /// The URI MUST point to a JSON file that conforms to the "ERC-1155 Metadata URI JSON Schema".
+    event URI(string _value, uint256 indexed _id);
+
+    /// @notice Transfers `_value` amount of an `_id` from the `_from` address to the `_to` address specified (with safety call).
+    /// @dev Caller must be approved to manage the tokens being transferred out of the `_from` account (see "Approval" section of the standard).
+    /// MUST revert if `_to` is the zero address.
```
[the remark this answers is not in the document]
mds1: Let me know how the fix in 210d516 looks

src/interfaces/IERC4626.sol
```diff
@@ -0,0 +1,189 @@
+pragma solidity >=0.6.2;
+
+import "src/interfaces/IERC20.sol";
+
+/// @dev Interface of the ERC4626 "Tokenized Vault Standard", as defined in
+/// https://eips.ethereum.org/EIPS/eip-4626[ERC-4626].
```
ZeroEkkusu, line 6:
```suggestion
/// https://eips.ethereum.org/EIPS/eip-4626
```

src/interfaces/IERC4626.sol
```diff
@@ -0,0 +1,189 @@
...
+    /// NOTE: This calculation MAY NOT reflect the “per-user” price-per-share, and instead should reflect the
+    /// “average-user’s” price-per-share, meaning what the average user should expect to see when exchanging to and
+    /// from.
+    function convertToAssets(uint256 shares) external view returns (uint256 assets);
+
+    /// @notice Returns the maximum amount of the underlying asset that can be deposited into the Vault for the receiver,
+    /// through a deposit call.
+    /// @dev
+    /// - MUST return a limited value if receiver is subject to some deposit limit.
+    /// - MUST return 2///* 256 - 1 if there is no limit on the maximum amount of assets that may be deposited.
```
ZeroEkkusu, line 57:
```suggestion
    /// - MUST return 2 ** 256 - 1 if there is no limit on the maximum amount of assets that may be deposited.
```

src/interfaces/IERC4626.sol
```diff
@@ -0,0 +1,189 @@
...
+    /// - MUST revert if all of assets cannot be deposited (due to deposit limit being reached, slippage, the user not
+    ///   approving enough underlying tokens to the Vault contract, etc).
+    ///
+    /// NOTE: most implementations will require pre-approval of the Vault with the Vault’s underlying asset token.
+    function deposit(uint256 assets, address receiver) external returns (uint256 shares);
+
+    /// @notice Returns the maximum amount of the Vault shares that can be minted for the receiver, through a mint call.
+    /// @dev
+    /// - MUST return a limited value if receiver is subject to some mint limit.
+    /// - MUST return 2///* 256 - 1 if there is no limit on the maximum amount of shares that may be minted.
```
ZeroEkkusu, line 90:
```suggestion
    /// - MUST return 2 ** 256 - 1 if there is no limit on the maximum amount of shares that may be minted.
```

src/interfaces/IERC1155.sol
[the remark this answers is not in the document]
ZeroEkkusu: @mds1 Nice, especially with multi-line descriptions. There are more different patterns (e.g. in `IERC721.sol`) but it's not a problem.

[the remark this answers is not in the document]
mds1: Yea, all the ERC's use different styles which made this a bit annoying haha. I think this is probably fine for now as it really only matters for consistent docgen output once `forge doc` is merged, so we can revisit at that time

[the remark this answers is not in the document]
mds1: Feel free to squash merge this into v0.3 once you're done reviewing 🙂

Issue #134, opened 2022-07-19
feat: `noPrecompiles` modifier or function


mds1:
Ref https://github.com/foundry-rs/foundry/issues/1228

The fuzzer will pass precompile addresses to tests. These are valid inputs so we don't want to strip them from the fuzzer, but in some tests they can cause problems. The scope of this issue is to add a modifier OR a function (I'm not yet sure which we should prefer, open to discussion here, as well a name other than `noPrecompiles`) that does something like this:

```solidity
// --- forge-std ---
// Use this version to infer the chain ID from the test environment.
function noPrecompiles(address who) internal {
  noPrecompiles(who, block.chainid);
}

// Use this version to specify which chain's precompiles to use.
function noPrecompiles(address who, uint256 chainId) internal {
  if (chainid == 1 || chainid == 31337) {
    // Mainnet precompiles: https://www.evm.codes/precompiled
    vm.assume(who > address(9));
  } else if (chainid == 10 || chainid == 69 || chainid == 420) {
    // Optimism precompiles: https://github.com/ethereum-optimism/optimism/blob/master/packages/contracts/contracts/libraries/constants/Lib_PredeployAddresses.sol
    vm.assume(who < 0x4200000000000000000000000000000000000000 && who > 0x4200000000000000000000000000000000000013);
  }
  // etc. for other chains
}

// --- myTest.t.sol ---
function testSomeAddress(address a, address b) public {
  noPrecomples(a); // Precompiles for `b` are allowed.
}
```

I think including mainnet, optimism, and arbitrum precompiles is a good start for this PR

mds1:
Additional modifiers to exclude foundry addresses (such as Vm and the test contract), as well as an array of user provided addresses, would also be useful

avaxmoon:
Per @mds1 request on #1078, here are some [additional precompiles](https://docs.avax.network/subnets/customize-a-subnet#precompiles) for Avalanche evm-based subnets. A list of such subnets is maintained [here](https://github.com/ava-labs/subnet-evm/tree/master/networks).

tynes:
One thing to note is that Optimism doesn't have extra precompiles, they are technically predeploys so they are evm bytecode placed at specific addresses. I do think that it makes sense to avoid those addresses while fuzzing, it makes sense to include them in `assumeNoPrecompiles`. In the address space for predeploys is 0x4200..00 to 0x4200..0800` in bedrock, the first 2048 addresses in that range

mds1:
[...]

PR #195, opened 2022-10-20
feat: `StdChains` and `assumeNoPrecompiles`

mds1:
Still a WIP, but will close https://github.com/foundry-rs/forge-std/issues/134. Also adds a `StdChains` contract which I think is useful and automatically populates `rpcUrls` based on `foundry.toml` which I think is pretty nifty. 

cc @mattsse to make sure there's no issues with doing these `rpcUrl` overrides in the constructor. Chose to do it there since that way we can hide that from the user, whereas putting it in `setUp` means every now needs to always override `setUp` which would be an annoying breaking


src/StdChains.sol
```diff
@@ -0,0 +1,115 @@
...
+        Chain AvalancheFuji;
+        Chain BnbSmartChain;
+        Chain BnbSmartChainTestnet;
+        Chain GnosisChain;
+    }
+
+    Chains stdChains = Chains({
+        Anvil: Chain("Anvil", 31337, "http://127.0.0.1:8545"),
+        Hardhat: Chain("Hardhat", 31337, "http://127.0.0.1:8545"),
+        Mainnet: Chain("Mainnet", 1, ""),
```
mds1, line 54: Need to populate these with publicly available defaults

src/Components.sol
```diff
@@ -1,13 +1,13 @@
...
+import "src/console.sol";
+import "src/console2.sol";
+import "src/StdAssertions.sol";
+import "src/StdCheats.sol";
+import "src/StdError.sol";
+import "src/StdJson.sol";
+import "src/StdMath.sol";
+import "src/StdStorage.sol";
+import "src/StdUtils.sol";
+import "src/Vm.sol";
```
mattsse, line 13: I'm hesitant to make this change, just because we had some issues with absolute paths in the past and --base-url is not supported by older solc versions

src/StdChains.sol
```diff
@@ -0,0 +1,115 @@
...
+        // The chain name, using underscores as the separator to match `foundry.toml` conventions.
+        string name;
+        // The chain's Chain ID.
+        uint256 chainId;
+        // A default RPC endpoint for this chain.
+        // NOTE: This default RPC URL is included for convenience to facilitate quick tests and
+        // experimentation. Do not use this RPC URL for production test suites, CI, or other heavy
+        // usage as you will be throttled and this is a disservice to others who need this endpoint.
+        string rpcUrl;
+    }
```
mattsse, line 25: this is pretty great, should just work. I don't think this will cause issues. but once this is ready we should add some tests to foundry as well

src/Components.sol
mds1: Was https://github.com/foundry-rs/foundry/issues/3440 ever resolved, and are there other remaining issues related to this?

> and --base-url is not supported by older solc versions

It looks like this was [introduced](https://github.com/ethereum/solidity/blob/develop/Changelog.md#069-2020-06-04.) in 0.6.9. So does that imply paths of this format won't work with projects using solc versions less than that? If so happy to switch back to relative paths for forge-std here (even though these fully qualified paths are definitely superior IMO 😛)

mds1: Updated to use relative paths

src/StdChains.sol
```diff
@@ -0,0 +1,103 @@
...
+        PolygonMumbai: Chain("PolygonMumbai", 80001, "https://rpc-mumbai.matic.today"),
+        Avalanche: Chain("Avalanche", 43114, "https://api.avax.network/ext/bc/C/rpc"),
+        AvalancheFuji: Chain("AvalancheFuji", 43113, "https://api.avax-test.network/ext/bc/C/rpc"),
+        BnbSmartChain: Chain("BnbSmartChain", 56, "https://bsc-dataseed1.binance.org"),
+        BnbSmartChainTestnet: Chain("BnbSmartChainTestnet", 97, "https://data-seed-prebsc-1-s1.binance.org:8545"),
+        GnosisChain: Chain("GnosisChain", 100, "https://rpc.gnosischain.com")
+    });
+
+    // TODO how can we hide the compiler warnings by default? We can't remove the visibility since it's needed for ^0.6.2 compatibility.
+    constructor() internal {
```
ZeroEkkusu, line 68: Create a `private` or `internal` function `_constructor` that returns a placeholder value, and move the logic there. Then, add a helper `private constant` variable and assign the value to it by calling `_constructor`.

```
bool private constant constructorTrigger = _constructor();
```

src/StdUtils.sol
```diff
@@ -2,8 +2,10 @@
 pragma solidity >=0.6.2 <0.9.0;
 
 import "./console2.sol";
+import "./StdChains.sol";
 
-abstract contract StdUtils {
+abstract contract StdUtils is StdChains {
+    Vm private constant vm = Vm(address(uint160(uint256(keccak256("hevm cheat code")))));
```
ZeroEkkusu, line 8: Should those functions be `StdCheats` instead?

src/StdUtils.sol
```diff
@@ -80,4 +82,36 @@ abstract contract StdUtils {
     function addressFromLast20Bytes(bytes32 bytesValue) private pure returns (address) {
         return address(uint160(uint256(bytesValue)));
     }
+
+    function assumeNoPrecompiles(address addr) internal {
```
ZeroEkkusu, line 86:
```suggestion
    function assumeNoPrecompiles(address addr) internal virtual {
```

src/StdUtils.sol
```diff
@@ -80,4 +82,36 @@ abstract contract StdUtils {
...
+        // Assembly required since `block.chainid` was introduced in 0.8.0; // TODO verify version
+        uint256 chainId;
+        assembly {
+            chainId := chainid()
+        }
+        assumeNoPrecompiles(addr, chainId);
+    }
+
+    function assumeNoPrecompiles(address addr, uint256 chainId) internal {
```
ZeroEkkusu, line 95:
```suggestion
    function assumeNoPrecompiles(address addr, uint256 chainId) internal virtual {
```

src/StdChains.sol
```diff
@@ -0,0 +1,103 @@
...
+pragma experimental ABIEncoderV2;
+
+import "./Vm.sol";
+
+/// @dev To add a new chain:
+///   1. Add it into the `Chains` struct, named using CamelCase.
+///   2. Initialize it in the `stdChains` declaration.
+///   3. Add a corresponding `else if` line into the constructor, remembering to check for both
+///      underscore and hyphenated versions of the chain-name when appropriate.
+abstract contract StdChains {
```
ZeroEkkusu, line 12: Should we export `StdChains` in `Components`?

src/StdUtils.sol
mds1: Sorry which functions? `StdChains` just exposes the `stdChains` state var which feels more like a util than a cheat, but I don't feel strongly

mds1: Oh maybe you mean `assumeNoPrecompiles`? Hmm I guess maybe `StdCheats` is the better fit because it uses cheat codes? Maybe a clarification on the distinction between StdUtils and StdCheats would help me choose 😅

src/StdChains.sol
mds1: Maybe? `StdUtils is StdChains`, so it's technically exposed anyway. TBH I'm not totally clear on the use case for `Components.sol`, I think if you could update the README with an explanation of the new architecture that would be really helpful (both for me and future contributors)

src/StdUtils.sol
mds1: Done in 3ad5e8c

src/StdUtils.sol
mds1: Done in 3ad5e8c

src/StdChains.sol
mds1: Ah man this is super clever and I was excited about it. Unfortunately I don't think it'll work though:

```
Compiler run failed
error[8349]: TypeError: Initial value for constant variable has to be compile-time constant.
  --> src/StdChains.sol:18:44:
   |
18 |     uint256 private constant CONSTRUCTOR = _constructor();
````

And we can't use immutables since it's not supported in 0.6.2. @mattsse is there a way forge can hide warnings based on the contract they come from so we can suppress logging of `StdChains` constructor warnings

ZeroEkkusu: @mds1, is it because we marked it `constant`? It doesn’t have to be; just thought it’d look more formal that way.

mds1: Oh right, of course it doesn't need to be constant 🤦

Resolved in 14e1d03

mds1: Huh actually that causes the initialization test to fail as you can see in the CI for that commit, so for some reason this approach actually does not work

src/StdUtils.sol
ZeroEkkusu: What if we made `StdChains` a standalone Component, like all other, and import it in `Test` instead? Seems the use case is somewhat specific - for dealing with chains.

mds1: I'm good with that! That seems better anyway, I forget why I made StdUtils inherit from it at this point anyway, but I don't see a reason why we need to

src/StdChains.sol
ZeroEkkusu: Users should import from `Components.sol`, but aren't required to. Devs must import from `Components.sol`.

If https://github.com/foundry-rs/forge-std/pull/195#discussion_r1008373663 sounds good, then we are going to export it.

mds1: Yep agreed with that linked comment, works for me 👌

src/StdUtils.sol
ZeroEkkusu: It's because `stdChains` is initialized after our `_constructor` is run, so it overrides the setup.

Solution: move `stdChains` initialization into our `_constructor`, before setting the values.

src/StdChains.sol
ZeroEkkusu: *Moved the comment from the wrong thread:*

It's because `stdChains` is initialized after our `_constructor` is run, so it overrides the setup.

Solution: move `stdChains` initialization into our `_constructor`, before setting the values.

ZeroEkkusu (approved):

PR #196, opened 2022-10-25
Add memory-safe notation so that compiling via-ir can optimize effect…

DrakeEvans:
…ively

### DIFF
--- a/src/console2.sol
+++ b/src/console2.sol
@@ -13,6 +13,7 @@ library console2 {
     function _sendLogPayload(bytes memory payload) private view {
         uint256 payloadLength = payload.length;
         address consoleAddress = CONSOLE_ADDRESS;
+        /// @solidity memory-safe-assembly
         assembly {
             let payloadStart := add(payload, 32)
             let r := staticcall(gas(), consoleAddress, payloadStart, payloadLength, 0, 0)


mds1:
Why do we need this? The [docs](https://docs.soliditylang.org/en/v0.8.13/assembly.html#memory-safety) say this comment style is going to be disallowed in a future release, so I'm hesitant to add this since (1) it'll break in the future, and (2) we don't really need to optimize console logs

DrakeEvans:
It is necessary because the IR compiler (via-ir flag) requires assembly to be annotated as memory safe in order to properly optimize code.  See more information here: https://blog.soliditylang.org/2022/03/16/solidity-0.8.13-release-announcement/ relevant information at the heading: "Memory-Safe Assembly / Stack Too Deep"

As you can see, every other assembly block in the code base is annotated correctly except for this one.

If we want to change the annotation style to `assembly ("memory-safe") {...}` I would be happy to do that throughout the code base

mds1:
Yea I'm familiar with what it is, just wasn't sure why we were concerned about optimizing the test contracts. But I didn't realize we were already using this elsewhere, so given that I'm good with adding this and we can revisit once that breaking change is introduced 👌

However, do you mind changing this PR to merge into https://github.com/foundry-rs/forge-std/pull/184 instead (the v0.3 branch)? We're planning to merge that soon so new things are going directly into that branch for an upcoming forge-std release.

(forge-std supports as old as 0.6.2 so we can't use `assembly ("memory-safe")` since that will break compatibility with the older versions)

DrakeEvans:
@mds1 Ive updated the base to be v0.3

mds1 (approved):
Thanks!

PR #197, opened 2022-10-26
test(bound): add even distribution test

ZeroEkkusu:
## Motivation

Check programmatically that wrapped `bound` results are evenly distributed.

## Solution

- Add a test
- Fix solc warning for `_bound` unassigned return

### DIFF
--- a/test/StdUtils.t.sol
+++ b/test/StdUtils.t.sol
@@ -31,6 +31,22 @@ contract StdUtilsTest is Test {
         assertEq(bound(type(uint256).max - 3, 50, 150), 147);
     }
 
+    function testBound_DistributionIsEven(uint256 min, uint256 size) public {
+        size = size % 100 + 1;
+        min = bound(min, UINT256_MAX / 2, UINT256_MAX / 2 + size);
+        uint256 max = min + size - 1;
+        uint256 result;
+
+        for (uint256 i = 1; i <= size * 4; ++i) {
+            // x > max
+            result = bound(max + i, min, max);
+            assertEq(result, min + (i - 1) % size);
+            // x < min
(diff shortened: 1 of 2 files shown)


test/StdUtils.t.sol
```diff
@@ -31,6 +31,22 @@ contract StdUtilsTest is Test {
         assertEq(bound(type(uint256).max - 3, 50, 150), 147);
     }
 
+    function testBound_DistributionIsEven(uint256 min, uint256 size) public {
```
mds1, line 34: Hmm, isn't this just basically reimplementing bound? Perhaps a stronger test is based on your example from #184:

```
in:   0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18 19 20
-------------------------------------------------------------------
out: 15 16 17 18 16 17 18 15 16 17 18 15 16 17 18 15 16 17 18 15 16
```

in other words, test the following behavior (some of these we already have):
- 0 through 3 results in min though min +3
- MAX_UINT through MAX_UINT-3 results in max through max-3
- values of x between min and max (inclusive) are unchanged
- everywhere else, `bound(x, min, max) + 1 == bound(x + 1, min, max)`, OR `bound(x, min, max) - size == bound(x + 1, min, max)` if the output of the LHS is max

ZeroEkkusu: I don’t think it does. It uses a mathematical formula to do the following:

Start from `max`, go number by number and check if it’s `min`, `min + 1`, etc. Do it for a few intervals. Similar for the other side.

The other cases are outside the scope of the test and are already covered, as you said.

mds1: Ok, I'm good with this then, thank you!

PR #198, opened 2022-10-27
Test.sol import statement changed 

GokhanPolat:
Import file path changed to relative path. By the way, only this line was using this type of import statement (with remappings).

If we want to different remappings or even don't want to use remappings, some checker/analyzer tools can be confused that hard to understand of remappings. Nested libraries can be problematic.

My suggestion is using relative path for libraries but I'm open too other suggestions.

### DIFF
--- a/src/Test.sol
+++ b/src/Test.sol
@@ -3,7 +3,7 @@ pragma solidity >=0.6.0 <0.9.0;
 pragma experimental ABIEncoderV2;
 
 import "./Script.sol";
-import "ds-test/test.sol";
+import "../lib/ds-test/src/test.sol";
 
 // Wrappers around Cheatcodes to avoid footguns
 abstract contract Test is DSTest, Script {


mds1:
@mattsse do you have a preference on this? given your comment here: https://github.com/foundry-rs/forge-std/pull/195#discussion_r1001448633 (though I'm not sure if this has ever caused issues, since ds-test is pretty stable and everyone is using the same version anyway)

@GokhanPolat if we do want this change, let's change the target branch to the `v0.3` branch of #184

mattsse:
I'm not aware of any issues re Test import.

there's a possibility that at different ds-test remapping will override this, but as mentioned it's pretty stable.

But the motivation for this PR seems unrelated. 
given that forge-std is bundled with ds-test I think we can safely make this change?

But I guess using a relative import here ensure we always pull in the expected version.

mds1:
> given that forge-std is bundled with ds-test I think we can safely make this change?

Yea, I think this should be safe, and I guess it can't hurt, so may as well do it? Though personally I don't like using relative paths hehe, but since forge-std is included by default, guess it's ok to err on the side of "ensure remappings always get correct ds-test version"

GokhanPolat:
@mds1 

> Though personally I don't like using relative paths

Absolutely agree with you. So how should we proceed?

mds1:
I'm going to close this for now since I don't think it's needed—ds-test has been stable in this repo for quite some time, and we haven't heard reports of any issues resulting from it, let's leave this as-is

But if we ever update the ds-test submodule version, we can make this change at the same time to ensure the new one is used. Thanks @GokhanPolat!

PR #202, opened 2022-11-01
docs: fix ci badge

ZeroEkkusu:
## Motivation

The CI badge broke after merging V1.


mds1:
Looks like CI is failing due to new `forge fmt` changes? There is a new `wrap_comments` flag we may want to set to true here

ZeroEkkusu:
I think all checks will pass after it’s merged merged.

PR #205, opened 2022-11-02
restrict cheats mutability

adhusson:
Many cheats should be view/pure. When you build tooling on top of cheats it's nice to be able to use regular view/pure modifiers.

edit: I haven't restricted the mutability of cheats that change the evm state, the environment, the file system, or the cheat interpreter state (eg `getRecordedLogs` stops recording so it's not `view`).

### DIFF
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -6,8 +6,8 @@ pragma experimental ABIEncoderV2;
 // Cheatcodes are marked as view/pure/none using the following rules:
 // 0. A call's observable behaviour includes its return value, logs, reverts and state writes.
 // 1. If you can influence a later call's observable behaviour, you're neither view nor pure (you are modifying some state be it the EVM, interpreter, filesystem, etc),
-// 2. otherwise if you can be influenced by an earlier call, you're `view`,
-// 3. otherwise you're `pure`.
+// 2. Otherwise if you can be influenced by an earlier call, you're `view`,
+// 3. Otherwise you're `pure`.
 
 interface VmSafe {
     struct Log {
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -227,7 +227,7 @@ interface Vm is VmSafe {
     function selectFork(uint256) external;
     /// Returns the currently active fork
     /// Reverts if no fork is currently active
-    function activeFork() external returns (uint256);
+    function activeFork() external view returns (uint256);
     // Updates the currently active fork to given block number
     // This is similar to `roll` but for the currently active fork
     function rollFork(uint256) external;
@@ -248,7 +248,7 @@ interface Vm is VmSafe {
     function revokePersistent(address) external;
     function revokePersistent(address[] calldata) external;
     // Returns true if the account is marked as persistent
-    function isPersistent(address) external returns (bool);
+    function isPersistent(address) external view returns (bool);
     // In forking mode, explicitly grant the given address cheatcode access
     function allowCheatcodes(address) external;
     // Fetches the given transaction from the active fork and executes it on the current state
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -10,41 +10,41 @@ interface VmSafe {
     }
 
     // Loads a storage slot from an address (who, slot)
-    function load(address, bytes32) external returns (bytes32);
+    function load(address, bytes32) external view returns (bytes32);
     // Signs data, (privateKey, digest) => (v, r, s)
-    function sign(uint256, bytes32) external returns (uint8, bytes32, bytes32);
+    function sign(uint256, bytes32) external pure returns (uint8, bytes32, bytes32);
     // Gets the address for a given private key, (privateKey) => (address)
-    function addr(uint256) external returns (address);
+    function addr(uint256) external pure returns (address);
     // Gets the nonce of an account
-    function getNonce(address) external returns (uint64);
+    function getNonce(address) external view returns (uint64);
     // Performs a foreign function call via the terminal, (stringInputs) => (result)
     function ffi(string[] calldata) external returns (bytes memory);
     // Sets environment variables, (name, value)
     function setEnv(string calldata, string calldata) external;
     // Reads environment variables, (name) => (value)
-    function envBool(string calldata) external returns (bool);
-    function envUint(string calldata) external returns (uint256);
-    function envInt(string calldata) external returns (int256);
-    function envAddress(string calldata) external returns (address);
-    function envBytes32(string calldata) external returns (bytes32);
-    function envString(string calldata) external returns (string memory);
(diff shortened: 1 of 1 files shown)


mds1:
This seems useful, thanks @adhusson. I can see a use case for wanting internal or pure script methods that use a cheat code, and being forced to mark them as mutating due to cheat code use isn't great. 

First question for @gakonst @mattsse, is there any reason why all cheats are mutating? And is there any change this would require upstream?

Second question is what's the concrete distinction we're introducing? I think we're basically taking solidity's rules and generalizing them which seems reasonable:
- `pure` functions don't read state, they a solely a function of the input 
- `view` functions read and return some requested data, whether from the EVM, the filesystem, etc
- everything else mutates (e.g. `setEnv` or `warp`) or cannot have guarantees on what's touched (e.g. `ffi`)

If we agree on those definitions, a few things still aren't clear:
- Where does `expect*` and `record*` cheats fall? By those definitions they're view
- Does this mean `sign` and `addr` should be pure (currently they're view)

mattsse:
> is there any reason why all cheats are mutating

I don't think so.

> I think we're basically taking solidity's rules and generalizing them which seems reasonable:

that sounds reasonable, and supportive of this.

> * Where does expect* and record* cheats fall? By those definitions they're view
> * Does this mean sign and addr should be pure (currently they're view)

I think we can make those changes for both cases.

mds1:
> or the cheat interpreter state (eg `getRecordedLogs` stops recording so it's not view).

@adhusson it looks like you did make `getRecordedLogs` view, curious if that was a mistake or if you changed your mind on this definition?

Either way, I'd say lets:
- Add some comments on our pure/view/mutable definitions at the top of `Vm.sol` for future contributors (once we agree on the definitions)
- Make any required changes based on that definition, such as the changes in my two bullets above. There may be others, so worth doing a double check of all cheats
- Then merge!

gakonst:
Sgtm

adhusson:
> it looks like you did make `getRecordedLogs` view
@mds1 yes, it was a mistake! I've edited.

> Where does expect* and record* cheats fall? By those definitions they're view
They change the interpreter state as they modify the behaviour of later function calls, so I would say they're not view.

> Does this mean sign and addr should be pure (currently they're view)
They are pure!

I also changed `assume` to pure, I think that's the right thing to do formally speaking. Should be good to merge now.

mattsse (approved):
lgtm, 
good to merge @mds1 ?

md1s:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

src/Vm.sol
```diff
@@ -3,48 +3,54 @@ pragma solidity >=0.6.2 <0.9.0;
 
 pragma experimental ABIEncoderV2;
 
+// Cheatcodes are marked as view/pure/none using the following rules:
+// 0. A call's observable behaviour includes its return value, logs, reverts and state writes.
+// 1. If you can influence a later call's observable behaviour, you're neither view nor pure (you are modifying some state be it the EVM, interpreter, filesystem, etc),
+// 2. otherwise if you can be influenced by an earlier call, you're `view`,
+// 3. otherwise you're `pure`.
```
mds1, line 10:
```suggestion
// 2. Otherwise if you can be influenced by an earlier call, you're `view`,
// 3. Otherwise you're `pure`.
```

src/Vm.sol
```diff
@@ -62,13 +68,13 @@ interface VmSafe {
     // Stops collecting onchain transactions
     function stopBroadcast() external;
     // Reads the entire content of file to string, (path) => (data)
-    function readFile(string calldata) external returns (string memory);
-    // Reads the entire content of file as binary. Path is relative to the project root. (path) => (data)
-    function readFileBinary(string calldata) external returns (bytes memory);
+    function readFile(string calldata) external view returns (string memory);
+    // Reads the entire content of file as ggbinary. Path is relative to the project root. (path) => (data)
```
mds1, line 72:
```suggestion
    // Reads the entire content of file as binary. Path is relative to the project root. (path) => (data)
```

src/Vm.sol
```diff
@@ -3,48 +3,54 @@ pragma solidity >=0.6.2 <0.9.0;
 
 pragma experimental ABIEncoderV2;
 
+// Cheatcodes are marked as view/pure/none using the following rules:
+// 0. A call's observable behaviour includes its return value, logs, reverts and state writes.
+// 1. If you can influence a later call's observable behaviour, you're neither view nor pure (you are modifying some state be it the EVM, interpreter, filesystem, etc),
+// 2. otherwise if you can be influenced by an earlier call, you're `view`,
```
mds1, line 9: Would this be an accurate clarification? For example, `vm.projectRoot` reads some data but is not influenced by earlier calls

```suggestion
// 2. otherwise if you can be influenced by an earlier call, or if reading some state, you're `view`,
```

src/Vm.sol
```diff
@@ -134,12 +140,12 @@ interface VmSafe {
...
     // Returns the RPC url for the given alias
-    function rpcUrl(string calldata) external returns (string memory);
+    function rpcUrl(string calldata) external view returns (string memory);
     // Returns all rpc urls and their aliases `[alias, url][]`
-    function rpcUrls() external returns (string[2][] memory);
+    function rpcUrls() external view returns (string[2][] memory);
 
     // If the condition is false, discard this run's fuzz inputs and generate new ones.
-    function assume(bool) external;
+    function assume(bool) pure external;
```
mds1, line 148: _(arbitrary comment location to start a thread)_

A few cheatcodes in `Vm` seem like they can be view too? I think just `isPersistent` and `activeFork`

mds1 (commented):
sorry just a few small nits before merging!

mds1:
@ZeroEkkusu would also love a sanity check from you on this PR if you don't mine

src/Vm.sol
adhusson: Haha I thought about that but can't ffi change that value?

src/Vm.sol
adhusson: Yes, will update!

src/Vm.sol
adhusson: Nevermind, it can't! I'll include your edit.

mds1 (approved):
LGTM, thanks @adhusson! Will merge once @ZeroEkkusu takes a look

ZeroEkkusu (approved):
Sorry for the wait, LGTM if you settled on `record` and a few others being state-changing.

mattsse:
@adhusson needs forge fmt

mds1:
Thanks @adhusson!

PR #206, opened 2022-11-02
Use relative imports in interfaces for VSCode linting compatibility

jameswenzel:
For some strange and frustrating reason, importing from a submodule that uses absolute imports from a common directory name seems to break Solidity linting in VSCode – even though the code compiles and runs fine. I've tried with both Juan Blanco and Hardhat Solidity extensions.

Here's an example code block:

```solidity
pragma solidity ^0.8.13;

import {Test} from "forge-std/Test.sol";
import {IERC721} from "forge-std/interfaces/IERC721.sol";


contract LintingTest is Test {

    function testTransact() public {
        // address payable receiver = payable(0);
        receiver.transfer(1 ether);
    }

}
```

Obviously, this won't compile as-is - but the linter (probably?) doesn't show any errors, either. Uncommenting the `receiver` declaration results in it compiling just fine.

Using relative imports will fix this behavior.

### DIFF
--- a/src/interfaces/IERC1155.sol
+++ b/src/interfaces/IERC1155.sol
@@ -1,6 +1,6 @@
 pragma solidity >=0.6.2;
 
-import "src/interfaces/IERC165.sol";
+import "./IERC165.sol";
 
 /// @title ERC-1155 Multi Token Standard
 /// @dev See https://eips.ethereum.org/EIPS/eip-1155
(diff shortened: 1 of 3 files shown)


mds1 (approved):
Ah, so I think the issue is that extensions have to make assumptions about project structure and where to find things (solidity doesn't yet have a full language server implementation afaik). When you use relative paths it's clear where to find the files, but with absolute paths the extensions don't know the project/framework-specific remappings.

So, thanks for the fix!

@mattsse I am quickly losing my battle to prefer absolute imports haha 😅

Issue #207, opened 2022-11-03
v1.0.0 causes compiling with `via-ir` to take substantially longer


Vectorized:
Recently, I added a `via-ir` testing to Solady for more robust testing, since it is a library.

However, with the latest version of forge-std, the CI is still unable to finish compiling the code after 30 minutes.

See this workflow run: https://github.com/Vectorized/solady/actions/runs/3385556435/jobs/5623841631 (not complete even after 30 minutes)

I left it compiling for more than 3 hours on my machine, to no avail too.

Reverting back `forge-std` to an eariler commit `c19dfd2f2a88a461216b0dd1f4961e1a85dcad46` seems to solve the issue.  

See this workflow run:  https://github.com/Vectorized/solady/actions/runs/3386504801/jobs/5625994122 (completes under 2 minutes)

leonardoalt:
Does the Solidity version also change between the two tests?

mds1:
It looks like they're both 0.8.17

Successful one:
<img width="691" alt="image" src="https://user-images.githubusercontent.com/17163988/199753238-403d719c-065f-44cf-bf69-422fcb2eabcf.png">

Failed one 
<img width="675" alt="image" src="https://user-images.githubusercontent.com/17163988/199753173-429e5446-ee6a-4f29-ba75-9be61c9f28e2.png">

jameswenzel:
I stumbled across this thread from the Seaport release - I don't fully follow the thread, but it seems like `via-ir` slowness isn't an issue with other `solc` compilers, just the one supplied by `svm-rs` – which means at least part of this slowness is probably due an upstream `foundry` or `svm-rs` issue?
https://github.com/ethereum/solidity/issues/13050

Vectorized:
Looks like compiling forge-std alone with `via-ir` takes incredibly long since the v1.0.0 commit.

The previous commit `2a2ce3692b8c1523b29de3ec9d961ee9fbbc43a6`, still compiles reasonably fast.

Vectorized:
Update: looks like commit `72cdd70ae608e32711b4bcf3bbcdafc9eddc3c56` in `v0.3` is the start of the insanely long compile times.

mattsse:
> I don't fully follow the thread, but it seems like via-ir slowness isn't an issue with other solc compilers

Do you mean different solc versions?

what's your OS?

Vectorized:
@mattsse @leonardoalt @mds1 See my previous comment.

[...]

PR #209, opened 2022-11-03
Fix long compile times when using `via-ir`

Vectorized:
For #207.

Tested on Solady. Compiles within 2 minutes instead of 30+ minutes.

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -9,10 +9,7 @@ import "./Vm.sol";
 abstract contract StdCheatsSafe {
     VmSafe private constant vm = VmSafe(address(uint160(uint256(keccak256("hevm cheat code")))));
 
-    /// @dev To hide constructor warnings across solc versions due to different constructor visibility requirements and
-    /// syntaxes, we put the constructor in a private method and assign an unused return value to a variable. This
-    /// forces the method to run during construction, but without declaring an explicit constructor.
-    uint256 private CONSTRUCTOR = _constructor();
+    bool private initialized;
 
     struct Chain {
         // The chain name, using underscores as the separator to match `foundry.toml` conventions.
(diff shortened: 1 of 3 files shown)


mds1:
Thanks! This is a great workaround, minimal UX hit 👌

I just want to see how #207 shakes out before merging this, it's possible there's a solc or svm-rs issue which would mean we can leave things as-is, which would be preferable IMO

marktoda:
We've been suffering wildly long compile times as well, this commit fixes for us too :ok_hand: Seems like maybe a faster way forward than a compiler-level fix? ie might be worth it for dev ex to merge this now and revert if the underlying issue is later resolved in 0.8.18 or something, given forge releases much more often than solc

ZeroEkkusu:
In order to restore `--via-ir` and not hurt the adoption of V1, we can do the following:

1. Remove the `stdChains` variable completely (no half-backed features); keep `assumeNoPrecompiles` (hardcode chain IDs)
3. Find a solution (#207)
4. Put `stdChains` back

Interested if others would find this to be a good solution.

Another idea: We can can change `stdChains` to be a mapping, e.g. `stdChains["mainnet"]`, if that helps with this issue (have not tested). It would also allow users to add custom chains to `stdChains`.

mds1:
Per some discussion in the foundry telegram, let's
- refactor to mapping approach due to added flexibility (also nice that the keys to access chain data matches the expected keys in the config file)
- test to make sure that it doesn't have the slow compile time issue
- if not, merge. if it does, use the struct with the solution in this PR

ZeroEkkusu:
Thanks for reporting the issue, @Vectorized! We are closing your PR, as we’ve implemented a solution with mappings in #217.

I hope you’ll like Forge-Std 1.1 better!

PR #213, opened 2022-11-08
Feat: add serialize (vm.writeJson) to stdjson

odyslam:


ZeroEkkusu:
There has been a massive codebase refactor. Looks like you need to update your Forge-Std first.

odyslam:
I messed up the conflicts. Will address

odyslam:
@ZeroEkkusu ready for review ✨

Next stop is the docs

src/Vm.sol
```diff
@@ -138,12 +138,50 @@ interface VmSafe {
...
+    function serializeString(string calldata, string calldata, string calldata) external returns (string memory);
+    function serializeBytes(string calldata, string calldata, bytes calldata) external returns (string memory);
+
+    function serializeBool(string calldata, string calldata, bool[] calldata) external returns (string memory);
+    function serializeUint(string calldata, string calldata, uint256[] calldata) external returns (string memory);
+    function serializeInt(string calldata, string calldata, int256[] calldata) external returns (string memory);
+    function serializeAddress(string calldata, string calldata, address[] calldata) external returns (string memory);
+    function serializeBytes32(string calldata, string calldata, bytes32[] calldata) external returns (string memory);
+    function serializeString(string calldata, string calldata, string[] calldata) external returns (string memory);
+    function serializeBytes(string calldata, string calldata, bytes[] calldata) external returns (string memory);
```
mds1, line 178: This would require a PR to foundry, but since these `serialize*` methods all have difference signatures should they all just be named `serialize`? I'm ok with either, just noticed in `stdJson` they're all overloads called `serialize` so maybe we should be consistent here.

It looks like most cheatcodes just use overloads if function selectors are sufficient to distinguish them, and types are added to the name if it's return values (not part of the selector) that distinguishes them, so perhaps we should make this change

src/Vm.sol
```diff
@@ -255,4 +293,7 @@ interface Vm is VmSafe {
     function transact(bytes32 txHash) external;
     // Fetches the given transaction from the given fork and executes it on the current state
     function transact(uint256 forkId, bytes32 txHash) external;
+
+
+
 }
```
mds1, line 299: Extra lines introduced, needs `forge fmt`
```suggestion
    function transact(uint256 forkId, bytes32 txHash) external;
}
```

src/Vm.sol
```diff
@@ -138,12 +138,50 @@ interface VmSafe {
     // struct json = { uint256 a; address b; }
     // If we defined a json struct with the opposite order, meaning placing the address b first, it would try to
     // decode the tuple in that order, and thus fail.
```
mds1, line 140: The comments from here and above are for `readJson` right? Let's organize this a bit, since right now the layout is:
- read JSON cheatcodes
- comments on reading JSON
- comments on writing JSON
- write JSON cheatcodes

We should change it to:
- comments on reading JSON
- read JSON cheatcodes
- comments on writing JSON
- write JSON cheatcodes

src/Vm.sol
```diff
@@ -138,12 +138,50 @@ interface VmSafe {
...
+
+    function serializeBool(string calldata, string calldata, bool[] calldata) external returns (string memory);
+    function serializeUint(string calldata, string calldata, uint256[] calldata) external returns (string memory);
+    function serializeInt(string calldata, string calldata, int256[] calldata) external returns (string memory);
+    function serializeAddress(string calldata, string calldata, address[] calldata) external returns (string memory);
+    function serializeBytes32(string calldata, string calldata, bytes32[] calldata) external returns (string memory);
+    function serializeString(string calldata, string calldata, string[] calldata) external returns (string memory);
+    function serializeBytes(string calldata, string calldata, bytes[] calldata) external returns (string memory);
+    function writeJson(string calldata, string calldata) external;
+    function writeJson(string calldata, string calldata, string calldata) external;
```
mds1, line 180: What does `vm.writeJson` do different than `vm.writeFile`, or are they identical? Also can we add some comments explaining the input args here, similar to how we have:

```solidity
// Writes data to file, creating a file if it does not exist, and entirely replacing its contents if it does.
// (path, data) => ()
function writeFile(string calldata, string calldata) external;
```

In particular I'm not clear on what the third `string` argument is

src/StdJson.sol
```diff
@@ -68,4 +68,80 @@ library stdJson {
     function readBytesArray(string memory json, string memory key) internal pure returns (bytes[] memory) {
         return abi.decode(vm.parseJson(json, key), (bytes[]));
     }
+    function serialize(string memory jsonKey, string memory key, bool value) internal returns (string memory) {
```
mds1, line 71: Nit on adding a blank line
```suggestion

    function serialize(string memory jsonKey, string memory key, bool value) internal returns (string memory) {
```

So the TLDR usage of this library, with `using stdJson for string;` is:
- To read JSON: Use `parseRaw` and the `read*` methods
- To write JSON: Use `write` and the `serialize` methods

Is that correct? If so let's add some comments at the top of the file with that info so its easier for devs to parse/skim this file

mds1 (commented):
Thanks for the work on `writeJson`, super excited about this!

src/Vm.sol
odyslam: @mds1 I had some weird conflicts with how macro was generating the bindings enum, so this was more explicit and solved all issues. Would prefer to keep them like this.

odyslam:
@mds1 resolved all your comments and opened a PR with a fix to foundry

odyslam:
@mds1 I implemented a change to how writeJson works (https://github.com/foundry-rs/foundry/pull/3647) to support the use-case of updating a single value with another value (instead of a value with a whole json). In that change, I also changed how writeJson works, so now you pass the stringified JSON (instead of the key for the in-memory object).

Will update the comments accordingly

mds1 (approved):
LGTM, thanks! Test failure unrelated and should be fixed in forge with next nightly

PR #216, opened 2022-11-09
chore: remove `\`, update `ds-test`

ZeroEkkusu:
## Motivation

`ds-test` was rolled back in #213. There's `\` file in the root.


mds1:
Oops I fixed locally before merging that branch and forgot to push, thanks for catching!

PR #217, opened 2022-11-10
Fix slow compilation with `via-ir`

### LINKED ISSUE #207: v1.0.0 causes compiling with `via-ir` to take substantially longer

### LINKED ISSUE #215: `Nested arrays not yet implemented.` error crashing compilation against solc 0.6.x
Changes introduced in PR #184 - commit https://github.com/foundry-rs/forge-std/commit/17656a2fa5453f495d8c1302a0cedded912457eb , seems to be causing issues with compilation using Solc 0.6x (tested with 0.6.2, 0.6.7, 0.6.9, 0.6.10).

Each one of them fails to compile, either by `UnimplementedFeatureError: Nested arrays not yet implemented.` error. Including `pragma experimental ABIEncoderV2;` in compiled file results instead in `YulException: Stack too deep when compiling inline assembly: Variable dataEnd is 1 slot(s) too deep inside the stack.`

Preceding commit works just fine (https://github.com/foundry-rs/forge-std/commit/2a2ce3692b8c1523b29de3ec9d961ee9fbbc43a6)

mds1:
Closes #207, closes #215, replaces #209

Changes:
- `stdChains` is now a mapping. The keys match what would be expected in `foundry.toml`, e.g. `stdChains["mainnet"]` or `stdChains["optimism_goerli"]`. This is a breaking change.
- I removed support for the `optimism-goerli` syntax to keep things simpler, as most config files I've seen use underscores instead of hyphens anyway. If this is requested, it can be easily added in a future PR
- Confirmed that building with via-IR through foundry takes ~100 seconds on my machine, which is what I'd expect
- Added a `src/test/Compilation.sol` file that inherits `Test`, used to benchmark via-ir compilation times in CI

cc'ing people who may be interested: @Vectorized @marktoda @mattsse @leonardoalt @ZeroEkkusu


mds1:
Seems the addition of `CompilationTest` now triggers the same issue reported in #215, so we should figure out the source of this stack too deep error first. Though, I don't get stack too deep using these new forge-std versions in the repos this was tested with, so it seems strange we do get it with an empty `contract CompilationTest is Test {}` contract.

mattsse (commented):
can't really comment on DX since I haven't used this, but perhaps we can improve this by adding constants for the chain names? otherwise this lgtm.


Interestingly, I just tried moving Chain to a separate library:

https://gist.github.com/mattsse/3b0429f863202f3779521672f9e5b4a6

this reduces compile time from 180s to 22s, for the `solc` example mentioned in the issue.

mds1:
Ok once https://github.com/foundry-rs/foundry/pull/3659 is merged we should be all set:
- via-ir compilation time down to a normal amount from the mapping switch
- "Stack too deep" resolved by the new cheatcode introduced in #3659 that uses structs instead of array of strings, closes https://github.com/foundry-rs/forge-std/issues/215
- "Nested arrays not yet implemented" resolved by using `pragma experimental ABIEncoderV2;`, also https://github.com/foundry-rs/forge-std/issues/215

.github/workflows/ci.yml
```diff
@@ -21,12 +21,25 @@ jobs:
       - name: Install dependencies
         run: forge install
 
-      - name: Check backward compatibility
-        run: |
-          forge build --skip test --use solc:0.8.0
-          forge build --skip test --use solc:0.7.6
-          forge build --skip test --use solc:0.7.0
-          forge build --skip test --use solc:0.6.2
+      - name: Check compatibility with 0.8.17 --via-ir
```
ZeroEkkusu, line 24:
```suggestion
      - name: Measure compilation time with 0.8.17 --via-ir
```

src/StdCheats.sol
```diff
@@ -227,49 +208,47 @@ abstract contract StdCheatsSafe {
...
+            else if (isEqual(name, "arbitrum_one")) stdChains["arbitrum_one"].rpcUrl = rpcUrl;
+            else if (isEqual(name, "arbitrum_one_goerli")) stdChains["arbitrum_one_goerli"].rpcUrl = rpcUrl;
+            else if (isEqual(name, "arbitrum_nova")) stdChains["arbitrum_nova"].rpcUrl = rpcUrl;
+            else if (isEqual(name, "polygon")) stdChains["polygon"].rpcUrl = rpcUrl;
+            else if (isEqual(name, "polygon_mumbai")) stdChains["polygon_mumbai"].rpcUrl = rpcUrl;
+            else if (isEqual(name, "avalanche")) stdChains["avalanche"].rpcUrl = rpcUrl;
+            else if (isEqual(name, "avalanche_fuji")) stdChains["avalanche_fuji"].rpcUrl = rpcUrl;
+            else if (isEqual(name, "bnb_smart_chain")) stdChains["bnb_smart_chain"].rpcUrl = rpcUrl;
+            else if (isEqual(name, "bnb_smart_chain_testnet")) stdChains["bnb_smart_chain_testnet"].rpcUrl = rpcUrl;
+            else if (isEqual(name, "gnosis_chain")) stdChains["gnosis_chain"].rpcUrl = rpcUrl;
```
ZeroEkkusu, line 251:
```suggestion
            stdChains[name].rpcUrl = rpcUrl;
```

src/StdCheats.sol
```diff
@@ -278,11 +257,6 @@ abstract contract StdCheatsSafe {
         return keccak256(abi.encode(a)) == keccak256(abi.encode(b));
```
ZeroEkkusu, line 257: Delete `isEqual` function.

src/StdCheats.sol
```diff
@@ -300,13 +274,13 @@ abstract contract StdCheatsSafe {
         vm.assume(addr < address(0x1) || addr > address(0x9));
 
         // forgefmt: disable-start
-        if (chainId == stdChains.Optimism.chainId || chainId == stdChains.OptimismGoerli.chainId) {
+        if (chainId == stdChains["optimism"].chainId || chainId == stdChains["optimism-goerli"].chainId) {
```
ZeroEkkusu, line 277:
```suggestion
        if (chainId == stdChains["optimism"].chainId || chainId == stdChains["optimism_goerli"].chainId) {
```

src/StdCheats.sol
```diff
@@ -300,13 +274,13 @@ abstract contract StdCheatsSafe {
...
             // https://github.com/ethereum-optimism/optimism/blob/eaa371a0184b56b7ca6d9eb9cb0a2b78b2ccd864/op-bindings/predeploys/addresses.go#L6-L21
             vm.assume(addr < address(0x4200000000000000000000000000000000000000) || addr > address(0x4200000000000000000000000000000000000800));
-        } else if (chainId == stdChains.ArbitrumOne.chainId || chainId == stdChains.ArbitrumOneGoerli.chainId) {
+        } else if (chainId == stdChains["arbitrum-one"].chainId || chainId == stdChains["arbitrum-one-goerli"].chainId) {
```
ZeroEkkusu, line 280:
```suggestion
        } else if (chainId == stdChains["arbitrum-one"].chainId || chainId == stdChains["arbitrum_one_goerli"].chainId) {
```

src/StdCheats.sol
```diff
@@ -300,13 +274,13 @@ abstract contract StdCheatsSafe {
...
             // https://developer.arbitrum.io/useful-addresses#arbitrum-precompiles-l2-same-on-all-arb-chains
             vm.assume(addr < address(0x0000000000000000000000000000000000000064) || addr > address(0x0000000000000000000000000000000000000068));
-        } else if (chainId == stdChains.Avalanche.chainId || chainId == stdChains.AvalancheFuji.chainId) {
+        } else if (chainId == stdChains["avalanche"].chainId || chainId == stdChains["avalanche-fuji"].chainId) {
```
ZeroEkkusu, line 283:
```suggestion
        } else if (chainId == stdChains["avalanche"].chainId || chainId == stdChains["avalanche_fuji"].chainId) {
```

test/StdUtils.t.sol
```diff
@@ -90,7 +90,7 @@ contract StdUtilsTest is Test {
     }
 
     function testAssumeNoPrecompilesL1(address addr) external {
-        assumeNoPrecompiles(addr, stdChains.Mainnet.chainId);
+        assumeNoPrecompiles(addr, stdChains["mainnet"].chainId);
         assertTrue(addr < address(1) || addr > address(9));
     }
```
ZeroEkkusu, line 95: This should be in `StdCheats.t.sol` and it should test Optimism, for example, because it didn't catch the `-` vs `_` change above.

src/test/Compilation.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.6.2 <0.9.0;
+
+pragma experimental ABIEncoderV2;
+
+import "../Test.sol";
+
+// Even though this contract is in the `src/` directory, it is a test contract. It simply inherits
```
ZeroEkkusu, line 8: Move this file to `test/`. See the CI suggestion below.

.github/workflows/ci.yml
```diff
@@ -21,12 +21,25 @@ jobs:
...
+        if: always()
+        run: forge build --skip test --use solc:0.8.17 --via-ir
```
ZeroEkkusu, line 26: This is for `Compilation.sol`, right? Do this instead:

```suggestion
        run: forge build --skip test --contracts test/Compilation.sol --use solc:0.8.17 --via-ir
```

ZeroEkkusu (changes requested):
Good job! Does @mattsse's suggestion compile even faster? See suggestions.

.github/workflows/ci.yml
mds1: Done in 3212794

src/StdCheats.sol
mds1: Ah great call, done in 3212794

src/StdCheats.sol
mds1: Done in 3212794

src/StdCheats.sol
mds1: Fixed in 3212794

src/StdCheats.sol
mds1: Fixed in 3212794

src/StdCheats.sol
mds1: Fixed in 3212794

test/StdUtils.t.sol
mds1: Done in 3212794

src/test/Compilation.sol
mds1: I added a few more contracts to this file to test compilation of each abstract contract in forge-std. Is it possible to compile all 4 with a single forge invocation if they live in the test dir? I'm not too familiar with the `--contracts` / other flags for this

.github/workflows/ci.yml
mds1: See above comment

mds1:
@ZeroEkkusu Addressed most comments in the last commit, one remaining question for you around the compilation test contracts.

Regarding compilation times:
- @mattsse's version was struct based (so users can't add their chains) and IIRC that took ~25 seconds with via-ir on my machine
- The original mapping version of this PR (0527dca, the commit right before 3212794) took ~105 seconds with via-ir on my machine
- The most recent commit (3212794), took 63 seconds with via-ir on my machine (thanks to your simplification in the for loop)

IMO sticking with the current UX/code is a good tradeoff between compilation time and flexibility (so users can add chains, change the aliases, etc), especially given that via-ir performance should improve over time.

mattsse (approved):
lgtm,

good to merge?

src/test/Compilation.sol
ZeroEkkusu: If you want to measure compilation time for all four, use the suggestion below. To get individual results, create `test/compilation/` and separate the contracts into files. Then have dedicated steps for each one, e.g. `forge build --skip test --contracts test/compilation/CompilationScrpit.sol --use solc:0.8.17 --via-ir`.

ZeroEkkusu:
Once the discussions are resolved, @mattsse.

Also, @mds1, should chains names be, e.g. `Arbitrum One` instead of `ArbitrumOne`?

src/test/Compilation.sol
mds1: Nice, thanks. Four separate benchmarks seems useful, just updated that in b024cee

mds1:
Agreed on name changes, all changes are now implemented. @ZeroEkkusu could you take one more quick look? Feel free to merge after if everything looks good!

test/StdCheats.t.sol
```diff
@@ -247,6 +247,14 @@ contract StdCheatsTest is Test {
...
         }
     }
+
+    function testAssumeNoPrecompilesL1(address addr) external {
+        assumeNoPrecompiles(addr, stdChains["optimism_goerli"].chainId);
+        assertTrue(
+            addr < address(1) || (addr > address(9) && addr < address(0x4200000000000000000000000000000000000000))
+                || addr > address(0x4200000000000000000000000000000000000800)
+        );
+    }
```
ZeroEkkusu, line 257: Why are `address(1)`-`address(9)` here?

```suggestion
    function testAssumeNoPrecompiles(address addr) external {
        assumeNoPrecompiles(addr, stdChains["optimism_goerli"].chainId);
        assertTrue(
           addr < address(0x4200000000000000000000000000000000000000))
                || addr > address(0x4200000000000000000000000000000000000800)
        );
    }
```

.github/workflows/ci.yml
```diff
@@ -21,12 +21,37 @@ jobs:
       - name: Install dependencies
         run: forge install
 
-      - name: Check backward compatibility
-        run: |
-          forge build --skip test --use solc:0.8.0
-          forge build --skip test --use solc:0.7.6
-          forge build --skip test --use solc:0.7.0
-          forge build --skip test --use solc:0.6.2
+      - name: Measure compilation time of Test with 0.8.17 --via-ir
```
ZeroEkkusu, line 24: Each of these takes about 40 seconds. Let's create a separate job after `fmt` instead?

ZeroEkkusu (approved):
@mds1, #218 + release after this.

test/StdCheats.t.sol
mds1: AFAIK all EVM compatible chains will need to have the same precompiles at those addresses, otherwise they wouldn't be EVM compatible. Test name fixed in f42f04d

.github/workflows/ci.yml
mds1: Not sure I understand the suggestion here, each job (build, test, fmt) runs in parallel, so this seems ok

mds1: For example, see how test/fmt finished but build is still going for the last commit 
image

test/StdCheats.t.sol
ZeroEkkusu: Makes sense! Thanks for the info.

.github/workflows/ci.yml
ZeroEkkusu: I forgot. Sorry for nit picking- can we move them after the backward compatibility check then? I’d like the workflow to fail fast, instead of waiting 3 minutes for those to finish, to then check backward compatibility, which is more likely to catch something.

mds1: Makes sense. Done in fab939e, also added an 0.6.12 check for good measure

PR #220, opened 2022-11-15
Fix unexpected condition in `bound`

Rubilmax:
This is not causing a bug because it's expected that if `x < max` then `x < min` thanks to the condition met earlier in the function's execution: `x < min || x > max` (otherwise it would have returned `x`)

But this PR is for code clarity


ZeroEkkusu (approved):

mds1:
LGTM, CI failure unrelated

PR #222, opened 2022-11-16
refactor: move bytesToUint to StdUtils for shared use

NoahMarconi:


ZeroEkkusu (changes requested):
Feel free to just copy it over to `StdUtils` as an `internal` function.

Some observations:

- The function reverts with `Arithmetic over/underflow` if the input is longer than 32 bytes. That's bad DX.

    Add a manual check that reverts with  `"StdUtils bytesToUint(bytes): Bytes length exceeds 32."`.

- We can make the function more efficient:
   ```solidity
   abi.decode(abi.encodePacked(new bytes(32 - b.length), b), (uint256));
   ```

NoahMarconi:
`Feel free to just copy it over to StdUtils as an internal function.` by this do you mean to include the function in both contracts?

Without putting it into a `library` and accessing from the library in both `StdUtils` and `StdCheats` we run into a compilation problem as both define the same function.

I pushed an alternative to using a `library` and made prefixed the version in `StdCheats` with an `_` and left it private.

src/StdUtils.sol
```diff
@@ -74,6 +74,11 @@ abstract contract StdUtils {
         return addressFromLast20Bytes(keccak256(abi.encodePacked(bytes1(0xff), deployer, salt, initcodeHash)));
     }
 
+    function bytesToUint(bytes memory b) internal pure returns (uint256) {
+        require(b.length <= 32, "StdUtils bytesToUint(bytes): Bytes length exceeds 32.");
+        return abi.decode(abi.encodePacked(new bytes(32 - b.length), b), (uint256));
+    }
```
mds1, line 80: Is there a way we can avoid duplicating the implementations, and have a single `bytesToUint` or have this one just call `_bytesToUint`? @ZeroEkkusu from what I gather it seems this was originally a library that was deduped but you prefer this version?

NoahMarconi: If there's agreement,  I can put it back into a library and have both use the single implementation from there.

ZeroEkkusu: `bytesToUint` in `StdUtils` is *the* implementation. The one in `StdCheats`, which has been renamed to `_bytesToUint`, is just a `private` helper function for some std-cheats. We do not care about that one, so to speak, because it's not an actual feature - it was used only during the development and we do not expect to ever revisit it. Does that make sense?

mds1: Right, but they serve the same purpose so seems logical to dedupe. But tbh I don't feel strongly and I don't think deduping here is a big deal so happy to concede and not block on this

mds1 (approved):

ZeroEkkusu:
Okay, we’re merging this. Thanks, @NoahMarconi.

ZeroEkkusu:
Thanks.

PR #223, opened 2022-11-16
Fix IERC1155 Interface name

jameswenzel:
The interface in `IERC1155.sol` is named `ERC1155`, rather than `IERC1155` 

Unfortunately this a breaking change; we could also include `interface ERC1155 is IERC1155 {}` for backwards compatibility?


mds1:
Ah nice catch. @ZeroEkkusu we should probably merge this then tag as v1.2.0 since it's breaking, wdyt? I don't think we should keep `ERC1155` for backwards compatibility, since this is a pretty new feature we should just fix it

ZeroEkkusu:
Yeah; let's merge the other PRs, and release `1.2.0`.

PR #226, opened 2022-11-18
feat: add keyExists helper into StdJson

vicnaum:
Proposing `keyExists(...)` helper that will explicitly determine if a given key exists in a JSON.

**Rationale:** For an average user, who is using the StdJson decode helpers, it is not trivial to find a way around sudden EVM reverts if requested JSON key doesn't exist.
The documentation doesn't mention what to do in such a case, and try/catch statements do not work either.
This helper can help to avoid reverts when parsing JSON.

**Implementation:** Although verifying the key exists is just checking the return length of raw `parseJson` bytes, for an average user this method is not trivial nor well documented (and also looks ugly in code).

**Testing:** Although lack of formal tests, the implementation was manually verified against an empty string (encodes as 64 bytes: 0x20 and 0x00), `null` (encoded as 32 zero bytes). Not existing key returns `0x` empty bytes array with zero length. More edge cases might need to be verified, depending on how rust JSON parser works.

**Example use case:**
Assume we use `addresses.json` containing deployment addresses, and we need to know if a contract was already deployed to a given network or we need to deploy it. `keyExists('network.ContractName')` will give us the answer.

### DIFF
--- a/src/StdJson.sol
+++ b/src/StdJson.sol
@@ -33,6 +33,10 @@ library stdJson {
         return vm.parseJson(json, key);
     }
 
+    function keyExists(string memory json, string memory key) internal pure returns (bool) {
+        return vm.parseJson(json, key).length > 0;
+    }
+
     function readUint(string memory json, string memory key) internal pure returns (uint256) {
         return abi.decode(vm.parseJson(json, key), (uint256));
     }


mds1:
This seems reasonable to me, I like it. Would be great to add a simple test for each case, there's a `broadcast.log.json` already in the repo we can use

@odyslam wdyt of this?

odyslam:
Yeah, please add more tests and a mention to the docs.

TBH, it's not super intuitive that it returns an empty string (vs reverting). At nomad we sort of do the same to see if a key exists, so it's it's nice to standardise it.

mds1:
@odyslam Do you think we should modify `vm.parseJson` to revert if the key isn't found, then add a `vm.tryParseJson` that returns an empty string if not found? That way the default/current usage becomes a little stricter and safer, and then this forge-std method wold operate on `tryParseJson`, since older solidity versions don't have try/catch?

odyslam:
We discussed this internally and aligned that it's better to return an empty string, as it adds flexibility. So, after using this for prod, we aligned that we don't think a revert behavior had any value for us. 

Pinging @anna-carroll and  @prestwich in case they want to chime in about the UX. We have used JSON input/output **extensively**  over the last month.

mds1:
Hmm, but doesn't the above split into two methods enable the same flexibility but add add safety when that's not needed? 

Personally I'd prefer to revert in `parseJson` if the key isn't found (and tbh I assumed it did), that way it's easy to catch if my JSON input file has a typo in the key name or is missing a key, and my script doesn't succeed anyway with empty/wrong input values. This feels like an easy footgun when using JSON inputs in scripts. How are you currently mitigating this risk?

Another approach is modifying the `read*` stdJson methods to revert if parseJson returns an empty string, and adding `tryRead*` alternatives for when you need the flexibility. But this forces users to use the stdJson lib to get that safety, so this seems less ideal

Now I'm wondering how many people have accidentally ran scripts/deployed contracts with the wrong values due to a misconfigured JSON input 😅

prestwich:
> Now I'm wondering how many people have accidentally ran scripts/deployed contracts with the wrong values due to a misconfigured JSON input 

probably very few, because you generally have to abi decode the output of `readJson`

> Personally I'd prefer to revert in parseJson if the key isn't found (and tbh I assumed it did), that way it's easy to catch if my JSON input file has a typo in the key name or is missing a key

optional keys are standard practice in most real-world json (see e.g. the `x-nullable` in openapi ,or `"type: ["string", "null"]` in json schema). A config may not need a key under certain conditions, and should have a way to test if that is the case without reverting the tx. Your desire to revert is served by the failing ABI decoding after the parseJson returns an empty string.

basically, reverting would be bad behavior, considering that optional keys are standard in json, and not indicative of malformatted input

mds1:
> probably very few, because you generally have to abi decode the output of `parseJson`...Your desire to revert is served by the failing ABI decoding after the parseJson returns an empty string.

Yea, my thinking was this would mainly be an issue for string/bytes types which admittedly are less common input types. Or would ABI decoding revert for those types also? I just assumed it'd return the same empty value.

> optional keys are standard practice in most real-world json (see e.g. the `x-nullable` in openapi ,or `"type: ["string", "null"]` in json schema). A config may not need a key under certain conditions, and should have a way to test if that is the case without reverting the tx. 
> 
> basically, reverting would be bad behavior, considering that optional keys are standard in json, and not indicative of malformatted input

This is a fair point, though I think there's an argument to be made that things are a bit different with e.g. contract deploys or one off scripts, where you might never have optional keys and need to get it right the first execution. Unless ABI decoding into all types fails when an empty value is given, in which case I'm happy to concede 🙂 (will have to test this later, or perhaps you know offhand)

prestwich:
empty string is not a valid abi encoding of empty string. strings are
always 96 bytes or linger when encoded. in fact, it's not a valid abi
encoding of any type. the minimum length of an abi encoded blob is 32. so
decoding it will always revert.

for a real world example, you might have configs specialized by network, so
you check `config.NETWORK_NAME.OPTIONAL_KEY` to see if the network needs a
particular setting

mds1:
> the minimum length of an abi encoded blob is 32. so
decoding it will always revert.

Ahh right of course, ok awesome. Thanks

So current approach + PR all LGTM

mds1 (approved):

mds1:
@vicnaum do you mind adding a few tests before we merge? you can use the existing JSON file in the repo

odyslam:
I would also urge @vicnaum to add a small mention to the docs :)

vicnaum:
> @vicnaum do you mind adding a few tests before we merge? you can use the existing JSON file in the repo

Sure! Will add the tests. Should I test only this feature? Cause I don't seem to find any StdJson tests...

vicnaum:
@mds1  @odyslam @prestwich I've committed some tests for keyExists and... houston, we have a problem:
- Empty object (as in `"returns": {}`) is parsed as "0x"
- "0x0" (as in `"value": "0x0"`) is also parsed as "0x"

So both of these result in the same empty bytes array as if key didn't exist. How should we fix this?

Also, `null` (as in `"to": null`) is parsed as an empty string "", which doesn't sound right for me, but I'm not sure what it should parse to - there's no `null` in Solidity.

prestwich:
Valid solidity types should output valid encodings, while invalid types should output the empty encoding

> "0x0" (as in "value": "0x0") is also parsed as "0x"

this seems like a bug? 

> Empty object (as in "returns": {}) is parsed as "0x"

this seems correct, there is no abi encoding of an empty struct. Empty structs are not valid solidity either.
 
Screenshot 2022-11-20 at 10 09 34 AM

Empty arrays are similarly illegal.

Screenshot 2022-11-20 at 10 11 54 AM

> Also, null (as in "to": null) is parsed as an empty string "", which doesn't sound right to me, but I'm not sure what it should parse to - there's no null in Solidity.

This seems correct, as null is not a legal type in solidity.

prestwich:
the documentation of `parseJson` should probably explicitly state that invalid solidity types will be parsed into invalid abi encodings 🤔

vicnaum:
> Valid solidity types should output valid encodings, while invalid types should output the empty encoding
> 
> > "0x0" (as in "value": "0x0") is also parsed as "0x"
> 
> this seems like a bug in the parser? I would expect `0x0` to be treated as either

I'm super sorry! That was a bug on my tests side - the json key path was wrong and pointed to non-existent object :)

I've fixed the tests and yes: the `value: "0x0"` is parsed as bytes `0x00` of length 1, which is correct imho.

For the rest of the cases - I've left TODO's in tests with these comments - in case if somebody would want to change the parser to revert on `null` and `emptyArray`. Right now the parser behavior in these special cases it's a little bit inconsistent (it still reverts on `emptyObject`), but I reckon this is a separate issue.

vicnaum:
> > emptyArray
> 
> Oh one edge-case, empty dynamic arrays are legal. so an empty json array should probably be parsed as the 32 byte `\x00` null string

Let's move the parser discussion to a `foundry` repo, I've created an issue there:
https://github.com/foundry-rs/foundry/issues/3731

After there is a decision on parser behavior - I can update the tests accordingly.

mds1:
Closing since this will be implemented natively in https://github.com/foundry-rs/foundry/pull/5431

PR #229, opened 2022-11-20
feat: add common addresses

### LINKED ISSUE #211: feat: add special forge addresses as constants for easier access by users
Let's create constants such as:

```solidity
address internal constant CONSOLE_ADDRESS = 0x000000000000000000636F6e736F6c652e6c6f67
```

This (1) makes it easier for users to find/know these special addresses, and (2) lets them more easily exclude them from fuzzer inputs. Addresses to include:
- the console.log address
- default test contract address
- default msg.sender / tx.origin / caller addresses

mds1:
Closes #211

### DIFF
--- a/src/Common.sol
+++ b/src/Common.sol
@@ -4,10 +4,25 @@ pragma solidity >=0.6.2 <0.9.0;
 import {StdStorage, Vm} from "./Components.sol";
 
 abstract contract CommonBase {
+    // Cheat code address, 0x7109709ECfa91a80626fF3989D68f67F5b1DD12D.
     address internal constant VM_ADDRESS = address(uint160(uint256(keccak256("hevm cheat code"))));
+
+    // console.sol and console2.sol work by executing a staticcall to this address.
+    address internal constant CONSOLE = 0x000000000000000000636F6e736F6c652e6c6f67;
+
+    // Default address for tx.origin and msg.sender, 0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38.
+    address internal constant DEFAULT_SENDER = address(uint160(uint256(keccak256("foundry default caller"))));
+
+    // Address of the test contract, deployed by the DEFAULT_SENDER.
+    address internal constant TEST_CONTRACT = 0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f;
+
+    // Create2 factory used by scripts when deploying with create2, https://github.com/Arachnid/deterministic-deployment-proxy.
+    address internal constant CREATE2_FACTORY = 0x4e59b44847b379578588920cA78FbF26c0B4956C;
+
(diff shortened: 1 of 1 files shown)


mds1:
@mattsse any idea about this forge panic in CI? https://github.com/foundry-rs/forge-std/actions/runs/3506916201/jobs/5874265444

ZeroEkkusu (commented):
We already have `VM_ADDRESS`, and `CONSOLE_ADDRESS` in `console`.

What do you think about naming the constants `CONSOLE_ADDRESS`, `SENDER_ADDRESS`, `TEST_ADDRESS`, `CREATE2_ADDRESS`, for consistency?

mds1:
> We already have `VM_ADDRESS`, and `CONSOLE_ADDRESS` in `console`.

Yea, I still think they should be redeclared here. If I `import {Test} from foundry/forge-std` I won't have the console address handy, but the fuzzer can still pass it as an input and therefore I need to now import console to add `vm.assume(addr != CONSOLE)`. That's why I think having these in the common file makes sense

> What do you think about naming the constants `CONSOLE_ADDRESS`, `SENDER_ADDRESS`, `TEST_ADDRESS`, `CREATE2_ADDRESS`, for consistency?

Personally not a fan for two reasons:
- Adding `_ADDRESS` wastes characters, since we already know it's an address due to the variable's type
- The truncated versions here lose important details IMO, i.e. "default sender" is more explicit than sender and "test contract" more explicit than test

If others disagree though, happy to go with the renames

ZeroEkkusu (approved):

src/Common.sol
```diff
@@ -4,10 +4,25 @@ pragma solidity >=0.6.2 <0.9.0;
...
     address internal constant VM_ADDRESS = address(uint160(uint256(keccak256("hevm cheat code"))));
+
+    // console.sol and console2.sol work by executing a staticcall to this address.
+    address internal constant CONSOLE = 0x000000000000000000636F6e736F6c652e6c6f67;
+
+    // Default address for tx.origin and msg.sender, 0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38.
+    address internal constant DEFAULT_SENDER = address(uint160(uint256(keccak256("foundry default caller"))));
+
+    // Address of the test contract, deployed by the DEFAULT_SENDER.
+    address internal constant TEST_CONTRACT = 0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f;
```
mattsse, line 17: should we also prefix this with `DEFAULT_`?

mattsse (commented):
panic unrelated.

src/Common.sol
mds1: Ah yea we should, updated in 69aae3d

mattsse (approved):
lgtm

Issue #233, opened 2022-11-22
new patch release v1.1.1


mattsse:
@mds1 can we publish a new patch release, the mycrypto API error is persistent in foundry CI

mds1:
On it now

mds1:
Published: https://github.com/foundry-rs/forge-std/releases/tag/v1.1.1

PR #234, opened 2022-11-22
chore: bump version to v1.1.1

mds1:
For https://github.com/foundry-rs/forge-std/issues/233

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "forge-std",
-  "version": "1.0.0",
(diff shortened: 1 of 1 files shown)


mds1:
@mattsse I know you mentioned in https://github.com/foundry-rs/forge-std/pull/229 the panic in CI is unrelated, curious if you know the cause fix? I don't see it when I run the tests locally

mattsse:
this looks like a new rate limit variant, adding support now

Issue #232, opened 2022-11-21
`[rpc_endpoints]` vars should be lazily loaded


adhusson:
By default in forge, `[rpc_endpoints]` entries are loaded lazily. So if it contains e.g. `goerli = ${GOERLI_URL}` but the `GOERLI_URL` is not defined, that's fine as long as goerli isn't used by a test/script.

Whatever the wisdom of doing that, forge-std breaks this since #195. Maybe moving `stdChains` from a mapping to a function that does the custom-or-default logic on the fly would be as good?

Some context to the use case: a foundry repo used a dependency by a non-foundry repo. Some scripts of the foundry repo are used, but the networks never are.

PR #235, opened 2022-11-22
feat: reimplement `StdChains`

### LINKED ISSUE #225: Compilation still slow with via-ir
Was chatting with Georgios about this and ran some benchmarks / tests on https://github.com/uniswap/universal-router. Posting here to hopefully help debug this

Repo:
- solc 0.8.17
- via-ir = true
- optimizer runs = 1,000,000
- forge-std version: 1.1.0 33d4895fb629aabf12e0f60740367bd03e702bd1

Cold build with hardhat: 16s

Cold build with forge: 45s
Removing only test/: 23s
Removing only script/: 30s
Removing test/ and script/: 7s
Removing forge-std and test/ and script: 6.5s

With the hunch that console.log or console2.log was causing this, i removed those from forge-std and all callsites: 40s

So, something in forge-std is causing the slow compilation and it's not console. I can maybe go one-by-one and test further later but wanted to post here in case other folks have time to TAL. 45s obviously not terrible but 3x worse than hardhat and bit of a drag on development

ZeroEkkusu:
## Motivation

Resolve #225, #232

## Solution

Lazy-initialize chains.

Tested on [Uniswap / permit2](https://github.com/Uniswap/permit2). Takes the same time to compile as the pre-V1 Forge-Std version used there.
@marktoda, could you please test the changes on your end?

### What's new

- Add `StdChains` Component
- **BREAKING:** Add `getChain` and `setChain` functions
- **BREAKING:** `Vm.Rpc.name` -> `Vm.Rpc.key`

@mds1, please let me know what you think about the design decisions.

### DIFF
--- a/src/StdChains.sol
+++ b/src/StdChains.sol
@@ -20,13 +20,16 @@ abstract contract StdChains {
         string rpcUrl;
     }
 
-    // Maps from a chain's name (matching what's in the `foundry.toml` file) to chain data.
+    bool private initialized;
+
+    // Maps from a chain's key (matching the alias in the `foundry.toml` file) to chain data.
     mapping(string => Chain) private chains;
+    // Maps from a chain ID to that chain's key name
     mapping(uint256 => string) private idToKey;
 
-    function getChain(string memory name) internal virtual returns (Chain memory) {
+    function getChain(string memory key) internal virtual returns (Chain memory) {
         initialize();
-        return chains[name];
+        return chains[key];
     }
 
     function getChain(uint256 chainId) internal virtual returns (Chain memory) {
@@ -38,8 +41,6 @@ abstract contract StdChains {
         chains[key] = Chain(name, chainId, rpcUrl);
         idToKey[chainId] = key;
     }
-
-    bool private initialized;
 
     function initialize() private {
         if (initialized) return;
@@ -65,7 +66,7 @@ abstract contract StdChains {
         // Loop over RPC URLs in the config file to replace the default RPC URLs
         Vm.Rpc[] memory rpcs = vm.rpcUrlStructs();
         for (uint256 i = 0; i < rpcs.length; i++) {
-            chains[rpcs[i].name].rpcUrl = rpcs[i].url;
+            chains[rpcs[i].key].rpcUrl = rpcs[i].url;
         }
 
         initialized = true;
--- a/src/StdChains.sol
+++ b/src/StdChains.sol
@@ -9,7 +9,7 @@ abstract contract StdChains {
     VmSafe private constant vm = VmSafe(address(uint160(uint256(keccak256("hevm cheat code")))));
 
     struct Chain {
-        // The chain name, using underscores as the separator to match `foundry.toml` conventions.
+        // The chain name.
         string name;
         // The chain's Chain ID.
         uint256 chainId;
@@ -22,14 +22,21 @@ abstract contract StdChains {
 
     // Maps from a chain's name (matching what's in the `foundry.toml` file) to chain data.
     mapping(string => Chain) private chains;
+    mapping(uint256 => string) private idToKey;
 
     function getChain(string memory name) internal virtual returns (Chain memory) {
         initialize();
         return chains[name];
     }
 
-    function setChain(string memory name, uint256 chainId, string memory rpcUrl) internal virtual {
-        chains[name] = Chain(name, chainId, rpcUrl);
+    function getChain(uint256 chainId) internal virtual returns (Chain memory) {
+        initialize();
+        return chains[idToKey[chainId]];
+    }
+
+    function setChain(string memory key, string memory name, uint256 chainId, string memory rpcUrl) internal virtual {
+        chains[key] = Chain(name, chainId, rpcUrl);
+        idToKey[chainId] = key;
     }
 
     bool private initialized;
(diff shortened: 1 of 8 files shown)


src/StdChains.sol
```diff
@@ -0,0 +1,66 @@
...
+pragma experimental ABIEncoderV2;
+
+import "./Vm.sol";
+
+abstract contract StdChains {
+    VmSafe private constant vm = VmSafe(address(uint160(uint256(keccak256("hevm cheat code")))));
+
+    struct Chain {
+        // The chain name, using underscores as the separator to match `foundry.toml` conventions.
+        string name;
```
mds1, line 13: This comment is misleading, since this `name` is actual the real name like "Optimism Goerli" but the key name used is `optimism_goerli`. Similarly, the `setChain` method needs another param to differentiate the mapping key name from the chain's name

src/StdChains.sol
```diff
@@ -0,0 +1,66 @@
...
+        // NOTE: This default RPC URL is included for convenience to facilitate quick tests and
+        // experimentation. Do not use this RPC URL for production test suites, CI, or other heavy
+        // usage as you will be throttled and this is a disservice to others who need this endpoint.
+        string rpcUrl;
+    }
+
+    // Maps from a chain's name (matching what's in the `foundry.toml` file) to chain data.
+    mapping(string => Chain) private chains;
+
+    function getChain(string memory name) internal virtual returns (Chain memory) {
```
mds1, line 26: This is kinda nice because we can also have overloads to get chains by chain ID for user's who want to keep the getter more brief. Of course we'd need a second mapping to map from uint to `Chain` instead, but that might be nice to have.

On a related note, wondering if we should call this method `chains` (and rename the mapping) instead of `getChains`, e.g. which to go with of:
- `getChains("mainnet").rpcUrl`
- `chains("mainnet").rpcUrl`

src/StdChains.sol
```diff
@@ -0,0 +1,66 @@
...
+    function setChain(string memory name, uint256 chainId, string memory rpcUrl) internal virtual {
+        chains[name] = Chain(name, chainId, rpcUrl);
+    }
+
+    bool private initialized;
+
+    function initialize() private {
+        if (initialized) return;
+
+        chains["anvil"] = Chain("Anvil", 31337, "http://127.0.0.1:8545");
```
mds1, line 40: Probably should use the `setChain` method here, especially if that might write to two mappings now instead of one

marktoda:
Hmm not seeing a huge speedup on my end, could be doing something wrong
```
> forge build
[⠒] Compiling...
[⠢] Compiling 38 files with 0.8.17
[⠔] Solc 0.8.17 finished in 122.14s

> cd lib/forge-std
> git fetch

> git checkout feat/stdchains

Switched to a new branch 'feat/stdchains'
> cd ../../
> forge build
[⠒] Compiling...
[⠒] Compiling 42 files with 0.8.17
[⠃] Solc 0.8.17 finished in 123.58s
```

ZeroEkkusu:
@marktoda, that's for Permit 2, isn't it?

Permit 2 uses [`27e14b7`](https://github.com/foundry-rs/forge-std/commit/27e14b7f2448e5f5ac32719f51fe652aa0b0733e), which is a pre-V1 Forge Standard. The new changes take the same time to compile as that old version (123 seconds), whereas the latest `master` would take much longer.

Correct me if I'm wrong (the original problem was that V1 compiled slow `via-ir` and we wanted to achieve pre-V1 compile times).

I'm also interested in benchmarking against your https://github.com/foundry-rs/forge-std/issues/225#issue-1454382836.

marktoda:
AH, yes you're right

ZeroEkkusu:
Thanks!

To recap, Hardhat compiles in 16s (contracts only), and Forge does it in 20s (contracts + tests + scripts).

Pinging @mds1.

There may be a way to further speed up Forge Standard (compilation).

src/StdChains.sol
ZeroEkkusu: Great idea. Left `getChain` because it "gets" it from the config (at least the initial ones). Feel free to change.

There is a problem if two Chains have the same chain ID (e.g. Anvil and Hardhat) - last Chain will be returned. We can print a warning when the user sets another chain with existing chain ID.

ZeroEkkusu:
Re optimizing: I implemented console workarounds. It took 3 seconds less to compile Permit 2 (137 seconds)- not pushing those changes.

@mds1, see the open discussion.

Otherwise, I think this PR is ready. See #239 as well.

src/StdChains.sol
mds1: > There is a problem if two Chains have the same chain ID (e.g. Anvil and Hardhat) - last Chain will be returned. We can print a warning when the user sets another chain with existing chain ID (or remove the functionality completely so that someone doesn't make a mistake and deploy to wrong chain, etc).

This seems ok to me. I don't think there's any risk of deploying to the wrong chain here, since this would really only happen with anvil/hardhat, otherwise no production chains should duplicate chain ID. Worth noting this in the docs/code comments though, just in case someone adds e.g. two mainnet entries where one is Infura RPC and the other is flashbots

src/StdChains.sol
```diff
@@ -0,0 +1,73 @@
...
+        // A default RPC endpoint for this chain.
+        // NOTE: This default RPC URL is included for convenience to facilitate quick tests and
+        // experimentation. Do not use this RPC URL for production test suites, CI, or other heavy
+        // usage as you will be throttled and this is a disservice to others who need this endpoint.
+        string rpcUrl;
+    }
+
+    // Maps from a chain's name (matching what's in the `foundry.toml` file) to chain data.
+    mapping(string => Chain) private chains;
+    mapping(uint256 => string) private idToKey;
```
mds1, line 25: Nit just to clarify the relation here a bit 
```suggestion
    // Maps from a chain's key (matching the alias in the `foundry.toml` file) to chain data.
    mapping(string => Chain) private chains;
    
    // Maps from a chain ID to that chain's key name
    mapping(uint256 => string) private idToKey;
```

src/StdChains.sol
```diff
@@ -0,0 +1,73 @@
...
+        initialize();
+        return chains[idToKey[chainId]];
+    }
+
+    function setChain(string memory key, string memory name, uint256 chainId, string memory rpcUrl) internal virtual {
+        chains[key] = Chain(name, chainId, rpcUrl);
+        idToKey[chainId] = key;
+    }
+
+    bool private initialized;
```
mds1, line 42: Let's move this state var declaration right below `idToKey` so all state vars are together. Not too common IME to have state vars below function declarations like this

src/StdChains.sol
```diff
@@ -0,0 +1,73 @@
...
+        setChain("avalanche", "Avalanche", 43114, "https://api.avax.network/ext/bc/C/rpc");
+        setChain("avalanche_fuji", "Avalanche Fuji", 43113, "https://api.avax-test.network/ext/bc/C/rpc");
+        setChain("bnb_smart_chain", "BNB Smart Chain", 56, "https://bsc-dataseed1.binance.org");
+        setChain("bnb_smart_chain_testnet", "BNB Smart Chain Testnet", 97, "https://data-seed-prebsc-1-s1.binance.org:8545");// forgefmt: disable-line
+        setChain("gnosis_chain", "Gnosis Chain", 100, "https://rpc.gnosischain.com");
+
+        // Loop over RPC URLs in the config file to replace the default RPC URLs
+        Vm.Rpc[] memory rpcs = vm.rpcUrlStructs();
+        for (uint256 i = 0; i < rpcs.length; i++) {
+            chains[rpcs[i].name].rpcUrl = rpcs[i].url;
```
mds1, line 68: Two other questions:

1. Does this PR also close https://github.com/foundry-rs/forge-std/issues/232? If an env var in the toml is undefined, my guess is `rpcs[i].url` would be empty in which case we should check for that here and avoid overwriting the value? I see @adhusson opened https://github.com/foundry-rs/forge-std/pull/238 but haven't looked at it closely yet
2. Wdyt of updating the `Vm.sol` `Rpc` struct to change the field from `name` to `alias`, to remove naming ambiguity (I though this line was using the `Chain` struct's `name` property at first read, but really it's the alias). Since, if we're going to introduce breaking changes in this PR, may as well get them all out of the way

src/StdChains.sol
ZeroEkkusu: My concern is that the user can make an error and end up with multiple entries with the same chain ID, in which case last chain will be returned when getting by chain ID. Let's ping someone from Foundry for opinion. @mds1

src/StdChains.sol
ZeroEkkusu: `alias` is a reserved keyword in Solidity. Changed it to `key`.

ZeroEkkusu: @adhusson Do you want to address the first point by modifying and merging your PR into this branch?

src/StdChains.sol
mds1: Right, I don't think it's a big risk since chains have different IDs, and the RPC is the source of truth, but mitigating it is probably better anyway. Some options:
1. When adding a chain in `setChain`, remove the `chainId` input arg and fetch it from the RPC by selecting a fork with that URL then switching back to the prior context. This ensures chains always have the correct chain ID set. We should bypass this and set manually in `initialize()` to avoid RPC calls since we are sure that initial config present is correct, e.g. maybe a private `setChain` overload is used in `initialize` and then we expose a different one for testing
2. Remove the default `hardhat` chain. Then update `setChain` to only support a single instance of a chain ID, e.g. if my test calls `setChain("hardhat", "http://127.0.0.1:8545")` this _replaces_ the existing `anvil` chain and emits a warning log

src/StdChains.sol
mds1: > `alias` is a reserved keyword in Solidity.

Ah right, that's why I didn't use this originally. `key` works, thanks!

mds1: > Do you want to address the first point by modifying and merging your PR into this branch?

I'd say let's get this branch merged first, where I think https://github.com/foundry-rs/forge-std/pull/235#discussion_r1035303826 is the main decision left

After that @adhusson can update his PR and keep it as merging into master. Smaller PRs into master is easier to follow/reason about IMO, and since nothing is updated without a new release by default anyway, there's not much value add to making this PR a big one by merging into it

adhusson: Sounds good, will update the PR after this is merged.

src/StdChains.sol
ZeroEkkusu: I think `getChain(uint256)` is the problem. Unlike mapping keys, chain IDs may not be unique. We want to support multiple endpoints, for example flashbots, as you said. This means chain IDs can't have 1-1 relation with mapping keys, regardless of `setChain` logic. My proposal is to remove `getChain(uint256)` in favor of multiple endpoints.

mds1: > We want to support multiple endpoints, for example flashbots, as you said.

Hmm, well this was just an idea I had as to when you may when multiple endpoints for a single chain ID. But no one has requested this yet, and if unsupported you can workaround it with repeated `setChain` + `createSelectFork` use. So I think we should only support one entry per chain ID to keep this feature simple/safer and avoid over-engineering it to support a niche feature that we're not even sure anyone has a use case for

ZeroEkkusu: Remove `hardhat` and add revert to `setChain`, then?

mds1: I'd say so, yea let's keep anvil as the default and people can always `setChain` to update the name/RPC URL. My suggestion is:
- Use chain ID as the source of truth for a chain
- Map chain IDs to the key names (where key name = the alias in foundry.toml)
- Have mappings from (1) chain ID to `Chain` struct, and (2) from key name to `Chain` struct
- Expose getters to get `Chain` info by either chain ID or key name
- Expose setter that takes key name + RPC URL, selects that fork, fetches chain ID, switches back to original fork/context (not sure if this is currently possible, let's check with @mattsse) (this also serves to validate the RPC URL)
- Setter can both overwrite an existing RPC or set a new one
- Our `initialize` method should _not_ use that setter, because we don't want dozens of chain ID RPC requests on initialization

Sorry for the churn here, but lmk how that sounds to you

ZeroEkkusu: A mapping from chain ID to `Chain` is redundant. I think the current set up is better: chain ID -> key -> `Chain`.

mds1: That's fine! I just meant conceptually both should exist in that the user can access `Chain` info from either approach.

Also it seems there's currently no way to switch back to a no-fork context, so for now let's just keep `setChain(name, key, chainId, rpcUrl)`, and we can add the no-chainId version in the future

src/StdChains.sol
```diff
@@ -0,0 +1,74 @@
...
+    function setChain(string memory key, string memory name, uint256 chainId, string memory rpcUrl) internal virtual {
+        chains[key] = Chain(name, chainId, rpcUrl);
+        idToKey[chainId] = key;
+    }
+
+    function initialize() private {
+        if (initialized) return;
+
+        setChain("anvil", "Localhost", 31337, "http://127.0.0.1:8545");
+        setChain("hardhat", "Localhost", 31337, "http://127.0.0.1:8545");
```
mds1, line 49: Let's remove this since hardhat and anvil have the same chain ID, and users can overwrite the info if needed

mds1 (commented):
LGTM, thanks! Left one comment, which I think is the last thing before merging?

test/StdChains.t.sol
```diff
@@ -25,6 +24,16 @@ contract StdChainsTest is Test {
         }
     }
 
+    function testCannotSetChain_ChainIdExists() public {
+        setChain({key: "custom_chain", name: "Custom Chain", chainId: 123456789, rpcUrl: "https://custom.chain/"});
+
+        vm.expectRevert(
+            'StdChains setChain(string,string,uint256,string): Chain ID 123456789 already used by "custom_chain"'
+        );
```
mds1, line 32: Does this mean users can't change the name of the default `Anvil` chain to `Hardhat`?

ZeroEkkusu: Name, as in `Chain.name`? Yes.

ZeroEkkusu: What they cannot do is add `"hardhat"` Chain that has the same chain ID as `"anvil"`.

ZeroEkkusu: This lets you:

- add a new Chain (must have unique chain ID)
- modify an existing chain

ZeroEkkusu: If good merge.

mds1: Great, thanks!

PR #236, opened 2022-11-22
feature: isFork and skip modifiers in StdCheats

prestwich:
## Motivation

Allow tests to easily change behavior (or be skipped) when running in forking or non-forking mode

## Solution

Adds convenience methods to `StdCheats`:

- `isFork` - returns the current forking status
- `skipNonForking` - run only when forked 
- `skipForking` - run only when not forked

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -569,4 +569,22 @@ abstract contract StdCheats is StdCheatsSafe {
             stdstore.target(token).sig(0x18160ddd).checked_write(totSup);
         }
     }
+
+    function isFork() internal virtual returns (bool status) {
+        try vm.activeFork() {
+            status = true;
+        } catch (bytes memory) {}
+    }
+
+    modifier skipForking() {
+        if (!isFork()) {
+            _;
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


src/StdCheats.sol
```diff
@@ -569,4 +569,22 @@ abstract contract StdCheats is StdCheatsSafe {
...
         }
     }
+
+    function isFork() internal virtual returns (bool status) {
+        try vm.activeFork() {
+            status = true;
+        } catch (bytes memory) {}
+    }
+
+    modifier skipForking() {
```
mds1, line 579: Nit on naming, wdyt of `skipWhenForking` and `skipWhenNotForking` instead (or `skipIfForking` and `skipIfNotForking`)? `skipForking` is a bit less clear, e.g. can be interpreted as the modifier resulting in fork-related cheats not executing / having different behavior.

Otherwise LGTM, test failure unrelated.

prestwich: ya, `when` is better naming. changed 👍

mds1 (approved):
LGTM, test failure unrelated. Thanks!

PR #237, opened 2022-11-23
feat: add getEnv vm.env helpers with default fallbacks

jameswenzel:
Added a suite of `getEnvX` functions to `stdCheats` – let me know if this is not a good place for them.

They're potentially useful for publishing repos with configurable tests or scripts (where someone who initially pulls may not have an `.env` set up), or where adding secrets to a repo doesn't necessarily make sense.

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -466,6 +466,146 @@ abstract contract StdCheatsSafe {
         who = vm.rememberKey(privateKey);
     }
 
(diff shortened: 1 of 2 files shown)


mds1:
I think we should close this, as per https://github.com/foundry-rs/foundry/issues/3732 it looks like this functionality will be added natively as a cheatcode. I think that is preferable because it's "free" whereas the more code we add to test contracts the longer compilation will take

PR #241, opened 2022-11-27
fix: remove StdStorage in CommonBase contract

graykode:
Remove `StdStorage` in [Common.sol](https://github.com/foundry-rs/forge-std/blob/master/src/Common.sol)
If there is `StdStorage` in Common.sol, this testcase cant return ERROR when run with forge

```solidity
pragma solidity ^0.8.0;

import "forge-std/Test.sol";

contract good is Test {
    uint256 a;

    function testAAA() public {
        uint256 id = vm.snapshot();
        a = 1;
        assertEq(a, uint256(2));
        vm.revertTo(id);
    }
}
```
image

### DIFF
--- a/src/Common.sol
+++ b/src/Common.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.2 <0.9.0;
 
-import {StdStorage, Vm} from "./Components.sol";
+import {Vm} from "./Components.sol";
 
 abstract contract CommonBase {
     // Cheat code address, 0x7109709ECfa91a80626fF3989D68f67F5b1DD12D.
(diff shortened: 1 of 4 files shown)


mds1:
Holding off on merging this per discussion in https://github.com/foundry-rs/foundry/issues/3055 — let's move discussion there, cc @ZeroEkkusu if you have any thoughts on this PR

ZeroEkkusu (changes requested):
Initializing a `struct` before `DSTest` is inherited causes the example to pass.

Since it's a false positive (dangerous), let's fix the issue immediately. @graykode, could you please move `TestBase` to the of the inheritance chain instead:

```solidity
abstract contract Test is DSTest, StdAssertions, StdCheats, StdUtils, TestBase {}
```

mattsse (approved):
agree, let's merge and I'll debug the root cause.

@mds1 probably need another patch release?

mds1:
@ZeroEkkusu I pushed the required changes and verified the test now reverts as expected. Feel free to review/merge

ZeroEkkusu (approved):

PR #246, opened 2022-12-01
Name arguments in Vm.sol

paulrberg:
Benefits:

1. Enable Forge Std users to call cheatcodes with [named arguments](https://twitter.com/PaulRBerg/status/1574071928544976896), which improves readability.
2. Better documentation for people who read the source code.

Examples:

```solidity
vm.roll({ newHeight: block.height + 100 });
vm.getCode({ path: "../../path/to/MyContract.sol" });
vm.expectEmit({ checkTopic1: true, checkTopic2: true, checkTopic3: true, checkData: true });
```

I think that this is especially useful for the `expectEmit` cheatcode, which, to quote the [Foundry Book](https://book.getfoundry.sh/forge/cheatcodes), is known for being "non-intuitive".

Notes:

- Generally I have tried to follow the argument names suggested in the comments
- When there was no name suggestion in the comments, I came up with my own (feedback welcome)

Questions:

1. Should we keep the argument names in the comments above the functions? They look a bit superfluous now.
2. I am torn between `txid` and `txhash`, which is needed in multiple functions. Which one do you like more?

### DIFF
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -68,13 +68,27 @@ interface VmSafe {
     function envOr(string calldata name, string calldata defaultValue) external returns (string memory value);
     function envOr(string calldata name, bytes calldata defaultValue) external returns (bytes memory value);
     // Read environment variables as arrays with default value, (name, value[]) => (value[])
-    function envOr(string calldata name, string calldata, bool[] calldata defaultValue) external returns (bool[] memory value);
-    function envOr(string calldata name, string calldata, uint256[] calldata defaultValue) external returns (uint256[] memory value);
-    function envOr(string calldata name, string calldata, int256[] calldata defaultValue) external returns (int256[] memory value);
-    function envOr(string calldata name, string calldata, address[] calldata defaultValue) external returns (address[] memory value);
-    function envOr(string calldata name, string calldata, bytes32[] calldata defaultValue) external returns (bytes32[] memory value);
-    function envOr(string calldata name, string calldata, string[] calldata defaultValue) external returns (string[] memory value);
-    function envOr(string calldata name, string calldata, bytes[] calldata defaultValue) external returns (bytes[] memory value);
+    function envOr(string calldata name, string calldata, bool[] calldata defaultValue)
+        external
+        returns (bool[] memory value);
+    function envOr(string calldata name, string calldata, uint256[] calldata defaultValue)
+        external
+        returns (uint256[] memory value);
+    function envOr(string calldata name, string calldata, int256[] calldata defaultValue)
+        external
+        returns (int256[] memory value);
+    function envOr(string calldata name, string calldata, address[] calldata defaultValue)
+        external
+        returns (address[] memory value);
+    function envOr(string calldata name, string calldata, bytes32[] calldata defaultValue)
+        external
+        returns (bytes32[] memory value);
+    function envOr(string calldata name, string calldata, string[] calldata defaultValue)
+        external
+        returns (string[] memory value);
+    function envOr(string calldata name, string calldata, bytes[] calldata defaultValue)
+        external
+        returns (bytes[] memory value);
     // Records all storage reads and writes
     function record() external;
     // Gets all accessed reads and write slot from a recording session, for a given address
@@ -322,7 +336,7 @@ interface Vm is VmSafe {
     // Revert the state of the EVM to a previous snapshot
     // Takes the snapshot id to revert to.
     // This deletes the snapshot and all snapshots taken after the given snapshot id.
-    function revertTo(uint256 snapshotId) external returns (bool result);
+    function revertTo(uint256 snapshotId) external returns (bool success);
     // Creates a new fork with the given endpoint and block and returns the identifier of the fork
     function createFork(string calldata endpoint, uint256 blockNumber) external returns (uint256 forkId);
     // Creates a new fork with the given endpoint and the _latest_ block and returns the identifier of the fork
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -4,8 +4,8 @@ pragma solidity >=0.6.2 <0.9.0;
 pragma experimental ABIEncoderV2;
 
 // Cheatcodes are marked as view/pure/none using the following rules:
-// 0. A call's observable behaviour includes its return value, logs, reverts and state writes.
-// 1. If you can influence a later call's observable behaviour, you're neither `view` nor `pure` (you are modifying some state be it the EVM, interpreter, filesystem, etc),
+// 0. A call's observable behaviour includes its return value, logs, reverts and state writes,
+// 1. If you can influence a later call's observable behaviour, you're neither `view` nor `pure (you are modifying some state be it the EVM, interpreter, filesystem, etc),
 // 2. Otherwise if you can be influenced by an earlier call, or if reading some state, you're `view`,
 // 3. Otherwise you're `pure`.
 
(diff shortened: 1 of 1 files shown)


mds1:
Definitely better UX, thanks for this. I'll take a closer look at all the param names later. Also needs `forge fmt` 

> Should we keep the argument names in the comments above the functions? They look a bit superfluous now.

I'm indifferent here, I think I agree with removing them to make maintenance easier. Though we should probably also name return parameters too, which are currently missing

> I am torn between `txid` and `txhash`, which is needed in multiple functions. Which one do you like more?

I'd go with `txHash` personally, I think that's the most common

paulrberg:
Thanks, glad you like it.

> Also needs `forge fmt`

Oh, shoot, I don't have it set up on VSCode yet. I applied `forge gmt` in my last commit.

> removing them to make maintenance easier.

Makes sense to me.

> name return parameters too, which are currently missing

Agree. I will try to do this later.

> I'd go with txHash personally, I think that's the most common

Good, changed in my last commit.

paulrberg:
@mds1 I pushed another commit to name the return arguments.

While doing this, I realized that there is no explanation for the return argument of the `revertTo` cheatcode. I didn't know what to call it so I set as `result`, but I will let you decide if that makes sense or not based on the return value actually is.

src/Vm.sol
```diff
@@ -160,151 +207,131 @@ interface VmSafe {
...
     // Returns the id of the snapshot that was created.
     // To revert a snapshot use `revertTo`
-    function snapshot() external returns (uint256);
-    // Revert the state of the evm to a previous snapshot
+    function snapshot() external returns (uint256 snapshotId);
+    // Revert the state of the EVM to a previous snapshot
     // Takes the snapshot id to revert to.
     // This deletes the snapshot and all snapshots taken after the given snapshot id.
-    function revertTo(uint256) external returns (bool);
+    function revertTo(uint256 snapshotId) external returns (bool result);
```
mds1, line 293: [Seems](https://github.com/foundry-rs/foundry/blob/427c1b549be814b255d8a4dea8fa3c2409dddc4c/evm/src/executor/inspector/cheatcodes/snapshot.rs#L17) this returns `true` if you were able to revert to the given `snapshotId` and false otherwise, e.g. if it doesn't exist
```suggestion
    function revertTo(uint256 snapshotId) external returns (bool success);
```

mds1:
There's also a few new cheatcodes now (such as https://github.com/foundry-rs/forge-std/pull/249) so needs conflicts resolved. @paulrberg if you want to add names for those that would be awesome, if not no problem and I can do it in a follow up PR

src/Vm.sol
paulrberg: Thanks @mds1, fixed in last commit.

paulrberg:
No worries @mds1, I refactored from master and fixed the git conflicts myself.

src/Vm.sol
```diff
@@ -31,133 +31,146 @@ interface VmSafe {
...
+    function envString(string calldata name) external view returns (string memory);
+    function envBytes(string calldata name) external view returns (bytes memory);
+    // Reads environment variables as arrays
+    function envBool(string calldata name, string calldata delim) external view returns (bool[] memory);
+    function envUint(string calldata name, string calldata delim) external view returns (uint256[] memory);
+    function envInt(string calldata name, string calldata delim) external view returns (int256[] memory);
+    function envAddress(string calldata name, string calldata delim) external view returns (address[] memory);
+    function envBytes32(string calldata name, string calldata delim) external view returns (bytes32[] memory);
+    function envString(string calldata name, string calldata delim) external view returns (string[] memory);
+    function envBytes(string calldata name, string calldata delim) external view returns (bytes[] memory);
```
mds1, line 61: Do we want to give these a named return of `value` as we did with `envOr`?

mds1 (approved):
Looks great, found one missing a named return and once resolved we can merge!

src/Vm.sol
paulrberg: Yes. Actually they used to have the return value named, but I botched the rafactor.

paulrberg: Fixed in the last commit.

mds1 (approved):
LGTM, thanks! Will wait for @ZeroEkkusu to take one more look before merging

paulrberg:
I have just realized that the Foundry Book had already named the arguments, e.g. for [`createSelectFork`](https://book.getfoundry.sh/cheatcodes/create-select-fork):

Screenshot 2022-12-06 at 5 19 15 PM

Should we sync this PR with the Foundry Book?

I think it's @onbjerg who provided the names in the book.

mds1:
Ah interesting! I'd say let's go with whichever is more descriptive/self-documenting for this PR, and we can always update the book to match

paulrberg:
I might not have the time to go through each entry in the Foundry Book and update the args one-by-one.

I guess we could just this merge this PR as-is and update the arg names on the go?

ZeroEkkusu:
Self-documentation > predictability ? @paulrberg

ZeroEkkusu:
Thanks. I will review it for self-documentation first.

src/Vm.sol
```diff
@@ -188,156 +249,135 @@ interface VmSafe {
...
+    function roll(uint256 newHeight) external;
+    // Sets block.basefee
+    function fee(uint256 newBasefee) external;
+    // Sets block.difficulty
+    function difficulty(uint256 newDifficulty) external;
     // Sets block.chainid
-    function chainId(uint256) external;
-    // Stores a value to an address' storage slot, (who, slot, value)
-    function store(address, bytes32, bytes32) external;
+    function chainId(uint256 newChainId) external;
```
ZeroEkkusu, line 286: Same as in Solidity (`block.chainid`).

```suggestion
    function chainId(uint256 newChainid) external;
```

src/Vm.sol
```diff
@@ -188,156 +249,135 @@ interface VmSafe {
...
+    // Stores a value to an address' storage slot.
+    function store(address who, bytes32 slot, bytes32 value) external;
```
ZeroEkkusu, line 288:
```suggestion
    function store(address target, bytes32 slot, bytes32 value) external;
```

src/Vm.sol
```diff
@@ -188,156 +249,135 @@ interface VmSafe {
...
     // Sets the nonce of an account; must be higher than the current nonce of the account
-    function setNonce(address, uint64) external;
+    function setNonce(address account, uint64 newNonce) external;
```
ZeroEkkusu, line 290:
```suggestion
    function setNonce(address target, uint64 newNonce) external;
```

src/Vm.sol
```diff
@@ -188,156 +249,135 @@ interface VmSafe {
...
-    function startPrank(address, address) external;
+    function startPrank(address msgSender, address txOrigin) external;
     // Resets subsequent calls' msg.sender to be `address(this)`
     function stopPrank() external;
-    // Sets an address' balance, (who, newBalance)
-    function deal(address, uint256) external;
-    // Sets an address' code, (who, newCode)
-    function etch(address, bytes calldata) external;
+    // Sets an address' balance
+    function deal(address who, uint256 newBalance) external;
```
ZeroEkkusu, line 302:
```suggestion
    function deal(address target, uint256 newBalance) external;
```

src/Vm.sol
```diff
@@ -188,156 +249,135 @@ interface VmSafe {
...
+    // Sets an address' code
+    function etch(address who, bytes calldata newCode) external;
```
ZeroEkkusu, line 304:
```suggestion
    function etch(address target, bytes calldata newRuntimeBytecode) external;
```

src/Vm.sol
```diff
@@ -188,156 +249,135 @@ interface VmSafe {
...
+    function expectRevert(bytes calldata revertData) external;
+    function expectRevert(bytes4 revertData) external;
     function expectRevert() external;
     // Prepare an expected log with (bool checkTopic1, bool checkTopic2, bool checkTopic3, bool checkData).
     // Call this function, then emit an event, then call a function. Internally after the call, we check if
     // logs were emitted in the expected order with the expected topics and data (as specified by the booleans)
-    function expectEmit(bool, bool, bool, bool) external;
-    function expectEmit(bool, bool, bool, bool, address) external;
+    function expectEmit(bool checkTopic1, bool checkTopic2, bool checkTopic3, bool checkData) external;
+    function expectEmit(bool checkTopic1, bool checkTopic2, bool checkTopic3, bool checkData, address emitter)
```
ZeroEkkusu, line 313:
```suggestion
    function expectEmit(bool checkTopic1, bool checkTopic2, bool checkTopic3, bool checkData, address target)
```

src/Vm.sol
```diff
@@ -188,156 +249,135 @@ interface VmSafe {
...
+        external;
     // Mocks a call to an address, returning specified data.
     // Calldata can either be strict or a partial match, e.g. if you only
     // pass a Solidity selector to the expected calldata, then the entire Solidity
     // function will be mocked.
-    function mockCall(address, bytes calldata, bytes calldata) external;
+    function mockCall(address callee, bytes calldata data, bytes calldata returnData) external;
```
ZeroEkkusu, line 319:
```suggestion
    function mockCall(address target, bytes calldata data, bytes calldata returnData) external;
```

src/Vm.sol
```diff
@@ -188,156 +249,135 @@ interface VmSafe {
...
     // Mocks a call to an address with a specific msg.value, returning specified data.
     // Calldata match takes precedence over msg.value in case of ambiguity.
-    function mockCall(address, uint256, bytes calldata, bytes calldata) external;
+    function mockCall(address callee, uint256 msgValue, bytes calldata data, bytes calldata returnData) external;
```
ZeroEkkusu, line 322:
```suggestion
    function mockCall(address target, uint256 msgValue, bytes calldata data, bytes calldata returnData) external;
```

src/Vm.sol
```diff
@@ -188,156 +249,135 @@ interface VmSafe {
...
     // Clears all mocked calls
     function clearMockedCalls() external;
     // Expects a call to an address with the specified calldata.
     // Calldata can either be a strict or a partial match
-    function expectCall(address, bytes calldata) external;
+    function expectCall(address callee, bytes calldata data) external;
```
ZeroEkkusu, line 327:
```suggestion
    function expectCall(address target, bytes calldata data) external;
```

src/Vm.sol
```diff
@@ -188,156 +249,135 @@ interface VmSafe {
...
     // Expects a call to an address with the specified msg.value and calldata
-    function expectCall(address, uint256, bytes calldata) external;
-    // Sets block.coinbase (who)
-    function coinbase(address) external;
+    function expectCall(address callee, uint256 msgValue, bytes calldata data) external;
```
ZeroEkkusu, line 329:
```suggestion
    function expectCall(address target, uint256 msgValue, bytes calldata data) external;
```

src/Vm.sol
```diff
@@ -188,156 +249,135 @@ interface VmSafe {
...
-    // Revert the state of the evm to a previous snapshot
+    function snapshot() external returns (uint256 snapshotId);
+    // Revert the state of the EVM to a previous snapshot
     // Takes the snapshot id to revert to.
     // This deletes the snapshot and all snapshots taken after the given snapshot id.
-    function revertTo(uint256) external returns (bool);
+    function revertTo(uint256 snapshotId) external returns (bool success);
     // Creates a new fork with the given endpoint and block and returns the identifier of the fork
-    function createFork(string calldata, uint256) external returns (uint256);
+    function createFork(string calldata endpoint, uint256 blockNumber) external returns (uint256 forkId);
```
ZeroEkkusu, line 341:
```suggestion
    function createFork(string calldata url, uint256 blockNumber) external returns (uint256 forkId);
```

src/Vm.sol
```diff
@@ -188,156 +249,135 @@ interface VmSafe {
...
     // Creates a new fork with the given endpoint and the _latest_ block and returns the identifier of the fork
-    function createFork(string calldata) external returns (uint256);
-    // Creates a new fork with the given endpoint and at the block the given transaction was mined in, and replays all transaction mined in the block before the transaction
-    function createFork(string calldata, bytes32) external returns (uint256);
+    function createFork(string calldata endpoint) external returns (uint256 forkId);
```
ZeroEkkusu, line 343:
```suggestion
    function createFork(string calldata url) external returns (uint256 forkId);
```

src/Vm.sol
```diff
@@ -188,156 +249,135 @@ interface VmSafe {
...
+    // Creates a new fork with the given endpoint and at the block the given transaction was mined in, replays all transaction mined in the block before the transaction,
+    // and returns the identifier of the fork
+    function createFork(string calldata endpoint, bytes32 txHash) external returns (uint256 forkId);
```
ZeroEkkusu, line 346:
```suggestion
    function createFork(string calldata url, bytes32 txHash) external returns (uint256 forkId);
```

src/Vm.sol
```diff
@@ -188,156 +249,135 @@ interface VmSafe {
...
     // Creates _and_ also selects a new fork with the given endpoint and block and returns the identifier of the fork
-    function createSelectFork(string calldata, uint256) external returns (uint256);
-    // Creates _and_ also selects new fork with the given endpoint and at the block the given transaction was mined in, and replays all transaction mined in the block before the transaction
-    function createSelectFork(string calldata, bytes32) external returns (uint256);
+    function createSelectFork(string calldata endpoint, uint256 blockNumber) external returns (uint256 forkId);
```
ZeroEkkusu, line 348:
```suggestion
    function createSelectFork(string calldata url, uint256 blockNumber) external returns (uint256 forkId);
```

src/Vm.sol
```diff
@@ -188,156 +249,135 @@ interface VmSafe {
...
+    // Creates _and_ also selects new fork with the given endpoint and at the block the given transaction was mined in, replays all transaction mined in the block before
+    // the transaction, returns the identifier of the fork
+    function createSelectFork(string calldata endpoint, bytes32 txHash) external returns (uint256 forkId);
```
ZeroEkkusu, line 351:
```suggestion
    function createSelectFork(string calldata url, bytes32 txHash) external returns (uint256 forkId);
```

ZeroEkkusu (commented):
Sorry for the wait.

Good work.

Some suggestions: keep in mind I don't have a strong opinion about these - consider them, and discard all that aren't good.

src/Vm.sol
```diff
@@ -31,133 +31,146 @@ interface VmSafe {
...
     }
 
-    // Loads a storage slot from an address (who, slot)
-    function load(address, bytes32) external view returns (bytes32);
-    // Signs data, (privateKey, digest) => (v, r, s)
-    function sign(uint256, bytes32) external pure returns (uint8, bytes32, bytes32);
-    // Gets the address for a given private key, (privateKey) => (address)
-    function addr(uint256) external pure returns (address);
+    // Loads a storage slot from an address
+    function load(address who, bytes32 slot) external view returns (bytes32 data);
```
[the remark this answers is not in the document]
mds1: Instead of target I'd suggest `contract` for the `load` and `store` cheats, since they aren't intended for use with EOAs

src/Vm.sol
```diff
@@ -31,133 +31,146 @@ interface VmSafe {
...
+    function sign(uint256 privateKey, bytes32 digest) external pure returns (uint8 v, bytes32 r, bytes32 s);
+    // Gets the address for a given private key
+    function addr(uint256 privateKey) external pure returns (address addr);
     // Gets the nonce of an account
-    function getNonce(address) external view returns (uint64);
-    // Performs a foreign function call via the terminal, (stringInputs) => (result)
-    function ffi(string[] calldata) external returns (bytes memory);
-    // Sets environment variables, (name, value)
-    function setEnv(string calldata, string calldata) external;
+    function getNonce(address account) external view returns (uint64 nonce);
```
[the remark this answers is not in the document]
mds1: Personally prefer `account` here since it's standard nomenclature

src/Vm.sol
```diff
@@ -31,133 +31,146 @@ interface VmSafe {
...
+        external
+        returns (string[] memory value);
+    function envOr(string calldata name, string calldata, bytes[] calldata defaultValue)
+        external
+        returns (bytes[] memory value);
     // Records all storage reads and writes
     function record() external;
     // Gets all accessed reads and write slot from a recording session, for a given address
-    function accesses(address) external returns (bytes32[] memory reads, bytes32[] memory writes);
+    function accesses(address who) external returns (bytes32[] memory reads, bytes32[] memory writes);
```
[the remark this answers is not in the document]
mds1: `contract` instead of `target makes sense here for the same reason IMO

src/Vm.sol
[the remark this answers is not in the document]
mds1: Oh right contract is a keyword, so maybe just `account`?

src/Vm.sol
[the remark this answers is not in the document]
mds1: Same as https://github.com/foundry-rs/forge-std/pull/246#discussion_r1041668939

src/Vm.sol
```diff
@@ -31,133 +31,146 @@ interface VmSafe {
...
     // Gets the _creation_ bytecode from an artifact file. Takes in the relative path to the json file
-    function getCode(string calldata) external view returns (bytes memory);
+    function getCode(string calldata artifactPath) external view returns (bytes memory bytecode);
     // Gets the _deployed_ bytecode from an artifact file. Takes in the relative path to the json file
-    function getDeployedCode(string calldata) external view returns (bytes memory);
+    function getDeployedCode(string calldata artifactPath) external view returns (bytes memory bytecode);
     // Labels an address in call traces
-    function label(address, string calldata) external;
+    function label(address who, string calldata newLabel) external;
```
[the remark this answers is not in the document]
mds1: Maybe `account` again? `target` _feels_ a bit off though I don't have a great rationale 😅

src/Vm.sol
```diff
@@ -188,156 +249,135 @@ interface VmSafe {
...
+    // will find the object in-memory that is keyed by "some key".
+    function writeJson(string calldata json, string calldata path) external;
     // Write a serialized JSON object to an **existing** JSON file, replacing a value with key = <value_key>
     // This is useful to replace a specific value of a JSON file, without having to parse the entire thing
-    // (stringified_json, path, value_key)
-    function writeJson(string calldata, string calldata, string calldata) external;
+    function writeJson(string calldata json, string calldata path, string calldata valueKey) external;
     // Returns the RPC url for the given alias
-    function rpcUrl(string calldata) external view returns (string memory);
+    function rpcUrl(string calldata rpcAlias) external view returns (string memory json);
```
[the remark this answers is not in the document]
mds1: The forge book/docs calls them aliases though, and I do think `alias` is more clear. The fact that we call it `key` in `StdChains` is an implementation detail that we don't need to leak to this interface, so I'd suggest sticking with `rpcAlias`

src/Vm.sol
mds1: definitely a nit, though I prefer the `newChainId` since it's more readable and `chainId` style casing is very common elsewhere

src/Vm.sol
mds1: `urlOrAlias` is probably best as @paulrberg pointed out above, since both are supported

mds1:
Left a few replies to @ZeroEkkusu's suggestions, many I didn't reply to otherwise I'd end up leaving many duplicate comments. Overall I'm mostly indifferent on the suggestions fwiw as all of them are an improvement from the status quo

ZeroEkkusu:
I’ve read @mds1 comments. @paulrberg, I’ll leave it to you to make any final changes, as both of us are indifferent on the suggestions.

And let’s get this merged.

src/Vm.sol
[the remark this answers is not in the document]
paulrberg: Yeah, I like `account` better in this context.

src/Vm.sol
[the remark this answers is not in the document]
paulrberg: I like `account` better in this context. It's EOAs who are usually labelled in tests.

src/Vm.sol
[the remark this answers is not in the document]
paulrberg: I went with `rpcAlias`.

src/Vm.sol
paulrberg: I went with `newChainId`.

src/Vm.sol
paulrberg: `target` is definitely a good fit in this context.

src/Vm.sol
paulrberg: I went with `account` here.

src/Vm.sol
paulrberg: I went with `account` here.

src/Vm.sol
paulrberg: I went with `emitter`, this is what's used in the Foundry Book. And I just like it better (nit).

src/Vm.sol
paulrberg: I like `callee` better in this context because it refers to the action of calling a contract. `target` is much more broad.

ZeroEkkusu (approved):
Thank you. @mds1, you can merge if LG to you.

mds1:
Thanks all!

PR #247, opened 2022-12-02
feat: improve StdChains

### LINKED ISSUE #232: `[rpc_endpoints]` vars should be lazily loaded

adhusson:
* Lazy-load `[rpc_endpoints]` entries as needed [1] (closes #232 for real)
* `rpcUrl` can be set as 1) custom, 2) config entry, 3) default value
* No empty chainAlias, no 0 chainId, clearer error messages
* Shorter via_ir compilation time (on my machine, 68s -> 42s with default runs)
* Remove extraneous keccak256 calls

[1] forge lazy-loads `[rpc_endpoints]`. This is useful when e.g. CI is only running tests that use the `TESTNET_URL` env var, but don't know about e.g. `MAINNET_URL` and `FLASHBOTS_URL`. forge-std iterating through all rpc urls breaks this.

### DIFF
--- a/src/Script.sol
+++ b/src/Script.sol
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.2 <0.9.0;
+pragma experimental ABIEncoderV2;
 
 import {CommonBase} from "./Common.sol";
 // forgefmt: disable-next-line
--- a/src/Script.sol
+++ b/src/Script.sol
@@ -1,6 +1,5 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.2 <0.9.0;
-pragma experimental ABIEncoderV2;
 
 import {CommonBase} from "./Common.sol";
 // forgefmt: disable-next-line
--- a/src/StdChains.sol
+++ b/src/StdChains.sol
@@ -5,6 +5,35 @@ pragma experimental ABIEncoderV2;
 
 import {VmSafe} from "./Vm.sol";
 
+/**
+ * StdChains provides information about EVM compatible chains that can be used in scripts/tests.
+ * For each chain, the chain's name, chain ID, and a default RPC URL are provided. Chains are
+ * identified by their alias, which is the same as the alias in the `[rpc_endpoints]` section of
+ * the `foundry.toml` file. For best UX, ensure the alias in the `foundry.toml` file match the
+ * alias used in this contract, which can be found as the first argument to the
+ * `setChainWithDefaultRpcUrl` call in the `initialize` function.
+ *
+ * There are two main ways to use this contract:
+ *   1. Set a chain with `setChain(string memory chainAlias, Chain memory chain)`
+ *   2. Get a chain with `getChain(string memory chainAlias)` or `getChain(uint256 chainId)`.
+ *
+ * The first time either of those are used, chains are initialized with the default set of RPC URLs.
+ * This is done in `initialize`, which uses `setChainWithDefaultRpcUrl`. Defaults are recorded in
+ * `defaultRpcUrls`.
+ *
+ * The `setChain` function is straightforward, and it simply saves off the given chain data.
+ *
+ * The `getChain` methods use `withRpcUrl` to return a chain. For example, let's say we want to
+ * retrieve `mainnet`'s RPC URL:
+ *   - If you haven't set any mainnet chain info with `setChain` and you haven't specified that
+ *     chain in `foundry.toml`, the default data and RPC URL will be returned.
+ *   - If you have set a mainnet RPC URL in `foundry.toml` it will return that, if valid (e.g. if
+ *     a URL is given or if an environment variable is given and that environment variable exists).
+ *     Otherwise, the default data is returned.
+ *   - If you specified data with `setChain` it will return that.
+ *
+ * Summarizing the above, the prioritization hierarchy is `setChain` -> `foundry.toml` -> defaults.
+ */
 abstract contract StdChains {
     VmSafe private constant vm = VmSafe(address(uint160(uint256(keccak256("hevm cheat code")))));
 
--- a/src/StdChains.sol
+++ b/src/StdChains.sol
@@ -3,7 +3,7 @@ pragma solidity >=0.6.2 <0.9.0;
 
 pragma experimental ABIEncoderV2;
 
-import "./Vm.sol";
+import {VmSafe} from "./Vm.sol";
 
 abstract contract StdChains {
     VmSafe private constant vm = VmSafe(address(uint160(uint256(keccak256("hevm cheat code")))));
--- a/src/StdChains.sol
+++ b/src/StdChains.sol
@@ -8,6 +8,11 @@ import "./Vm.sol";
 abstract contract StdChains {
     VmSafe private constant vm = VmSafe(address(uint160(uint256(keccak256("hevm cheat code")))));
 
+    /// @dev To hide constructor warnings across solc versions due to different constructor visibility requirements and
+    /// syntaxes, we put the constructor in a private method and assign an unused return value to a variable. This
+    /// forces the method to run during construction, but without declaring an explicit constructor.
+    uint256 private CONSTRUCTOR = _constructor();
+
     struct Chain {
         // The chain name.
         string name;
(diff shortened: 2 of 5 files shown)


ZeroEkkusu:
Haven't reviewed;

> don't know about e.g. MAINNET_URL and FLASHBOTS_URL

Sorry, could you explain to me or give me an example of the problem? I see the code is more complicated.

> Shorter via_ir compilation time (on my machine, 68s -> 42s with default runs)

I saw that it uses the `_constructor` again, so I tested it on Uniswap/permit2. It takes 200s vs the current 140s. Out checks also show longer `via-ir` compilation times (9s vs 3s).

adhusson:
Thanks for taking a look!

re: 1, suppose you have
```toml
[rpc_endpoints]
testnet="${TESTNET_URL}"
mainnet="${MAINNET_URL}"
```

With foundry only, if a consumer of the project only knows about `TESTNET_URL` or `MAINNET_URL` (a CI, or a client library calling scripts), that's fine. The entry with an undefined var will never be interpreted and never cause an error. With forge-std, `rpcUrls` is called whenever some chain info is accessed, which throws if at least one `[rpc_endpoints]` entry cannot be read. Basically the gains of lazy-loading are gone.

re: 2, I confirm that compiling permit2 with `_constructor` is slower. I've just brought back the `initialize()` function.

src/StdChains.sol
```diff
@@ -20,70 +22,131 @@ abstract contract StdChains {
...
     }
 
-    function setChain(string memory key, string memory name, uint256 chainId, string memory rpcUrl) internal virtual {
+    // set chain info, with priority to chainAlias' rpc url in foundry.toml
+    function setChainWithDefaultRpcUrl(string memory chainAlias, Chain memory chain) internal {
+        string memory rpcUrl = chain.rpcUrl;
+        defaultRpcUrls[chainAlias] = rpcUrl;
+        chain.rpcUrl = "";
+        setChain(chainAlias, chain);
+        chain.rpcUrl = rpcUrl; // restore argument
```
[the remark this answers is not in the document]
adhusson: `chain` is a memory reference. All modifications to it will be visible throughout the current EVM call frame. Not restoring the `rpcUrl` means that `setChainWithDefaultRpcUrl` sets its argument's `rpcUrl` property to the empty.

src/StdChains.sol
```diff
@@ -20,70 +22,131 @@ abstract contract StdChains {
...
+                        revert(add(32, err), mload(err))
+                    }
+                }
+            }
+        }
     }
 
-    function setChain(string memory key, string memory name, uint256 chainId, string memory rpcUrl) internal virtual {
+    // set chain info, with priority to chainAlias' rpc url in foundry.toml
+    function setChainWithDefaultRpcUrl(string memory chainAlias, Chain memory chain) internal {
```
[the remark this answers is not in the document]
adhusson: Same, I see myself using that in toolkit contracts that inherit from `StdChains`.

src/StdChains.sol
```diff
@@ -20,70 +22,131 @@ abstract contract StdChains {
...
 
-    bool private initialized;
-
-    // Maps from a chain's key (matching the alias in the `foundry.toml` file) to chain data.
-    mapping(string => Chain) private chains;
-    // Maps from a chain ID to that chain's key name
-    mapping(uint256 => string) private idToKey;
+    // chain alias -> chain data
+    mapping(string => Chain) internal chains;
+    mapping(string => string) internal defaultRpcUrls;
```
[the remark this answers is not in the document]
adhusson: People who extend `StdChains`/forge-std with their own tooling may want to customize how chains are handled.

src/StdChains.sol
```diff
@@ -20,70 +22,131 @@ abstract contract StdChains {
...
+            chain.chainId != 0,
+            string(abi.encodePacked("StdChains getChain(uint): Chain ID ", vm.toString(chainId), " not found."))
+        );
+
+        withRpcUrl(chainAlias, chain);
+    }
+
+    // lookup rpcUrl, in descending order of priority:
+    // current -> config (foundry.toml) -> default
+    function withRpcUrl(string memory chainAlias, Chain memory chain) private view {
```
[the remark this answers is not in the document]
adhusson: `chain` is a memory reference, so the argument passed is just a memory offset. No copying occurs.

ZeroEkkusu (changes requested):
I need to test this locally, but here's what I've noticed:

- I don't think the `memory chain` changes persist
- Reverts should says `uint256` instead of `uint`
- `setChain` doesn't call `initialize` so it wouldn't revert if you tried setting a Chain with chain ID 1 immediately
- Same as previous, you're able to set a Chain with chain ID 1, and if you then called `getChain("mainnet")` it would revert because a chain with chain ID 1 already exisits, which is confusing

src/StdChains.sol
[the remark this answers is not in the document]
ZeroEkkusu: Let's do `_*` then - e.g. `_chains`.

src/StdChains.sol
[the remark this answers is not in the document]
ZeroEkkusu: Same as above.

adhusson:
Thanks for the comments, I've applied the changes. For the `memory chain`, you can try by adding this test

```solidity
function testPersist() public {
  Chain memory chain = Chain("a",123,"URL");
  console.log("before",chain.rpcUrl);
  setChainWithDefaultRpcUrl("alias",chain);
  console.log("after",chain.rpcUrl);
  }
```
and then toggle that line and see the difference:
https://github.com/foundry-rs/forge-std/blob/d65123c8354ca60019442d9374a9b0acd8cfad43/src/StdChains.sol#L86

src/StdChains.sol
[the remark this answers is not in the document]
ZeroEkkusu: @adhusson - you didn't add an underscore before the name.

ZeroEkkusu:
Good to merge after the discussion is resolved.

src/Script.sol
```diff
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.2 <0.9.0;
+pragma experimental ABIEncoderV2;
```
mds1, line 3: Is the thinking here that we add this to `Script.sol` and `Test.sol` so users on <0.8 don't need to add it themselves? If that seems fine, and my only question is: ABI encoder v2 is the default in 0.8, so does explicitly specifying it here as experimental change anything? (i.e. like using an older/buggier version of the v2 encoder or something)

src/StdChains.sol
```diff
@@ -20,71 +22,126 @@ abstract contract StdChains {
         string rpcUrl;
     }
 
-    bool private initialized;
+    // chain alias -> chain data
+    mapping(string => Chain) internal _chains;
+    mapping(string => string) internal _defaultRpcUrls;
+    mapping(uint256 => string) internal _idToAlias;
```
mds1, line 28: Let's add the comments back to keep these a bit more self-documenting for future contributors
```suggestion
    // Maps from a chain ID to that chain's alias
    mapping(uint256 => string) internal _idToAlias;
```

src/StdChains.sol
```diff
@@ -20,71 +22,126 @@ abstract contract StdChains {
         string rpcUrl;
     }
 
-    bool private initialized;
+    // chain alias -> chain data
+    mapping(string => Chain) internal _chains;
+    mapping(string => string) internal _defaultRpcUrls;
```
mds1, line 27: Similarly
(matching the alias in the `foundry.toml` file) to chain data.
```suggestion
    // Maps from the chain's alias (matching the alias in the `foundry.toml` file) to chain data.
    mapping(string => Chain) internal _chains;
    // Maps from the chain's alias (matching the alias in the `foundry.toml` file) to it's default RPC URL
    mapping(string => string) internal _defaultRpcUrls;
```

test/StdChains.t.sol
```diff
@@ -17,25 +17,43 @@ contract StdChainsTest is Test {
 
     // Ensure we can connect to the default RPC URL for each chain.
     function testRpcs() public {
-        (string[2][] memory rpcs) = vm.rpcUrls();
-        for (uint256 i = 0; i < rpcs.length; i++) {
-            ( /* string memory name */ , string memory rpcUrl) = (rpcs[i][0], rpcs[i][1]);
-            vm.createSelectFork(rpcUrl);
-        }
+        string memory rpcUrl;
```
mds1, line 20: We now only check an arbitrary subset of default URLs, but we should still test all our default RPCs. Can conditionally skip any aliases that we know have invalid ones and will revert

src/StdChains.sol
```diff
@@ -20,71 +22,126 @@ abstract contract StdChains {
...
-
-        // Loop over RPC URLs in the config file to replace the default RPC URLs
-        Vm.Rpc[] memory rpcs = vm.rpcUrlStructs();
-        for (uint256 i = 0; i < rpcs.length; i++) {
-            chains[rpcs[i].key].rpcUrl = rpcs[i].url;
-        }
-
         initialized = true;
+
+        setChainWithDefaultRpcUrl("anvil", Chain("Anvil", 31337, "http://127.0.0.1:8545"));
```
mds1, line 122: @adhusson @ZeroEkkusu can we get some code comments in this PR, and maybe after some book updates, for StdChains / this PR? There's now a lot going on in `StdChains` and it's fairly complex, so the UX isn't the most intuitive by skimming the code (and there's a lot to skim). To start, a natspec block right above the `abstract contract StdChains {` line would be helpful. Some questions I'd like answered by those comments:
1. What's the default, out of the box behavior when I:
    1. Access a chain with a pre-existing alias, like "mainnet"
    2. Access a chain without a pre-existing alias, like "myNewChain"
    3.  For (1) and (2) how does it vary if nothing is in my foundry.toml vs. if I have a hardcoded RPC URL vs. if I have an env var, and that env var does or does not exist
2. How do I add a new chain?
3. What's the difference between `setChain` and `setChainWithDefaultRpcUrl` and when do I use each?
4. How do I rename an existing chain, like changing `anvil` to `hardhat`?

src/StdChains.sol
```diff
@@ -20,71 +22,126 @@ abstract contract StdChains {
...
+        string memory chainAlias = _idToAlias[chainId];
+
+        chain = _chains[chainAlias];
+
+        require(
+            chain.chainId != 0,
+            string(abi.encodePacked("StdChains getChain(uint256): Chain with ID ", vm.toString(chainId), " not found."))
+        );
+
+        withRpcUrl(chainAlias, chain);
```
mds1, line 56: Why do we call a view method at the end of this function? If that's intentional/required, def not intuitive and we should document what's going on here with some comments

mds1: Seems like you're modifying the `chain` memory var in `withRpcUrl` and then it bubbles up to here so that new RP URL is returned. But that's a bit strange, why not just return the new RPC URL and set it explicitly?

src/StdChains.sol
```diff
@@ -20,71 +22,126 @@ abstract contract StdChains {
...
+    }
+
+    // lookup rpcUrl, in descending order of priority:
+    // current -> config (foundry.toml) -> default
+    function withRpcUrl(string memory chainAlias, Chain memory chain) private view {
```
mds1, line 61: `withRpcUrl` feels like a bad name. maybe `findRpcUrl` or `getRpcUrl`, then return the RPC URL and set it in the parent method

src/Script.sol
ZeroEkkusu: No, I planned to remove this along with `Components`.

src/StdChains.sol
ZeroEkkusu: Do we need a more discouraging naming, so that users don’t shout themselves in the foot? E.g. `_id_to_alias_`.

This is only for devs.

src/StdChains.sol
ZeroEkkusu: Btw this function should also be only for devs and have the same discouraging naming style.

src/StdChains.sol
mds1: What about just making it `private` instead of a weird naming style?

ZeroEkkusu: https://github.com/foundry-rs/forge-std/pull/247#discussion_r1043412377 - def not the best solution - if @adhusson doesn’t intend to extend the features immediately, a better solution is to keep everything private, as was, and if someone requests it in the future, make changes then. What do you think?

mds1: Agreed, let's keep as much of the internal implementation `private` as possible to keep the UX/interface as simple/safe as possible, and we can iterate in the future when the need arises

ZeroEkkusu:
Pushed changes for `1.2.0` release.

After [implemet](https://github.com/foundry-rs/foundry/issues/3871) is resolved, we should be able to revert the changes without changing the API.

src/StdChains.sol
mds1: Done in 906e150

mds1:
Still a few open comments here to address before merging

src/StdChains.sol
ZeroEkkusu: There are 3 functions: 2 `getChain` and 1 `setChain`. The first time any one of these is called, chains are initialized with the default set of RPC URLs. This is done by `initialize`, which uses `setChainWithDefaultRpcUrl`. Defaults are recorded in `defaultRpcUrls`.

`setChain` is straightforward. It just sets a chain (in `chains`).

`getChain` uses `withRpcUrl` to return a chain. For example, let's say we want to retrieve `mainnet`. If you haven't set it with `setChain` and you haven't specified it in `foundry.toml`, the default one will be returned. If you set it in `foundry.toml` it will return that, if valid (e.g. the env exists), otherwise, the default one. If you specified it with `setChain` it will return that.

mds1: Thanks for the docs, I modified them slightly and ported them as comments in 4b9b197. We can pull that and reuse it for the book PR too

ZeroEkkusu (approved):

PR #250, opened 2022-12-04
Make "getRecordedLogs" and "accesses" view

paulrberg:
I think that these two cheatcodes should be of mutability `view`, based on the logic in the comments at the top of the file.

### DIFF
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -52,7 +52,7 @@ interface VmSafe {
     // Records all storage reads and writes
     function record() external;
     // Gets all accessed reads and write slot from a recording session, for a given address
-    function accesses(address) external returns (bytes32[] memory reads, bytes32[] memory writes);
+    function accesses(address) external view returns (bytes32[] memory reads, bytes32[] memory writes);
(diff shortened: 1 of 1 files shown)


adhusson:
As a data point I did not make them view because they do change internal state -- they enable calling their respective log/access record functions without causing an error. So strictly speaking, they do change the context! In practice it may be worth having them be view though.

mds1:
Let's leave these as mutable for the rationale @adhusson gave above + convo in https://github.com/foundry-rs/forge-std/pull/250. This rationale is explained in `Vm.sol` in the comment:

> 1. If you can influence a later call's observable behaviour, you're neither `view` nor `pure` (you are modifying some state be it the EVM, interpreter, filesystem, etc),

Calling e.g. `getRecordedLogs` clears the recorded logs, so calling it twice without a `recordLogs` in between results in different return data from the two calls, which is why it's not view

Def open to rephrasing for more clarity if needed

paulrberg:
In this case, renaming this cheatcode to something like `clearAndGetRecordedLogs` might be helpful.

Issue #129, opened 2022-07-15
Feature request: adding console.log(string,bytes32)


0xPhaze:
I often find myself needing to log bytes32 and currently this only works with console.logBytes32.
Is there a possibility of extending console to add in console.log(bytes32) or console.log(string,bytes32)?

paulrberg:
Related: https://github.com/NomicFoundation/hardhat/issues/1440

0xPhaze:
Maybe console2 could pick up these changes quicker than waiting for changes to console.
And couldn't I just be required to explicitly cast to bytes32 in the case of ambiguity. For example: console.log('var', bytes32(0x1337)). This would still be a nicer UX compared to console.logBytes32, imo, but either is good.

paulrberg:
I agree, not having to use `console.logBytes32` would be a better DX.

mds1:
I'd say PRs for things like this into `console2.sol` are good, but we should leave `console.sol` 1:1 with hardhat. @ZeroEkkusu let me know if that seems reasonable.

> And couldn't I just be required to explicitly cast to bytes32 in the case of ambiguity. For example: console.log('var', bytes32(0x1337)). This would still be a nicer UX compared to console.logBytes32, imo, but either is good.

Yea if there's ambiguity in your call solc will just force you to cast and be explicit, which I think is ok

ZeroEkkusu:
Agree with `console`/`console2`.

mds1:
> Is there a possibility of extending console to add in console.log(bytes32) or console.log(string,bytes32)?

As hinted at above, the reason it's `console.logBytes32(bytes32)` and why we shouldn't add support for `console.log(bytes32)` (and similar for the overload with a leading string arg) is because then something like `console.log("x")` becomes ambiguous, as solc doesn't know if `"x"` is bytes32 or a string. So this would end up being a big breaking change for users by requiring type casts to existing code. As a result I'm going to close this 🙂

If you'd prefer the overload + required casting, I'd suggest leaving a comment on https://github.com/foundry-rs/foundry/issues/3782 since there may eventually be a `vm.consoleLog` type of cheat

PR #253, opened 2022-12-09
feat: implement boundInt

JhChoy:
Implementing `boundInt(int256,int256,int256)` at StdUtils.sol to help fuzzing test with integer arguments.

One problem is that `log(string,int256)` does not exist so I used `log(string,uint256)` instead of it.

### DIFF
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -41,8 +41,8 @@ abstract contract StdUtils {
         console2.log("Bound Result", result);
     }
 
-    function boundInt(int256 x, int256 min, int256 max) internal view virtual returns (int256 result) {
-        require(min <= max, "StdUtils boundInt(int256,int256,int256): Max is less than min.");
+    function bound(int256 x, int256 min, int256 max) internal view virtual returns (int256 result) {
+        require(min <= max, "StdUtils bound(int256,int256,int256): Max is less than min.");
 
         // Shifting all int256 values to uint256 to use _bound function. The range of two types are:
         // int256 : -(2**255) ~ (2**255 - 1)
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -58,7 +58,7 @@ abstract contract StdUtils {
         uint256 y = _bound(_x, _min, _max);
 
         // To move it back to int256 value, subtract INT256_MIN_ABS at here.
-        result = y < INT256_MIN_ABS ? int256(~(INT256_MIN_ABS - y) + 1): int256(y - INT256_MIN_ABS);
+        result = y < INT256_MIN_ABS ? int256(~(INT256_MIN_ABS - y) + 1) : int256(y - INT256_MIN_ABS);
         console2.log("Bound Result", result);
     }
 
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -4,12 +4,12 @@ pragma solidity >=0.6.2 <0.9.0;
 import "./console2.sol";
 
 abstract contract StdUtils {
+    uint256 private constant INT256_MIN_ABS =
+        57896044618658097711785492504343953926634992332820282019728792003956564819968;
     uint256 private constant UINT256_MAX =
         115792089237316195423570985008687907853269984665640564039457584007913129639935;
 
     function _bound(uint256 x, uint256 min, uint256 max) internal pure virtual returns (uint256 result) {
-        require(min <= max, "StdUtils bound(uint256,uint256,uint256): Max is less than min.");
-
         // If x is between min and max, return x directly. This is to ensure that dictionary values
         // do not get shifted if the min is nonzero. More info: https://github.com/foundry-rs/forge-std/issues/188
         if (x >= min && x <= max) return x;
@@ -36,8 +36,22 @@ abstract contract StdUtils {
     }
 
     function bound(uint256 x, uint256 min, uint256 max) internal view virtual returns (uint256 result) {
+        require(min <= max, "StdUtils bound(uint256,uint256,uint256): Max is less than min.");
         result = _bound(x, min, max);
         console2.log("Bound Result", result);
+    }
+
+    function boundInt(int256 x, int256 min, int256 max) internal view virtual returns (int256 result) {
+        require(min <= max, "StdUtils boundInt(int256,int256,int256): Max is less than min.");
+
+        uint256 _x = x < 0 ? (INT256_MIN_ABS - ~uint256(x) - 1) : (uint256(x) + INT256_MIN_ABS);
+        uint256 _min = min < 0 ? (INT256_MIN_ABS - ~uint256(min) - 1) : (uint256(min) + INT256_MIN_ABS);
+        uint256 _max = max < 0 ? (INT256_MIN_ABS - ~uint256(max) - 1) : (uint256(max) + INT256_MIN_ABS);
+
+        uint256 y = _bound(_x, _min, _max);
+        result = y > INT256_MIN_ABS ? int256(y - INT256_MIN_ABS) : -int256(INT256_MIN_ABS - y);
+        // TODO: change after log(string,int256) is made
+        console2.log("Bound Result", uint256(result));
     }
 
     /// @dev Compute the address a contract will be deployed at for a given deployer address and nonce
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -1,9 +1,13 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.2 <0.9.0;
 
-import "./console2.sol";
+// TODO Remove import.
+import {Vm} from "./Vm.sol";
 
 abstract contract StdUtils {
+    Vm private constant vm = Vm(address(uint160(uint256(keccak256("hevm cheat code")))));
+    address private constant CONSOLE2_ADDRESS = 0x000000000000000000636F6e736F6c652e6c6f67;
+
     uint256 private constant INT256_MIN_ABS =
         57896044618658097711785492504343953926634992332820282019728792003956564819968;
     uint256 private constant UINT256_MAX =
@@ -38,7 +42,7 @@ abstract contract StdUtils {
 
     function bound(uint256 x, uint256 min, uint256 max) internal view virtual returns (uint256 result) {
         result = _bound(x, min, max);
-        console2.log("Bound Result", result);
+        console2_log("Bound Result", result);
     }
 
     function bound(int256 x, int256 min, int256 max) internal view virtual returns (int256 result) {
@@ -59,7 +63,7 @@ abstract contract StdUtils {
 
         // To move it back to int256 value, subtract INT256_MIN_ABS at here.
         result = y < INT256_MIN_ABS ? int256(~(INT256_MIN_ABS - y) + 1) : int256(y - INT256_MIN_ABS);
-        console2.log("Bound Result", result);
+        console2_log(string(abi.encodePacked("Bound result: ", vm.toString(result))));
     }
 
     /// @dev Compute the address a contract will be deployed at for a given deployer address and nonce
@@ -103,5 +107,15 @@ abstract contract StdUtils {
 
     function addressFromLast20Bytes(bytes32 bytesValue) private pure returns (address) {
         return address(uint160(uint256(bytesValue)));
+    }
+
+    function console2_log(string memory p0, uint256 p1) private view {
+        (bool status,) = address(CONSOLE2_ADDRESS).staticcall(abi.encodeWithSignature("log(string,uint256)", p0, p1));
+        status;
+    }
+
+    function console2_log(string memory p0) private view {
+        (bool status,) = address(CONSOLE2_ADDRESS).staticcall(abi.encodeWithSignature("log(string)", p0));
+        status;
     }
 }
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -63,7 +63,7 @@ abstract contract StdUtils {
 
         // To move it back to int256 value, subtract INT256_MIN_ABS at here.
         result = y < INT256_MIN_ABS ? int256(~(INT256_MIN_ABS - y) + 1) : int256(y - INT256_MIN_ABS);
-        console2_log(string(abi.encodePacked("Bound result: ", vm.toString(result))));
+        console2_log("Bound result", vm.toString(result));
     }
 
     /// @dev Compute the address a contract will be deployed at for a given deployer address and nonce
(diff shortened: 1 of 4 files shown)


test/StdUtils.t.sol
```diff
@@ -74,6 +74,76 @@ contract StdUtilsTest is Test {
...
+
+    function testBoundInt_EdgeCoverage() public {
+        assertEq(boundInt(type(int256).min, -50, 150), -50);
+        assertEq(boundInt(type(int256).min + 1, -50, 150), -49);
+        assertEq(boundInt(type(int256).min + 2, -50, 150), -48);
+        assertEq(boundInt(type(int256).min + 3, -50, 150), -47);
+        assertEq(boundInt(type(int256).max, -50, 150), 150);
+        assertEq(boundInt(type(int256).max - 1, -50, 150), 149);
+        assertEq(boundInt(type(int256).max - 2, -50, 150), 148);
+        assertEq(boundInt(type(int256).max - 3, -50, 150), 147);
```
mds1, line 101: Can we also tests for the inverse cases, e.g.
```solidity
assertEq(boundInt(type(int256).min, 10, 150), 10);
assertEq(boundInt(type(int256).min + 1, 10, 150), 11);
// add the other two
assertEq(boundInt(type(int256).max, -50, -10), -10);
assertEq(boundInt(type(int256).max - 1, -50, -10), -11);
// add the other two
```

Maybe good to explicitly test when zero is both the lower and upper bound in those scenarios also. Ints have some more edge cases than Uints IME so just want to make sure it's well tested 🙂

src/StdUtils.sol
```diff
@@ -4,12 +4,12 @@ pragma solidity >=0.6.2 <0.9.0;
 import "./console2.sol";
 
 abstract contract StdUtils {
+    uint256 private constant INT256_MIN_ABS =
```
mds1, line 7: _(arbitrary comment location to start a thread)_ 

> One problem is that `log(string,int256)` does not exist so I used `log(string,uint256)` instead of it.

If you want to add support for this:
1. Add the required method to `console2.sol` in forge-std
2. Update the [ABI file](https://github.com/foundry-rs/foundry/blob/master/evm/abi/console.json) in foundry

I think that should be all that's needed, but I may be missing something. Also related is https://github.com/foundry-rs/forge-std/issues/129 if you wanted to add that overload at the same time

src/StdUtils.sol
```diff
@@ -36,10 +36,24 @@ abstract contract StdUtils {
...
         result = _bound(x, min, max);
         console2.log("Bound Result", result);
     }
 
+    function boundInt(int256 x, int256 min, int256 max) internal view virtual returns (int256 result) {
+        require(min <= max, "StdUtils boundInt(int256,int256,int256): Max is less than min.");
+
+        uint256 _x = x < 0 ? (INT256_MIN_ABS - ~uint256(x) - 1) : (uint256(x) + INT256_MIN_ABS);
+        uint256 _min = min < 0 ? (INT256_MIN_ABS - ~uint256(min) - 1) : (uint256(min) + INT256_MIN_ABS);
+        uint256 _max = max < 0 ? (INT256_MIN_ABS - ~uint256(max) - 1) : (uint256(max) + INT256_MIN_ABS);
```
mds1, line 49: Mind just adding a comment explaining how this works / linking to a reference? Want to make sure it's clear/understandable to maintainers (including myself) 😅

I think what you're doing is shifting all values from the int domain to the uint domain, running `_bound`, then shifting them back? Which makes sense and is how I'd do it, just not intuitive how the shift works

src/StdUtils.sol
```diff
@@ -36,10 +36,24 @@ abstract contract StdUtils {
     }
 
     function bound(uint256 x, uint256 min, uint256 max) internal view virtual returns (uint256 result) {
+        require(min <= max, "StdUtils bound(uint256,uint256,uint256): Max is less than min.");
         result = _bound(x, min, max);
         console2.log("Bound Result", result);
     }
 
+    function boundInt(int256 x, int256 min, int256 max) internal view virtual returns (int256 result) {
```
mds1, line 44: Let's just call this `bound`
```suggestion
    function bound(int256 x, int256 min, int256 max) internal view virtual returns (int256 result) {
```

src/StdUtils.sol
```diff
@@ -4,12 +4,12 @@ pragma solidity >=0.6.2 <0.9.0;
 import "./console2.sol";
 
 abstract contract StdUtils {
+    uint256 private constant INT256_MIN_ABS =
+        57896044618658097711785492504343953926634992332820282019728792003956564819968;
     uint256 private constant UINT256_MAX =
         115792089237316195423570985008687907853269984665640564039457584007913129639935;
 
     function _bound(uint256 x, uint256 min, uint256 max) internal pure virtual returns (uint256 result) {
-        require(min <= max, "StdUtils bound(uint256,uint256,uint256): Max is less than min.");
```
mds1, line 11: One side effect here: by removing this `require` statement, it's no longer safe to use `_bound` as an alternative to `bound` to get the same behavior without the log noise, so this is technically a breaking change that we should avoid IMO. I think we can keep this here, and just leave the new require statement in the `boundInt` method. You get a duplicate check as a result, but that's ok

src/StdUtils.sol
```diff
@@ -36,10 +36,24 @@ abstract contract StdUtils {
...
 
+    function boundInt(int256 x, int256 min, int256 max) internal view virtual returns (int256 result) {
+        require(min <= max, "StdUtils boundInt(int256,int256,int256): Max is less than min.");
+
+        uint256 _x = x < 0 ? (INT256_MIN_ABS - ~uint256(x) - 1) : (uint256(x) + INT256_MIN_ABS);
+        uint256 _min = min < 0 ? (INT256_MIN_ABS - ~uint256(min) - 1) : (uint256(min) + INT256_MIN_ABS);
+        uint256 _max = max < 0 ? (INT256_MIN_ABS - ~uint256(max) - 1) : (uint256(max) + INT256_MIN_ABS);
+
+        uint256 y = _bound(_x, _min, _max);
+        result = y > INT256_MIN_ABS ? int256(y - INT256_MIN_ABS) : -int256(INT256_MIN_ABS - y);
```
mds1, line 52: This shift back is more intuitive, but maybe worth a comment here anyway

mds1 (commented):
This is awesome, thanks a lot! Pinging @paulrberg since I know he was interesting in this method.

Just a few nits for you 🙂

test/StdUtils.t.sol
JhChoy: Great! I'll add them.

src/StdUtils.sol
JhChoy: Okay, I'll check this out.

src/StdUtils.sol
JhChoy: Agree, I'll add some comments about this code.

src/StdUtils.sol
JhChoy: I wanted to... But Solidity says it is duplicated with `bound(uint256,uint256,uint256)` 😭.

src/StdUtils.sol
JhChoy: That's a good point. I'll fix it.

src/StdUtils.sol
mds1: We can call it `bound`, because it has a different function signature than the other `bound`, so that's an allowed overload.

But if you change the name to `bound` with no other changes, solc will error because certain statements like `assertEq(bound(5, 0, 4), 0), 0)` are now ambiguous—are those inputs ints or uints? It can't tell which `bound` implementation to use.

So we can resolve that by just casting the first input value so it knows which overload to use: `assertEq(bound(uint256(5), 0, 4), 0)`

This is a bit clunky for tests where we hardcode constants and need to cast everywhere now, but I think that's ok. In standard fuzz tests, solidity will know the type of that first input to bound because it a function input of uint256 or int256 in the fuzz test, so the casting won't be needed and UX will be better for users

src/StdUtils.sol
```diff
@@ -4,12 +4,12 @@ pragma solidity >=0.6.2 <0.9.0;
 import "./console2.sol";
 
 abstract contract StdUtils {
+    uint256 private constant INT256_MIN_ABS =
+        57896044618658097711785492504343953926634992332820282019728792003956564819968;
```
mds1, line 8: _(arbitrary comment location to start a thread)_

Looks like there was also a test failure in CI:

```
[FAIL. Reason: Arithmetic over/underflow Counterexample: calldata=0x87b338ab000000000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000000, args=[0, -57896044618658097711785492504343953926634992332820282019728792003956564819967, -57896044618658097711785492504343953926634992332820282019728792003956564819968]] testBoundInt(int256,int256,int256) (runs: 630, μ: 7029, ~: 7054)
Traces:
  [2013] StdUtilsTest::testBoundInt(0, -57896044618658097711785492504343953926634992332820282019728792003956564819967, -57896044618658097711785492504343953926634992332820282019728792003956564819968) 
    └─ ← "Arithmetic over/underflow"
```

src/StdUtils.sol
JhChoy: [328f4df](https://github.com/foundry-rs/forge-std/pull/253/commits/328f4dfe531b66876ccafa8be1184b923b04c02d)

src/StdUtils.sol
JhChoy: [328f4df](https://github.com/foundry-rs/forge-std/pull/253/commits/328f4dfe531b66876ccafa8be1184b923b04c02d)

src/StdUtils.sol
JhChoy: [c43e95b](https://github.com/foundry-rs/forge-std/pull/253/commits/c43e95be0f0771d0d188cd1b855ad0fe972b5b5c)

test/StdUtils.t.sol
JhChoy: [c43e95b](https://github.com/foundry-rs/forge-std/pull/253/commits/c43e95be0f0771d0d188cd1b855ad0fe972b5b5c)

src/StdUtils.sol
JhChoy: [174b694](https://github.com/foundry-rs/forge-std/pull/253/commits/174b694c8b60d2670d5e56f6d70ca554c80cc37b)

src/StdUtils.sol
JhChoy: Oh, I understand. Now it works well.
Applied at here. [25ff698](https://github.com/foundry-rs/forge-std/pull/253/commits/25ff69883e47be0c10b0e10fdbf7e391efc97acd)

test/StdUtils.t.sol
mds1: Small nit, do you mind reordering the tests like this? It makes it easier to read since sequential/related tests are next to each other. (I'm happy to push the commit, but don't want to push yet in case you're working on https://github.com/foundry-rs/forge-std/pull/253#discussion_r1045104818).

src/StdUtils.sol
JhChoy: What will be the better one, `log(string,int256)` or `logInt(string,int256)`?
At the [related thread](https://github.com/NomicFoundation/hardhat/issues/1440) of #129, they decided to name `logInt(int256...)` rather than `log(int256...)` because of the ambiguity.

test/StdUtils.t.sol
JhChoy: Got it, I'll apply it.

src/StdUtils.sol
mds1: I'm ok with `log(string,int256)` because in practice you'll be doing `console.log(myVar)` (unambiguous since `myVar` has a type) instead of `console.log(3)` (ambiguous because we don't know 3's type)

test/StdUtils.t.sol
JhChoy: Fixed at [here](https://github.com/foundry-rs/forge-std/pull/253/commits/3e50c5d244cdfa055aa6468334ad6193e746d7fe).

src/StdUtils.sol
JhChoy: Applied at [here](https://github.com/foundry-rs/forge-std/pull/253/commits/3e50c5d244cdfa055aa6468334ad6193e746d7fe)!

paulrberg:
I don't currently have bandwidth to review but this is good stuff, thanks for implementing a signed bound, @JhChoy! I needed this for my [PRBMath](prb-math) tests.

src/StdUtils.sol
mds1: Thanks! Note that the logs won't actually be shown until we upstream the corresponding change to forge. I can do that soon though

mds1 (approved):
LGTM, thanks @JhChoy! 

@ZeroEkkusu mind taking a look before we merge?

mds1:
@JhChoy just a heads up I pushed two small commits to this branch and opened https://github.com/foundry-rs/foundry/issues/3884

JhChoy:
@mds1 Thanks for the review! 🙌

ZeroEkkusu (approved):
Thanks.

ZeroEkkusu:
Pushed changes for `1.2.0` release.

After https://github.com/foundry-rs/foundry/issues/3884 is resolved, we will remove the workaround.

Note: `console2_log` functions are part of the upcoming "remove Components" PR, so I made those changes here to avoid conflicts.

src/StdUtils.sol
```diff
@@ -37,7 +42,28 @@ abstract contract StdUtils {
...
+        // So, use `~uint256(x) + 1` instead.
+        uint256 _x = x < 0 ? (INT256_MIN_ABS - ~uint256(x) - 1) : (uint256(x) + INT256_MIN_ABS);
+        uint256 _min = min < 0 ? (INT256_MIN_ABS - ~uint256(min) - 1) : (uint256(min) + INT256_MIN_ABS);
+        uint256 _max = max < 0 ? (INT256_MIN_ABS - ~uint256(max) - 1) : (uint256(max) + INT256_MIN_ABS);
+
+        uint256 y = _bound(_x, _min, _max);
+
+        // To move it back to int256 value, subtract INT256_MIN_ABS at here.
+        result = y < INT256_MIN_ABS ? int256(~(INT256_MIN_ABS - y) + 1) : int256(y - INT256_MIN_ABS);
+        console2_log(string(abi.encodePacked("Bound result: ", vm.toString(result))));
```
mds1, line 66: Instead of using `abi.encodePacked`, can just use the `log(string,string)` version. Also note the removal of the `:` for consistency since it wasn't there before
```suggestion
        console2_log("Bound result", vm.toString(result));
```

mds1: @ZeroEkkusu feel free to merge after this change

ZeroEkkusu:
Thanks, @JhChoy - good work.

PR #255, opened 2022-12-14
fix: remove `Components.sol`

### LINKED ISSUE #239: Remove `Components.sol`
Breaking change.

This will have minimal impact on composability.

## Problem

`Components.sol`

```
import A
import B
import C
```

`Test.sol`

```
import A, B from Components
```

The selective import causes `C` to compile as well. This will increase the compilation time as components pile up.

## Solution

Remove `Components.sol`. Require importing directly from files (most users do this anyway).

Let's include this in `1.2.0`.

ZeroEkkusu:
## Motivation

Resolve #239 

## Solution

 - This PR removes `Components.sol`
 - #253 removes `console2` usage in `StdUtils.sol`
 - Move some constants from `CommonBase` to `TestBase` and `ScriptBase`, depending on usage


src/Test.sol
```diff
@@ -2,10 +2,25 @@
...
+import {StdUtils} from "./StdUtils.sol";
+import {Vm} from "./Vm.sol";
 
-abstract contract TestBase is CommonBase {}
+abstract contract TestBase is CommonBase {
+    // Default address for tx.origin and msg.sender, 0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38.
+    address internal constant DEFAULT_SENDER = address(uint160(uint256(keccak256("foundry default caller"))));
+
+    // Address of the test contract, deployed by the DEFAULT_SENDER.
+    address internal constant DEFAULT_TEST_CONTRACT = 0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f;
```
mds1, line 23: I'd say just keep these in `CommonBase` as they are constants of forge, and afaik apply to scripts too,. So it's easier to track and find them all when they're in a single spot, and you typically want more than one (e.g. when `vm.assume`ing to exclude all forge addresses)

mds1 (approved):
Thanks!

ZeroEkkusu:
Cc @paulrberg, I saw you use it in [`PRBMath`](https://github.com/paulrberg/prb-math/blob/2c597391f16e5ada196aca3d77c81fd653574974/test/BaseTest.t.sol#L4).

Here's the [reason](https://github.com/foundry-rs/forge-std/issues/239) why we removed it.

paulrberg:
Thanks for the heads-up @ZeroEkkusu, and good catch. This is definitely counter-intuitive and odd. The Solidity team should know about this.

PR #256, opened 2022-12-16
fix: move cheats, improve docs

ZeroEkkusu:
## Motivation

- Make some std-cheats `Safe`
- Encourage users to customize Test and Script DX
- Move the `Base`s for the same reason as #255 

## Solution

- Move `isFork`, `skipWhenForking`, `skipWhenNotForking` to `StdCheatsSafe`
- Improve `Test.sol` and `Script.sol`
- Rename `Common.sol` to `Base.sol`
- Move `TestBase` and `ScriptBase` to `Base.sol`
- Misc changes

### DIFF
--- a/src/Script.sol
+++ b/src/Script.sol
@@ -1,6 +1,19 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.2 <0.9.0;
 
+// 💬 ABOUT
+/**
+ *  Standard Library's default Script DX.
+ *
+ *  You can customize it to suit the needs of your project by creating a similar file:
+ *
+ *      - import the standard modules you need
+ *      - customize the standard modules by overriding and extending functionality
+ *      - make your own modules
+ *      - install community modules
+ */
+
+// 🧩 MODULES
 import {CommonBase} from "./Common.sol";
 import {console} from "./console.sol";
 import {console2} from "./console2.sol";
@@ -12,13 +25,16 @@ import {StdStorage, stdStorageSafe} from "./StdStorage.sol";
 import {StdUtils} from "./StdUtils.sol";
 import {VmSafe} from "./Vm.sol";
 
+// 📦 BOILERPLATE
 abstract contract ScriptBase is CommonBase {
-    // Create2 factory used by scripts when deploying with create2, https://github.com/Arachnid/deterministic-deployment-proxy.
+    // Used when deploying with create2, https://github.com/Arachnid/deterministic-deployment-proxy.
     address internal constant CREATE2_FACTORY = 0x4e59b44847b379578588920cA78FbF26c0B4956C;
 
     VmSafe internal constant vmSafe = VmSafe(VM_ADDRESS);
 }
 
+// ⚡️ SCRIPT DX
 abstract contract Script is StdChains, StdCheatsSafe, StdUtils, ScriptBase {
+    // Note: IS_SCRIPT() must return true.
     bool public IS_SCRIPT = true;
 }
--- a/src/Script.sol
+++ b/src/Script.sol
@@ -14,7 +14,7 @@ pragma solidity >=0.6.2 <0.9.0;
  */
 
 // 🧩 MODULES
-import {CommonBase} from "./Common.sol";
+import {ScriptBase} from "./Base.sol";
 import {console} from "./console.sol";
 import {console2} from "./console2.sol";
 import {StdChains} from "./StdChains.sol";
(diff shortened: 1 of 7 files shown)


src/Script.sol
```diff
@@ -12,13 +25,8 @@ import {StdStorage, stdStorageSafe} from "./StdStorage.sol";
...
 import {VmSafe} from "./Vm.sol";
 
-abstract contract ScriptBase is CommonBase {
-    // Create2 factory used by scripts when deploying with create2, https://github.com/Arachnid/deterministic-deployment-proxy.
-    address internal constant CREATE2_FACTORY = 0x4e59b44847b379578588920cA78FbF26c0B4956C;
-
-    VmSafe internal constant vmSafe = VmSafe(VM_ADDRESS);
-}
-
+// ✨ SCRIPT DX
```
mds1, line 28: Nit, "Script Developer Experience" doesn't make much sense as a contract header comment 😅
```suggestion
// ✨ SCRIPT CONTRACT
```

src/Test.sol
```diff
@@ -15,6 +28,8 @@ import {StdStorage, stdStorage} from "./StdStorage.sol";
 import {StdUtils} from "./StdUtils.sol";
 import {Vm} from "./Vm.sol";
 
-abstract contract TestBase is CommonBase {}
-
-abstract contract Test is DSTest, StdAssertions, StdChains, StdCheats, StdUtils, TestBase {}
+// ✨ TEST DX
```
mds1, line 31:
```suggestion
// ✨ TEST CONTRACT
```

src/Vm.sol
```diff
@@ -382,4 +378,8 @@ interface Vm is VmSafe {
     function transact(bytes32 txHash) external;
     // Fetches the given transaction from the given fork and executes it on the current state
     function transact(uint256 forkId, bytes32 txHash) external;
+    // Pauses gas metering (i.e. gas usage is not counted). Noop if already paused.
+    function pauseGasMetering() external;
+    // Resumes gas metering (i.e. gas usage is counted again). Noop if already on.
+    function resumeGasMetering() external;
```
mds1, line 384: I think you currently have `noGasMetering` modifier in `StdCheatsSafe` but this in `Vm` and not `VmSafe`?

ZeroEkkusu: Correct.

mds1: That seems contradictory since the modifier uses the cheats—how can the modifier be safe, but not the cheats? I'd say put them both in safe, or both unsafe (I lean towards safe, but can see the arg for unsafe so am indifferent)

ZeroEkkusu: We're already doing it - for example, `stdStorageSafe` uses `Vm` to modify state, but it is `Safe` because we are ensuring that it is safe.

src/Script.sol
ZeroEkkusu: Okay, but these are not contract headers, they are there to form a template of what a `Script.sol` file should look like.

ZeroEkkusu: Any other ideas?

src/Vm.sol
mds1: In that case we're using unsafe cheat codes in a safe way. In this case the modifier just uses the cheats in the same way you'd use them directly

ZeroEkkusu: I'll move them to `VmSafe`.

src/Script.sol
mds1: > Okay, but these are not contract headers, they are there to form a template of what a Script.sol file should look like.

Maybe I'm misunderstanding what you're trying to convey here. The intent of `Script.sol` is for people to inherit and use in their scripts. I don't think we have a good reason to encourage people to write their own `Script.sol` files and use them instead of this one.

It's great that these contracts modular and people _can_ do that, but these Script and Test contracts can be used directly and that's what forge-std is primarily intended to enable. This contract's docs should be explaining how to use this file in your scripts (and same for test). Docs on how to build your own `Script.sol` file modularly should live elsewhere (the README, the book, etc)

ZeroEkkusu: For any sufficiently complex project, you have to customize `Test`.

 > these Script and Test contracts can be used directly and that's what forge-std is primarily intended to enable

Yes. That's why `Test.sol` and `Script.sol` are there - to provide the default testing/scripting setup. But, as I intentionally made the files looks like, there's nothing special in them, and anyone can make their own if they need to. If they don't, they are not expected to. I can communicate it better in the docs.

ZeroEkkusu: I think the wording is already clear?

`You can customize it to suit the needs of your project by creating a similar file`
Or `If required, you can...`

Re docs: They can be improved, but I think they should stay in the file. Let me know if you disagree.

mds1: > For any sufficiently complex project, you have to customize `Test`.

What do you mean by customizing `Test`? Even for complex projects I just inherit from `Test` then add any additional helpers I need, but I've never needed to modify the `Test` base contract itself

ZeroEkkusu: What if you want to replace `DSTest` or `StdAssertions`?

mds1: > You can customize it to suit the needs of your project by creating a similar file

I don't think we should suggest this IMO. I find that sentence confusing without more docs/context/rationale (i.e. "why would I want to do that? what am I missing?"), and I can't think of a good reason for doing that vs. just inheriting from `Script` or `Test` and adding methods you need.

forge-std is written modularly so people can do whatever they want, but forge-std shouldn't encourage breaking it apart unless we have a good reason to encourage that, and I don't think we do (which is good! because we want it to be dead-simple, drop-in usage—if you have to customize it, something is wrong with forge-std)

If there are good arguments for replacing DSTest or StdAssertions, those changes should be upstreamed to forge-std, but so far I don't know of any good arguments for that 🙂

mds1 (approved):
LGTM, thanks!

PR #258, opened 2022-12-19
refactor: update total supply by default in "deal"

paulrberg:
I find it a bit odd that by default `deal` (the overloaded method without an `adjust` argument) does not update the total supply after minting the tokens. The tokens are technically printed out of thin air, so IMO the total supply should by default be updated accordingly, just in case end users rely upon the total supply in their tests.

https://github.com/foundry-rs/forge-std/blob/eb980e1d4f0e8173ec27da77297ae411840c8ccb/src/StdCheats.sol#L540

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -537,9 +537,9 @@ abstract contract StdCheats is StdCheatsSafe {
     }
 
     // Set the balance of an account for any ERC20 token
-    // Use the alternative signature to update `totalSupply`
+    // Use the alternative signature to not update `totalSupply`
     function deal(address token, address to, uint256 give) internal virtual {
-        deal(token, to, give, false);
+        deal(token, to, give, true);
     }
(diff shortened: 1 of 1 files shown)


mds1:
You can imagine the opposite argument too though, since you typically don't need to worry about the `sum(balance) == totalSupply` invariant during tests and might want to keep the original total supply. For example, if testing withdraws in an ERC4626 vault you may just deal shares to someone but you want the total supply of shares/exchange rate to remain unchanged

You can see the conversation here where it was decided to leave `totalSupply` unchanged by default: https://github.com/foundry-rs/forge-std/pull/17#issuecomment-1057827152

paulrberg:
Fine, @ZeroEkkusu made a good point in that other thread (about the internal logic of `totalSupply`). I don't have a strong opinion on this so I will just close the PR now.

PR #259, opened 2022-12-23
feat(stdstorage): Read and write dynamic length elements (string + bytes)

brockelmore:
Adds support for reading & writing strings and bytes. Works by checking read slots to see if it is a tightly packed dynamic element or spread out dynamic and comparing that to the return data from the call.

Allows writing to these kinds of elements as well by performing the expected packing and zeroing out any old slots. 

Fix https://github.com/foundry-rs/foundry/issues/3869


mds1:
Closing as stale

PR #261, opened 2023-01-01
feat: Add `InvariantTest` contract

lucas-manuel:
Add functions for:
- Excluding target contracts `excludeContract`
- Excluding sender addresses `excludeSender`
- Adding target contracts `targetContract`
- Adding target senders `targetSender`
- Adding target selectors `targetSelector`
- View functions for all arrays

### DIFF
--- a/src/InvariantTest.sol
+++ b/src/InvariantTest.sol
@@ -1,0 +1,65 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.6.2 <0.9.0;
+
+import {Test} from "./Test.sol";
+
+contract InvariantTest is Test {
+
+    struct FuzzSelector {
+        address addr;
+        bytes4[] selectors;
+    }
+
+    address[] private _excludedContracts;
+    address[] private _excludedSenders;
+    address[] private _targetedContracts;
+    address[] private _targetedSenders;
+
+    FuzzSelector[] internal _targetedSelectors;
+
+    function excludeContract(address newExcludedContract_) internal {
+        _excludedContracts.push(newExcludedContract_);
+    }
+
+    function excludeContracts() public view returns (address[] memory excludedContracts_) {
+        require(_excludedContracts.length != uint256(0), "NO_EXCLUDED_CONTRACTS");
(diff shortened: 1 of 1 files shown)


gakonst (approved):
LGTM - taken from https://github.com/maple-labs/contract-test-utils/blob/main/contracts/test.sol#L155

src/InvariantTest.sol
```diff
@@ -0,0 +1,65 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.6.2 <0.9.0;
+
+import {Test} from "./Test.sol";
+
+contract InvariantTest is Test {
```
mds1, line 6: I don't think we should inherit from `Test` here

It's common to have something like `ProtocolTest is Test` where `ProtocolTest` deploys contracts and scaffolds the system, and you'd want to be able to reuse that in your invariant tests.

If `InvariantTest` is standalone you can do `MyInvariantTests is ProtocolTest, InvariantTest` (with `MyInvariantTests is Test, InvariantTest` being the default way to setup invariant tests), but when `InvariantTest` inherits from `Test` you'd be inheriting from `Test` twice which is not allowed, so you can't easily reuse `ProtocolTest` with this

src/InvariantTest.sol
```diff
@@ -0,0 +1,65 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.6.2 <0.9.0;
```
mds1, line 2: CI is failing because this needs `forge fmt`. Also need the ABIEncoderV2 pragma  so this builds on older solc versions 
```suggestion
pragma solidity >=0.6.2 <0.9.0;
pragma experimental ABIEncoderV2;
```

joshieDo:
There's still a couple missing (https://github.com/foundry-rs/foundry/pull/2635)

```solidity
function excludeArtifacts() returns (string[] memory) // ["Contract1","src/Contract2.sol:Contract2"]
function targetArtifactSelectors() returns (string, bytes4[])[] memory) // [("Contract1", [Contract1.hello.selector])]
```

PR #263, opened 2023-01-03
feat: Update `InvariantTest` inheritance, add new functions, fix CI [WIP]

lucas-manuel:
Incorporating feedback from https://github.com/foundry-rs/forge-std/pull/261

### DIFF
--- a/src/InvariantTest.sol
+++ b/src/InvariantTest.sol
@@ -4,7 +4,6 @@ pragma solidity >=0.6.2 <0.9.0;
 pragma experimental ABIEncoderV2;
 
 contract InvariantTest {
-
     struct FuzzSelector {
         address addr;
         bytes4[] selectors;
@@ -92,5 +91,4 @@ contract InvariantTest {
         require(_targetedSenders.length != uint256(0), "NO_TARGETED_SENDERS");
         targetedSenders_ = _targetedSenders;
     }
-
 }
--- a/src/InvariantTest.sol
+++ b/src/InvariantTest.sol
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.2 <0.9.0;
 
-import {Test} from "./Test.sol";
+pragma experimental ABIEncoderV2;
 
-contract InvariantTest is Test {
+contract InvariantTest {
 
     struct FuzzSelector {
         address addr;
@@ -15,7 +15,20 @@ contract InvariantTest is Test {
     address[] private _targetedContracts;
     address[] private _targetedSenders;
 
+    string[] private _excludedArtifacts;
+    string[] private _targetedArtifacts;
+
+    FuzzSelector[] internal _targetedArtifactSelectors;
     FuzzSelector[] internal _targetedSelectors;
+
+    function excludeArtifact(string memory newExcludedArtifact_) internal {
+        _excludedArtifacts.push(newExcludedArtifact_);
+    }
+
+    function excludeArtifacts() public view returns (string[] memory excludedArtifacts_) {
+        require(_excludedArtifacts.length != uint256(0), "NO_EXCLUDED_ARTIFACTS");
+        excludedArtifacts_ = _excludedArtifacts;
+    }
 
     function excludeContract(address newExcludedContract_) internal {
         _excludedContracts.push(newExcludedContract_);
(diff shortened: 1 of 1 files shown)


mds1 (commented):
Thanks!

src/InvariantTest.sol
```diff
@@ -15,8 +14,21 @@ contract InvariantTest is Test {
     address[] private _targetedContracts;
     address[] private _targetedSenders;
 
+    string[] private _excludedArtifacts;
+    string[] private _targetedArtifacts;
+
+    FuzzSelector[] internal _targetedArtifactSelectors;
     FuzzSelector[] internal _targetedSelectors;
```
mds1, line 22: Some of these vars are `private` and others are `internal`. Is there a reason for this difference? Wondering which we should prefer

src/InvariantTest.sol
```diff
@@ -35,6 +47,24 @@ contract InvariantTest is Test {
         excludedSenders_ = _excludedSenders;
     }
 
+    function targetArtifact(string memory newTargetedArtifact_) internal {
+        _targetedArtifacts.push(newTargetedArtifact_);
+    }
```
mds1, line 52: @joshieDo do all of these methods dedupe the arrays in forge or should we check for duplicates when adding to the array?

src/InvariantTest.sol
```diff
@@ -35,6 +47,24 @@ contract InvariantTest is Test {
...
+
+    function targetArtifacts() public view returns (string[] memory targetedArtifacts_) {
+        require(_targetedArtifacts.length != uint256(0), "NO_TARGETED_ARTIFACTS");
```
mds1, line 55: How come the artifact handlers have `require` statements to ensure non-zero length but others don't? Feels like they should be consistent. Maybe a @joshieDo question

src/InvariantTest.sol
```diff
@@ -35,6 +47,24 @@ contract InvariantTest is Test {
...
+    }
+
+    function targetArtifactSelector(FuzzSelector memory newTargetedArtifactSelector_) internal {
+        _targetedArtifactSelectors.push(newTargetedArtifactSelector_);
+    }
+
+    function targetArtifactSelectors() public view returns (FuzzSelector[] memory targetedArtifactSelectors_) {
+        require(targetedArtifactSelectors_.length != uint256(0), "NO_TARGETED_ARTIFACT_SELECTORS");
+        targetedArtifactSelectors_ = _targetedArtifactSelectors;
+    }
```
mds1, line 66: Two UX improvement suggestions, lmk what you think:
1. Re-ordering the contract so the methods user's are intended to call (e.g. `targetContract`) are at the top, and ones forge calls (e.g. `targetContracts`, plural), are at the bottom. Maybe with some comments indicating this difference, to make it easier to learn how to use this file
2. Renaming `targetContract` to `addTargetContract` so it's more clear and descriptive

mds1 (commented):
A few more thoughts that came to mind

src/InvariantTest.sol
lucas-manuel: Good call out, I don't really see any advantage to `private`, what do you think? Who know someone might want access to these values in the inherited contracts.

src/InvariantTest.sol
lucas-manuel: Good question

src/InvariantTest.sol
lucas-manuel: Agree with both of these.

I remember asking about updating the naming conventions here a few months ago. I agree that its confusing that `targetContract` adds a target contract and `targetContracts` returns all targetedContracts.

I'm good with `addTargetContract` naming convention. Will make that change if people agree with it.

lucas-manuel: I wanted to do 1) but hesitated because I didn't know which section label convention to use 😆 I'll use whichever you guys like.

src/InvariantTest.sol
mds1: Hmm, I think I lean towards keeping it `private` until someone has a use case for needing access to the array directly, just a bit safer / more clear API that way since you can't accidentally overwrite it / directly set it

src/InvariantTest.sol
mds1: > I wanted to do 1) but hesitated because I didn't know which section label convention to use 😆 I'll use whichever you guys like.

Heh, what about something like this:

```solidity
// Functions for users: These are intended to be called in your tests

// Functions for forge: These are called by forge to run invariant tests. You
// do not need to use these or worry about them
```

src/InvariantTest.sol
joshieDo: they're deduped yeah.... but to be honest if we're talking about `targetArtifactSelector` they might be getting replaced instead of appended for the same artifact. i'll have to check it out.

src/InvariantTest.sol
joshieDo: afaik there's no need for that

PR #265, opened 2023-01-03
refactor: use simple revert instead of "require(false"

paulrberg:
Logically equivalent but more succinct, and clearer.


mds1 (approved):
Thanks!

PR #268, opened 2023-01-05
Usability Updates for StdChains

hexonaut:
Ping @mds1

Two useful updates to the `StdChains`.

1. Adds the `chainAlias` to the`Chain` struct. We need access to this as the chain alias is useful for defining chain-specific stuff inside configuration json files such as:

```
{
    "domains": {
        "mainnet": {
            "chainlog": "0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F"
        },
        "optimism": {
            "l1Messenger": "0x25ace71c97B33Cc4729CF772ae268934F7ab5fA1",
            "l2Messenger": "0x4200000000000000000000000000000000000007"
        },
        "arbitrum_one": {
            "inbox": "0x4Dbd4fc535Ac27206064B68FfCf827b0A60BAB3f",
            "arbSys": "0x0000000000000000000000000000000000000064"
        }
    }
}
```

This is nicer to define than using chain ids.

2. Allow automatic environment variable overrides via `ALIAS_RPC_URL`. For example a user can define `MAINNET_RPC_URL` to set the mainnet rpc url.

This saves having to repeat this pattern in all `foundry.toml` which will be common for us at least:

```
[rpc_endpoints]
mainnet = "${MAINNET_RPC_URL}"
goerli = "${GOERLI_RPC_URL}"
optimism = "${OPTIMISM_RPC_URL}"
... etc
```

### DIFF
--- a/src/StdChains.sol
+++ b/src/StdChains.sol
@@ -39,6 +39,12 @@ abstract contract StdChains {
 
     bool private initialized;
 
+    struct ChainData {
+        string name;
+        uint256 chainId;
+        string rpcUrl;
+    }
+
     struct Chain {
         // The chain name.
         string name;
@@ -90,7 +96,7 @@ abstract contract StdChains {
     }
 
     // set chain info, with priority to argument's rpcUrl field.
-    function setChain(string memory chainAlias, Chain memory chain) internal virtual {
+    function setChain(string memory chainAlias, ChainData memory chain) internal virtual {
         require(
             bytes(chainAlias).length != 0, "StdChains setChain(string,Chain): Chain alias cannot be the empty string."
         );
@@ -98,7 +104,6 @@ abstract contract StdChains {
         require(chain.chainId != 0, "StdChains setChain(string,Chain): Chain ID cannot be 0.");
 
         initialize();
-        chain.chainAlias = chainAlias;
         string memory foundAlias = idToAlias[chain.chainId];
 
         require(
@@ -117,7 +122,12 @@ abstract contract StdChains {
         uint256 oldChainId = chains[chainAlias].chainId;
         delete idToAlias[oldChainId];
 
-        chains[chainAlias] = chain;
+        chains[chainAlias] = Chain({
+            name: chain.name,
+            chainId: chain.chainId,
+            chainAlias: chainAlias,
+            rpcUrl: chain.rpcUrl
+        });
         idToAlias[chain.chainId] = chainAlias;
     }
 
@@ -164,40 +174,36 @@ abstract contract StdChains {
         initialized = true;
 
         // If adding an RPC here, make sure to test the default RPC URL in `testRpcs`
-        setChainWithDefaultRpcUrl("anvil", Chain("Anvil", 31337, "", "http://127.0.0.1:8545"));
+        setChainWithDefaultRpcUrl("anvil", ChainData("Anvil", 31337, "http://127.0.0.1:8545"));
         setChainWithDefaultRpcUrl(
-            "mainnet", Chain("Mainnet", 1, "", "https://mainnet.infura.io/v3/6770454bc6ea42c58aac12978531b93f")
+            "mainnet", ChainData("Mainnet", 1, "https://mainnet.infura.io/v3/6770454bc6ea42c58aac12978531b93f")
         );
         setChainWithDefaultRpcUrl(
-            "goerli", Chain("Goerli", 5, "", "https://goerli.infura.io/v3/6770454bc6ea42c58aac12978531b93f")
+            "goerli", ChainData("Goerli", 5, "https://goerli.infura.io/v3/6770454bc6ea42c58aac12978531b93f")
         );
         setChainWithDefaultRpcUrl(
-            "sepolia", Chain("Sepolia", 11155111, "", "https://sepolia.infura.io/v3/6770454bc6ea42c58aac12978531b93f")
+            "sepolia", ChainData("Sepolia", 11155111, "https://sepolia.infura.io/v3/6770454bc6ea42c58aac12978531b93f")
         );
-        setChainWithDefaultRpcUrl("optimism", Chain("Optimism", 10, "", "https://mainnet.optimism.io"));
-        setChainWithDefaultRpcUrl("optimism_goerli", Chain("Optimism Goerli", 420, "", "https://goerli.optimism.io"));
-        setChainWithDefaultRpcUrl("arbitrum_one", Chain("Arbitrum One", 42161, "", "https://arb1.arbitrum.io/rpc"));
+        setChainWithDefaultRpcUrl("optimism", ChainData("Optimism", 10, "https://mainnet.optimism.io"));
+        setChainWithDefaultRpcUrl("optimism_goerli", ChainData("Optimism Goerli", 420, "https://goerli.optimism.io"));
+        setChainWithDefaultRpcUrl("arbitrum_one", ChainData("Arbitrum One", 42161, "https://arb1.arbitrum.io/rpc"));
         setChainWithDefaultRpcUrl(
-            "arbitrum_one_goerli", Chain("Arbitrum One Goerli", 421613, "", "https://goerli-rollup.arbitrum.io/rpc")
+            "arbitrum_one_goerli", ChainData("Arbitrum One Goerli", 421613, "https://goerli-rollup.arbitrum.io/rpc")
         );
-        setChainWithDefaultRpcUrl("arbitrum_nova", Chain("Arbitrum Nova", 42170, "", "https://nova.arbitrum.io/rpc"));
-        setChainWithDefaultRpcUrl("polygon", Chain("Polygon", 137, "", "https://polygon-rpc.com"));
+        setChainWithDefaultRpcUrl("arbitrum_nova", ChainData("Arbitrum Nova", 42170, "https://nova.arbitrum.io/rpc"));
+        setChainWithDefaultRpcUrl("polygon", ChainData("Polygon", 137, "https://polygon-rpc.com"));
+        setChainWithDefaultRpcUrl("polygon_mumbai", ChainData("Polygon Mumbai", 80001, "https://rpc-mumbai.maticvigil.com"));
+        setChainWithDefaultRpcUrl("avalanche", ChainData("Avalanche", 43114, "https://api.avax.network/ext/bc/C/rpc"));
         setChainWithDefaultRpcUrl(
-            "polygon_mumbai", Chain("Polygon Mumbai", 80001, "", "https://rpc-mumbai.maticvigil.com")
+            "avalanche_fuji", ChainData("Avalanche Fuji", 43113, "https://api.avax-test.network/ext/bc/C/rpc")
         );
-        setChainWithDefaultRpcUrl("avalanche", Chain("Avalanche", 43114, "", "https://api.avax.network/ext/bc/C/rpc"));
-        setChainWithDefaultRpcUrl(
-            "avalanche_fuji", Chain("Avalanche Fuji", 43113, "", "https://api.avax-test.network/ext/bc/C/rpc")
-        );
-        setChainWithDefaultRpcUrl(
-            "bnb_smart_chain", Chain("BNB Smart Chain", 56, "", "https://bsc-dataseed1.binance.org")
-        );
-        setChainWithDefaultRpcUrl("bnb_smart_chain_testnet", Chain("BNB Smart Chain Testnet", 97, "", "https://data-seed-prebsc-1-s1.binance.org:8545"));// forgefmt: disable-line
-        setChainWithDefaultRpcUrl("gnosis_chain", Chain("Gnosis Chain", 100, "", "https://rpc.gnosischain.com"));
+        setChainWithDefaultRpcUrl("bnb_smart_chain", ChainData("BNB Smart Chain", 56, "https://bsc-dataseed1.binance.org"));
+        setChainWithDefaultRpcUrl("bnb_smart_chain_testnet", ChainData("BNB Smart Chain Testnet", 97, "https://data-seed-prebsc-1-s1.binance.org:8545"));// forgefmt: disable-line
+        setChainWithDefaultRpcUrl("gnosis_chain", ChainData("Gnosis Chain", 100, "https://rpc.gnosischain.com"));
     }
 
     // set chain info, with priority to chainAlias' rpc url in foundry.toml
-    function setChainWithDefaultRpcUrl(string memory chainAlias, Chain memory chain) private {
+    function setChainWithDefaultRpcUrl(string memory chainAlias, ChainData memory chain) private {
         string memory rpcUrl = chain.rpcUrl;
         defaultRpcUrls[chainAlias] = rpcUrl;
         chain.rpcUrl = "";
--- a/src/StdChains.sol
+++ b/src/StdChains.sol
@@ -198,7 +198,10 @@ abstract contract StdChains {
         setChainWithDefaultRpcUrl(
             "bnb_smart_chain", ChainData("BNB Smart Chain", 56, "https://bsc-dataseed1.binance.org")
         );
-        setChainWithDefaultRpcUrl("bnb_smart_chain_testnet", ChainData("BNB Smart Chain Testnet", 97, "https://data-seed-prebsc-1-s1.binance.org:8545"));// forgefmt: disable-line
+        setChainWithDefaultRpcUrl(
+            "bnb_smart_chain_testnet",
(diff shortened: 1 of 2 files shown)


src/StdChains.sol
```diff
@@ -118,18 +121,31 @@ abstract contract StdChains {
         idToAlias[chain.chainId] = chainAlias;
     }
 
+    function _toUpper(string memory str) private pure returns (string memory) {
+		bytes memory strb = bytes(str);
+		bytes memory copy = new bytes(strb.length);
```
iamchrissmith, line 126:
```suggestion
        bytes memory strb = bytes(str);
        bytes memory copy = new bytes(strb.length);
```

src/StdChains.sol
```diff
@@ -118,18 +121,31 @@ abstract contract StdChains {
...
         private
-        view
         returns (Chain memory)
     {
         if (bytes(chain.rpcUrl).length == 0) {
             try vm.rpcUrl(chainAlias) returns (string memory configRpcUrl) {
                 chain.rpcUrl = configRpcUrl;
             } catch (bytes memory err) {
-                chain.rpcUrl = defaultRpcUrls[chainAlias];
+                chain.rpcUrl = vm.envOr(string(abi.encodePacked("RPC_URL_", _toUpper(chainAlias))), defaultRpcUrls[chainAlias]);
```
iamchrissmith, line 148:
```suggestion
                chain.rpcUrl =
                    vm.envOr(string(abi.encodePacked("RPC_URL_", _toUpper(chainAlias))), defaultRpcUrls[chainAlias]);
```

src/StdChains.sol
```diff
@@ -150,32 +166,32 @@ abstract contract StdChains {
...
+        setChainWithDefaultRpcUrl("polygon_mumbai", Chain("Polygon Mumbai", 80001, "", "https://rpc-mumbai.maticvigil.com"));
+        setChainWithDefaultRpcUrl("avalanche", Chain("Avalanche", 43114, "", "https://api.avax.network/ext/bc/C/rpc"));
         setChainWithDefaultRpcUrl(
-            "avalanche_fuji", Chain("Avalanche Fuji", 43113, "https://api.avax-test.network/ext/bc/C/rpc")
+            "avalanche_fuji", Chain("Avalanche Fuji", 43113, "", "https://api.avax-test.network/ext/bc/C/rpc")
         );
-        setChainWithDefaultRpcUrl("bnb_smart_chain", Chain("BNB Smart Chain", 56, "https://bsc-dataseed1.binance.org"));
-        setChainWithDefaultRpcUrl("bnb_smart_chain_testnet", Chain("BNB Smart Chain Testnet", 97, "https://data-seed-prebsc-1-s1.binance.org:8545"));// forgefmt: disable-line
-        setChainWithDefaultRpcUrl("gnosis_chain", Chain("Gnosis Chain", 100, "https://rpc.gnosischain.com"));
+        setChainWithDefaultRpcUrl("bnb_smart_chain", Chain("BNB Smart Chain", 56, "", "https://bsc-dataseed1.binance.org"));
```
iamchrissmith, line 192:
```suggestion
        setChainWithDefaultRpcUrl(
            "bnb_smart_chain", Chain("BNB Smart Chain", 56, "", "https://bsc-dataseed1.binance.org")
        );
```

src/StdChains.sol
```diff
@@ -150,32 +166,32 @@ abstract contract StdChains {
...
-            "arbitrum_one_goerli", Chain("Arbitrum One Goerli", 421613, "https://goerli-rollup.arbitrum.io/rpc")
+            "arbitrum_one_goerli", Chain("Arbitrum One Goerli", 421613, "", "https://goerli-rollup.arbitrum.io/rpc")
         );
-        setChainWithDefaultRpcUrl("arbitrum_nova", Chain("Arbitrum Nova", 42170, "https://nova.arbitrum.io/rpc"));
-        setChainWithDefaultRpcUrl("polygon", Chain("Polygon", 137, "https://polygon-rpc.com"));
-        setChainWithDefaultRpcUrl("polygon_mumbai", Chain("Polygon Mumbai", 80001, "https://rpc-mumbai.maticvigil.com"));
-        setChainWithDefaultRpcUrl("avalanche", Chain("Avalanche", 43114, "https://api.avax.network/ext/bc/C/rpc"));
+        setChainWithDefaultRpcUrl("arbitrum_nova", Chain("Arbitrum Nova", 42170, "", "https://nova.arbitrum.io/rpc"));
+        setChainWithDefaultRpcUrl("polygon", Chain("Polygon", 137, "", "https://polygon-rpc.com"));
+        setChainWithDefaultRpcUrl("polygon_mumbai", Chain("Polygon Mumbai", 80001, "", "https://rpc-mumbai.maticvigil.com"));
```
iamchrissmith, line 187:
```suggestion
        setChainWithDefaultRpcUrl(
            "polygon_mumbai", Chain("Polygon Mumbai", 80001, "", "https://rpc-mumbai.maticvigil.com")
        );
```

iamchrissmith:
one last format error that is showing up locally, but GH won't let me suggest a change:

the formatter wants this function declaration https://github.com/foundry-rs/forge-std/pull/268/files#diff-e21e7255e73417cde7931c5f5ffbe44f1009bbe12ccdf684942299c83448f7ddR140-R143

to be one line:

```solidity
function getChainWithUpdatedRpcUrl(string memory chainAlias, Chain memory chain) private returns (Chain memory) {
```

hexonaut:
Fixed all formatting.

src/StdChains.sol
```diff
@@ -118,18 +121,29 @@ abstract contract StdChains {
         idToAlias[chain.chainId] = chainAlias;
     }
 
+    function _toUpper(string memory str) private pure returns (string memory) {
```
mds1, line 124: Do we have a source we can link to for where this came from originally? (assuming you didn't write it just for this PR, but maybe you did)

src/StdChains.sol
```diff
@@ -118,18 +121,29 @@ abstract contract StdChains {
...
-    {
+    // current -> config (foundry.toml) -> environment variable -> default
+    function getChainWithUpdatedRpcUrl(string memory chainAlias, Chain memory chain) private returns (Chain memory) {
         if (bytes(chain.rpcUrl).length == 0) {
             try vm.rpcUrl(chainAlias) returns (string memory configRpcUrl) {
                 chain.rpcUrl = configRpcUrl;
             } catch (bytes memory err) {
-                chain.rpcUrl = defaultRpcUrls[chainAlias];
+                chain.rpcUrl =
+                    vm.envOr(string(abi.encodePacked("RPC_URL_", _toUpper(chainAlias))), defaultRpcUrls[chainAlias]);
```
mds1, line 146: Thoughts on `ALIAS_RPC_URL` instead of `RPC_URL_ALIAS`? I'm personally ok with either, but from a quick search it seems the former is a much more common naming scheme—1.1k results for the [former](https://sourcegraph.com/search?q=context:global+MAINNET_RPC_URL+count:5000&patternType=standard&sm=1), 101 for the [latter](https://sourcegraph.com/search?q=context:global+RPC_URL_MAINNET+count:5000&patternType=standard&sm=1)

image
image

src/StdChains.sol
```diff
@@ -150,32 +164,36 @@ abstract contract StdChains {
         initialized = true;
 
         // If adding an RPC here, make sure to test the default RPC URL in `testRpcs`
-        setChainWithDefaultRpcUrl("anvil", Chain("Anvil", 31337, "http://127.0.0.1:8545"));
+        setChainWithDefaultRpcUrl("anvil", Chain("Anvil", 31337, "", "http://127.0.0.1:8545"));
```
mds1, line 167: Did anything change in this section aside from adding the `""` throughout for chain alias? I don't think so but the diff came out annoying so just checking 😅

src/StdChains.sol
```diff
@@ -44,6 +44,8 @@ abstract contract StdChains {
         string name;
         // The chain's Chain ID.
         uint256 chainId;
+        // The chain's alias.
```
mds1, line 47: Nit, may as well as add a bit more detail
```suggestion
        // The chain's alias (i.e. what gets specified in `foundry.toml`).
```

mds1 (commented):
I like this idea, thanks a lot! @ZeroEkkusu you led a lot of the stdChains work so would love for you to take a look if you have time

src/StdChains.sol
mds1: Also this is a breaking change to the interface which I'd like to avoid if we can. What do you think of creating a new struct called `ChainData` and using that as the chain return data and keeping `Chain` as the input data?

src/StdChains.sol
hexonaut: I wrote it. :)

src/StdChains.sol
hexonaut: I'm okay with switching the order. I suppose putting the alias in front matches the style of `ETH_RPC_URL` that dapp tools defined.

hexonaut: done https://github.com/foundry-rs/forge-std/pull/268/commits/758fd9066dca595a71440c7942e8164d4b4c473c

src/StdChains.sol
hexonaut: done https://github.com/foundry-rs/forge-std/pull/268/commits/6ac3ebb618a2ea239860aeb45bddf506e6ec8ff7

src/StdChains.sol
hexonaut: > Did anything change in this section aside from adding the `""` throughout for chain alias? I don't think so but the diff came out annoying so just checking 😅

Yes this is just the `""` addition.

> Also this is a breaking change to the interface which I'd like to avoid if we can. What do you think of creating a new struct called `ChainData` and using that as the chain return data and keeping `Chain` as the input data?

Changing the output is also a breaking change. Does it make more sense to break the input or the output? If I had to guess breaking the output would actually be a larger breaking change as this is people defining custom chains vs using the common ones defined.

I can proceed if you think this is a better way to go, but wanted to double check.

ZeroEkkusu (commented):
Is the point of change 1 to be able to get chainAlias from a Chain? If so, what if instead of a breaking change, you add getAlias function that takes a Chain as a parameter, uses its id to get a Chain from getChain and then compares their hashes to see if they’re equal. Is so, it returns the alias from idToAlias. Otherwise, reverts.

ZeroEkkusu:
You have more insight, since you’re using it.

One thing I thought about before is whether anyone uses `Chain.name`. It seemed more useful to be `Chain(“chain_alias”, 123, “https://…”)`.

You should go with what’s more practical in general.

hexonaut:
That's a good point. Other than maybe log output I don't think the human-friendly name is all that useful, and even if it is it's not hard for users to define this outside of `StdChains`. Perhaps we can just use the chain alias in the name so things don't break?

ZeroEkkusu:
I think it’s the best solution DX wise, for example, you won’t need to pass a key and Chain in setChain anymore, only a Chain.

This solution is also a breaking change.

hexonaut:
We could also just leave the interface and by convention use the chain alias for the name for the defaults.

ZeroEkkusu:
There are quite a few options - I’ll let @mds1 and you decide, and pitch in if necessary.

mds1:
Thanks for the discussion here @hexonaut and @ZeroEkkusu. One other thought is that if we make this a breaking change to decouple the `Chain` input struct and the `ChainData` return struct (or whatever names we go with), that does set us up to expand the returned `ChainData` in the future, which could be useful. For example, you could imagine adding a `bool supportsEip1559` flag or something.

So perhaps we just go that route with the breaking change? Was hoping to avoid breaking changes but since we can't (and I agree having alias as a field makes sense), we may as well go with the one that sets us up best going forward. Which I think is adding a new struct so `setChain` can stay the same for now, and we have a breaking change on the return data struct to add the `alias`.

hexonaut:
Ok sounds good. Will make this change in a moment.

src/StdChains.sol
```diff
@@ -39,11 +39,19 @@ abstract contract StdChains {
 
     bool private initialized;
 
+    struct ChainData {
```
mds1, line 42: Nit, wdyt of switching the names of the `Chain` and `ChainData` structs? `ChainData` sounds like broader scope and the return data has more info. Def open to keeping this as-is though too. Otherwise this PR LGTM

mds1: Ah just saw [your comment](https://github.com/foundry-rs/forge-std/pull/268#issuecomment-1375961973), ok that works for me

mds1:
Just needs `forge fmt`, @ZeroEkkusu do you want to take a quick look also before merging?

mds1 (approved):
Thank you!

src/StdChains.sol
```diff
@@ -88,7 +96,7 @@ abstract contract StdChains {
     }
 
     // set chain info, with priority to argument's rpcUrl field.
-    function setChain(string memory chainAlias, Chain memory chain) internal virtual {
+    function setChain(string memory chainAlias, ChainData memory chain) internal virtual {
         require(
             bytes(chainAlias).length != 0, "StdChains setChain(string,Chain): Chain alias cannot be the empty string."
```
ZeroEkkusu, line 101: Can you update these error messages to say `ChainData`, and the description on top of the file?

Also, do you think it's useful, to be able to "convert" `Chain` into `ChainData` so you can pass it to `setChain`, by adding an alternative signature:

```
chain = getChain(...)
chain.name = ...
setChain(..., chain)
```

Otherwise, LGTM, @mds1 can merge when ready.

hexonaut: Done. https://github.com/foundry-rs/forge-std/pull/268/commits/43ff37017acaa235d5c99d42b73b7fa665a530df

ZeroEkkusu (approved):

src/StdChains.sol
```diff
@@ -114,22 +124,39 @@ abstract contract StdChains {
...
 
-        chains[chainAlias] = chain;
+        chains[chainAlias] =
+            Chain({name: chain.name, chainId: chain.chainId, chainAlias: chainAlias, rpcUrl: chain.rpcUrl});
         idToAlias[chain.chainId] = chainAlias;
     }
 
+    // set chain info, with priority to argument's rpcUrl field.
+    function setChain(string memory chainAlias, Chain memory chain) internal virtual {
+        setChain(chainAlias, ChainData({name: chain.name, chainId: chain.chainId, rpcUrl: chain.rpcUrl}));
```
ZeroEkkusu, line 134: We can require that `chain.chainAlias` is empty or equal to `chainAlias`. Otherwise, revert with `StdChains setChain(string,Chain): Will not convert Chain to ChainData. chainAlias is "abc", but Chain.chainAlias is "xyz". Make sure you are passing the right Chain.`.

hexonaut: I'm okay with either. @mds1 lmk your thoughts.

mds1: Right now `setChain` takes `ChainData` as input which doesn't have the `chainAlias` field. Then we set the chain alias in storage for the full `Chain` struct. So the current approach seems good to me unless I'm misunderstanding something?

ZeroEkkusu: It’s to protect users from accidentally pressing a wrong chain. Because this function reduces Chain to ChainData, it is possible to do setChain(“chain_1”, chain2). If the intention is not to assign “chain_1” to chain2, the function should let the users now of a possible error. In practice, it should happen rarely.

If you don’t think it should be there, you can merge the PR.

ZeroEkkusu: @mds1  Note that this is a new function to let you pass a Chain instead of ChainData for the ease of use. This is since Chain has enough data to satisfy the ChainData requirement (but not the other way around).

mds1: Another idea is just `setChain(Chain memory chain)` and removing the separate `chainAlias` parameter. Wdyt of merging this to avoid churn, and we can follow up in a separate PR to consider implementing that overload? Since that shouldn't be a breaking change

ZeroEkkusu: If you think so, yes - I don’t want to block the PR. The overload is already implemented, btw.

Re setChain(Chain): It’s possible that chain.chainAlias is not assigned; that would need to be checked. The original idea was just to not need to convert Chain to ChainData manually, if you have a Chain. The same signature, so things are simple. But, yes, there could be an alternative way of setting a Chain (this one).

mds1: Ah whoops yes it's already implemented, sorry 😅 

I'm going to merge this as is since CI passed, and @hexonaut or @ZeroEkkusu if either of you wants to open a follow up PR with the above changes to make setting safer, that should be a quick one to merge. Thanks all!

PR #270, opened 2023-01-06
add new assumePayable cheat

hexcowboy:
Reference issue: https://github.com/foundry-rs/foundry/issues/3631

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -463,6 +463,17 @@ abstract contract StdCheatsSafe {
             vm.resumeGasMetering();
         }
     }
+
+    // a cheat for fuzzing addresses that are payable only
+    // see https://github.com/foundry-rs/foundry/issues/3631
+    function assumePayable(address addr) internal virtual {
+        // VM address
+        vm.assume(addr != 0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);
+        // Console address
+        vm.assume(addr != 0x000000000000000000636F6e736F6c652e6c6f67);
+        // Create2Deployer
+        vm.assume(addr != 0x4e59b44847b379578588920cA78FbF26c0B4956C);
+    }
 }
 
 // Wrappers around cheatcodes to avoid footguns
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -473,6 +473,8 @@ abstract contract StdCheatsSafe {
         vm.assume(addr != 0x000000000000000000636F6e736F6c652e6c6f67);
         // Create2Deployer
         vm.assume(addr != 0x4e59b44847b379578588920cA78FbF26c0B4956C);
+        // precompiles are also not payable
+        assumeNoPrecompiles(addr);
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/StdCheats.sol
```diff
@@ -463,6 +463,19 @@ abstract contract StdCheatsSafe {
...
+    function assumePayable(address addr) internal virtual {
+        // VM address
+        vm.assume(addr != 0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);
+        // Console address
+        vm.assume(addr != 0x000000000000000000636F6e736F6c652e6c6f67);
+        // Create2Deployer
+        vm.assume(addr != 0x4e59b44847b379578588920cA78FbF26c0B4956C);
+        // precompiles are also not payable
+        assumeNoPrecompiles(addr);
+    }
```
mds1, line 478: The VM address, console address, and create2 address all will revert when trying to send ETH to them with `.transfer` or `.call` to it (the first two are because the forge VM won't understand the calldata, but on a live chain they can receive ETH because they have no code), so that seems good. 

But:
1. Some precompiles can receive ETH, e.g. `payable(address(2)).transfer(1 ether)` works while `payable(address(1)).transfer(1 ether)` reverts
2. Maybe precompiles on other EVM chains expect to receive ETH transfers? I'm not sure

So perhaps we remove the `assumeNoPrecompiles`, and users can manually stack the two if they want? Not certain on the best option 🤔

Let's also add a test or two like we have for `assumeNoPrecompiles`

hexcowboy:
@mds1 I may have found a better way to write this cheat, i've modified the source

with a panic assertion, there are more benefits to the initial approach
- no need to add new addresses in the future (future proof)
- this will also handle all precompile addresses
- in addition, if an end user creates a non-payable contract it will also exclude that

src/StdCheats.sol
```diff
@@ -463,6 +464,15 @@ abstract contract StdCheatsSafe {
             vm.resumeGasMetering();
         }
     }
+
+    // a cheat for fuzzing addresses that are payable only
+    // see https://github.com/foundry-rs/foundry/issues/3631
+    function assumePayable(address addr) internal virtual {
+        try PanicAssertions.assertPayable(addr) {}
```
mds1, line 471: Clever approach! LGTM, let's just remove this unnecessary PanicAssertions library and inline the call here instead

hexcowboy: totally agree the PanicAssertions library seems unnecessary but unfortunately the try/catch statement "can only catch errors from external function calls and contract creation." - [link](https://solidity-by-example.org/try-catch/)

if you propose a better place to put the external function, i can move it

mds1: would a low level call work instead? 

```solidity
(bool success, ) = payable(addr).call{value: 0}("");
vm.assume(success);
```

if not, maybe an external function in the same contract and call it with `this.isPayable(addr)` with a comment explaining that method's purpose so user's know they don't need to call it

hexcowboy: looks like low level call works, thanks

[9dbf1a4](https://github.com/foundry-rs/forge-std/pull/270/commits/9dbf1a47505bc7a6dd8a203f52e8d7592c759d0c)

mds1 (approved):
Thank you!

PR #271, opened 2023-01-06
feat: add "IMulticall3" interface

PaulRBerg:
## Description

As discussed in #266, this PR adds the `IMulticall3` interfaces in this repo (hardcoded, not imported from anywhere else). This PR also adds a `getTokenBalances` util in `StdUtils`, which I think will prove quite useful to Multicall users.

Note: I did not write tests for `getTokenBalances` just yet. Not sure if you are happy to merge the PR as is, or you would rather want me/ us to test it first.

## Full Changelog

- [x] chore: fix typos in comments
- [x] feat: add "getTokenBalances" util
- [x] feat: add "IMulticall3" interface
- [x] refactor: add header separators in "StdUtils"
- [x] refactor: order functions alphabetically in "StdUtils"

### DIFF
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -122,7 +122,11 @@ abstract contract StdUtils {
         virtual
         returns (uint256[] memory balances)
     {
-        require(token.code.length > 0, "StdUtils getTokenBalances(address,address[]): Token address is not a contract.");
+        uint256 tokenCodeSize;
+        assembly {
+            tokenCodeSize := extcodesize(tokenCodeSize)
+        }
+        require(tokenCodeSize > 0, "StdUtils getTokenBalances(address,address[]): Token address is not a contract.");
 
         // ABI encode the aggregate call to Multicall3.
         uint256 length = addresses.length;
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.2 <0.9.0;
 
+pragma experimental ABIEncoderV2;
+
 import {IMulticall3} from "./interfaces/IMulticall3.sol";
 // TODO Remove import.
 import {VmSafe} from "./Vm.sol";
@@ -120,10 +122,7 @@ abstract contract StdUtils {
         uint256 length = addresses.length;
         IMulticall3.Call[] memory calls = new IMulticall3.Call[](length);
         for (uint256 i = 0; i < length; ++i) {
-            calls[i] = IMulticall3.Call({
-                target: token,
-                callData: abi.encodeWithSelector(0x70a08231, (addresses[i]))
-            });
+            calls[i] = IMulticall3.Call({target: token, callData: abi.encodeWithSelector(0x70a08231, (addresses[i]))});
         }
 
         // Make the aggregate call.
@@ -136,7 +135,7 @@ abstract contract StdUtils {
         }
     }
 
-        /*//////////////////////////////////////////////////////////////////////////
+    /*//////////////////////////////////////////////////////////////////////////
                                  PRIVATE FUNCTIONS
     //////////////////////////////////////////////////////////////////////////*/
 
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -116,13 +116,9 @@ abstract contract StdUtils {
         return addressFromLast20Bytes(keccak256(abi.encodePacked(bytes1(0xff), deployer, salt, initcodeHash)));
     }
 
-    // Performs a single call with IMulticall3 to query the ERC-20 token balances of the given addresses.
-    function getTokenBalances(address token, address[] memory addresses)
-        internal
-        virtual
-        returns (uint256[] memory balances)
-    {
-        // ABI encode the aggregate call to IMulticall3.
+    // Performs a single call with Multicall3 to query the ERC-20 token balances of the given addresses.
+    function getTokenBalances(address token, address[] memory addresses) internal returns (uint256[] memory balances) {
+        // ABI encode the aggregate call to Multicall3.
         uint256 length = addresses.length;
         IMulticall3.Call[] memory calls = new IMulticall3.Call[](length);
         for (uint256 i = 0; i < length; ++i) {
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -3,7 +3,6 @@ pragma solidity >=0.6.2 <0.9.0;
 
 pragma experimental ABIEncoderV2;
 
-import {console2} from "./console2.sol";
 import {IMulticall3} from "./interfaces/IMulticall3.sol";
 // TODO Remove import.
 import {VmSafe} from "./Vm.sol";
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -3,6 +3,7 @@ pragma solidity >=0.6.2 <0.9.0;
 
 pragma experimental ABIEncoderV2;
 
+import {console2} from "./console2.sol";
 import {IMulticall3} from "./interfaces/IMulticall3.sol";
 // TODO Remove import.
 import {VmSafe} from "./Vm.sol";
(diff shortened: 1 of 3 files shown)


src/StdUtils.sol
```diff
@@ -100,11 +114,32 @@ abstract contract StdUtils {
         return addressFromLast20Bytes(keccak256(abi.encodePacked(bytes1(0xff), deployer, salt, initcodeHash)));
     }
 
-    function bytesToUint(bytes memory b) internal pure virtual returns (uint256) {
-        require(b.length <= 32, "StdUtils bytesToUint(bytes): Bytes length exceeds 32.");
-        return abi.decode(abi.encodePacked(new bytes(32 - b.length), b), (uint256));
+    // Performs a single call with Multicall3 to query the ERC-20 token balances of the given addresses.
+    function getTokenBalances(address token, address[] memory addresses) internal returns (uint256[] memory balances) {
```
mds1, line 118: Tests for this method would be great 🙏

paulrberg: I will write a few later this week.

src/interfaces/IMulticall3.sol
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.12 <0.9.0;
```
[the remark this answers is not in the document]
paulrberg: Thanks, fixed below.

src/StdUtils.sol
```diff
@@ -1,18 +1,27 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.2 <0.9.0;
 
+import {IMulticall3} from "./interfaces/IMulticall3.sol";
```
mds1, line 4: Ah this file also needs `pragma experimental ABIEncoderV2;` now

src/interfaces/IMulticall3.sol
```diff
@@ -0,0 +1,73 @@
...
+
+    function getBlockHash(uint256 blockNumber) external view returns (bytes32 blockHash);
+
+    function getBlockNumber() external view returns (uint256 blockNumber);
+
+    function getChainId() external view returns (uint256 chainid);
+
+    function getCurrentBlockCoinbase() external view returns (address coinbase);
+
+    function getCurrentBlockDifficulty() external view returns (uint256 difficulty);
```
sambacha, line 54: this will return 00000.... post merge, why not replace with prevrando?

sambacha: Yes its an alias, either way it should be changed or add prevrando if support for PoW chains is wanted I suppose.

👍

sambacha: Actually DIFFICULTY will return zeroed out response in the block im pretty sure

mds1: Given that the contract is already deployed on 50+ chains, I don't think this is worth changing. Users just need to understand how it behaves differently across different chains. Similarly, there is a `getBasefee` method that will revert on chains that don't support EIP-1559. Perhaps a PR to the README to document these differences would be valuable

sambacha: yea did not realize this was your deployed contract 🙃 

💯

mds1: Created https://github.com/mds1/multicall/issues/57 to track the documentation changes, thanks 🙂

Going to mark this comment resolved

src/StdUtils.sol
paulrberg: @mds1 I started testing this function but there must be something that I misunderstand about Multicall or fork testing or both because I cannot get the revert test to pass ..

I wrote this:

```solidity
modifier forkMainnet() {
    vm.createSelectFork({urlOrAlias: "mainnet", blockNumber: 16376000});
    _;
}

function testCannotGetTokenBalances_NonTokenContract() external forkMainnet {
    address token = address(1729);
    address[] memory addresses = new address[](1);
    addresses[0] = USDC_HOLDER;
    vm.expectRevert("Multicall3: call failed");
    getTokenBalances(token, addresses);
}
```

The test fails though I think it should pass because there is no token contract at address 1729 on Ethereum Mainnet .. the call [should revert with "Multicall3: call failed"](https://github.com/mds1/multicall/blob/eb34ad2954f9ceb475a24bb0155bff3bef0f5409/src/Multicall3.sol#L50), but it doesn't.

I also tried replacing the expect revert with the following:

```solidity
vm.expectRevert();
vm.expectRevert(bytes(""));
```

Nothing worked.

mds1: This is working as expected actually. A tx that sends calldata to an address without code is considered a successful transaction by the EVM and it doesn't revert. Normally if doing this in solidity, the revert comes from solidity: the `IERC20` interface (for example) that you defined says `balanceOf` should return a uint256, and when solidity sees that no data was returned because you called an address with no code, it reverts. 

You can probably solve this by instead calling a solidity contract that has code, but doesn't implement the `balanceOf` function and has no fallback method (maybe the uniswap factory would work for this?). The calldata will not match anything in the function dispatcher and the called contract should revert as a result

PaulRBerg: > A tx that sends calldata to an address without code is considered a successful transaction by the EVM and it doesn't revert.

TIL 😬

> You can probably solve this by instead calling a solidity contract that has code, but doesn't implement the balanceOf function and has no fallback method

Great idea, will do 🫡

mds1: Any update here? Planning to do a new forge-std release once this PR gets in, would love to do it this week if possible

PaulRBerg: Sorry @mds1, I started working on this last week but I bumped into a blocker. I should have posted an update here.

I picked up my work again, and I still bumping into the same blocker.

The issue is related to forking mainnet. To test `getTokenBalance`, I need to rewrite the `StdUtils` tests to contain a modifier for forking mainnet, and I also need to rewrite the test contract to call a mock rather than an internal function (If I don't this, the function does not revert on `multicall.aggregate(calls)` - it reverts due to an "index out of bounds" error trying to access the returned data which is nil).

I wrote this:

```solidity
    StdUtilsMock internal utils = new StdUtilsMock();

    address internal USDC_HOLDER = 0xDa9CE944a37d218c3302F6B82a094844C6ECEb17;

    modifier forkMainnet() {
        vm.createSelectFork({urlOrAlias: "mainnet", blockNumber: 16376000});
        _;
    }

    function testCannotGetTokenBalances_NonTokenContract() external forkMainnet {
        // The UniswapV2Factory contract has neither a "balanceOf" function nor a fallback function.
        address token = address(0x5C69bEe701ef814a2B6a3EDD4B1652CB9cc5aA6f);
        address[] memory addresses = new address[](1);
        addresses[0] = USDC_HOLDER;
        vm.expectRevert();
        utils.getTokenBalances_(token, addresses);
    }
```

But I cannot for the life of me understand why it doesn't work. Even if you add a dummy `foo` function in the `StdUtilsMock` contract, and call it in the test function, the call still reverts.

Might the problem be related to the deployment of the `utils` mock? Can deployments not be made when forking?

Or is the issue related to the RPC URL? I tried setting my own key, but I observed the same behavior.

mds1: Just undid your last commit, then pushed a commit which seems to work as expected!

PaulRBerg: Thank you!

The issue was that I was deploying the mock contract at the contract level (right underneath `StdUtilsTest`), which means the mock was being deployed to the local network.

You added it in the test function body, where it gets deployed to the mainnet fork - as it should.

PaulRBerg: @mds1 - I added more tests:

- EOA as an input, should revert
- USDC holders
- SHIB holders

The PR should be ready for merging, pending your review of course.

mds1: Pushed two small commits to fix a typo + resolve failing CI, should be good to go once it passes. Thanks a lot for this! 🙌

mds1 (approved):

src/StdUtils.sol
PaulRBerg: Awesome stuff, thanks @mds1 for the quick review. And yeah sorry, I keep forgetting about supporting older solc versions in my PRs.

PR #274, opened 2023-01-10
feat: parseJSON cheatcodes with type coercion

odyslam:
Should be merged after https://github.com/foundry-rs/foundry/pull/4061

### DIFF
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -33,137 +33,210 @@ interface VmSafe {
 
     // Loads a storage slot from an address
     function load(address target, bytes32 slot) external view returns (bytes32 data);
+
     // Signs data
     function sign(uint256 privateKey, bytes32 digest) external pure returns (uint8 v, bytes32 r, bytes32 s);
+
     // Gets the address for a given private key
     function addr(uint256 privateKey) external pure returns (address keyAddr);
+
     // Gets the nonce of an account
     function getNonce(address account) external view returns (uint64 nonce);
+
     // Performs a foreign function call via the terminal
     function ffi(string[] calldata commandInput) external returns (bytes memory result);
+
     // Sets environment variables
     function setEnv(string calldata name, string calldata value) external;
+
     // Reads environment variables, (name) => (value)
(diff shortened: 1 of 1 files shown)


odyslam:
Failed test seems irrelevant to the PR

src/Vm.sol
```diff
@@ -33,137 +33,210 @@ interface VmSafe {
 
     // Loads a storage slot from an address
     function load(address target, bytes32 slot) external view returns (bytes32 data);
+
```
mds1, line 36: Why were all these blank lines introduced? We should undo that

src/Vm.sol
```diff
@@ -183,50 +256,97 @@ interface VmSafe {
...
     function parseJson(string calldata json) external pure returns (bytes memory abiEncodedData);
 
+    // The following parseJson cheatcodes will do type coercion, for the type that they indicate.
+    // For example, parseJsonUint will coerce all values to a uint256. That includes stringified numbers '12'
+    // and hex numbers '0xEF'.
+    // Type coercion works ONLY for discrete values or arrays. That means that the key must return a value or array, not
+    // a JSON object.
+
+    function parseJsonUint(string calldata, string calldata) external returns (uint256);
+
```
mds1, line 269: And we should remove the blank lines between all these new cheats too

odyslam: This is the result of running `forge fmt`. It's more legible this way, no?

mds1: Maybe you're on an old forge version or have some custom config? But for me and in CI `forge fmt` doesn't introduce line breaks. Just pushed dac8350 to match the original formatting for now to reduce the diff (give that a review to make sure I didn't accidentally remove anything, then we can merge). Open to organizing/cleaning up the `Vm` file in a separate issue/PR

odyslam: @mds1 uuuh, didn't think about that. Probs you are right. lgtm

odyslam:
LGTM

mds1 (approved):
Thanks!

PR #276, opened 2023-01-20
Add create2 helper methods

karmacoma-eth:
- computeCreate2Address(bytes32 salt, bytes32 initCodeHash) returns an address using the default create2 deployer
- hashInitCode(bytes, bytes) and hashInitCodeNoConstructorArgs(bytes) make it more convenient to figure out the init code hash in deploy scripts

### DIFF
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -19,7 +19,9 @@ abstract contract StdUtils {
         57896044618658097711785492504343953926634992332820282019728792003956564819968;
     uint256 private constant UINT256_MAX =
         115792089237316195423570985008687907853269984665640564039457584007913129639935;
-    address private constant DEFAULT_CREATE2_DEPLOYER = 0x4e59b44847b379578588920cA78FbF26c0B4956C;
+
+    // Used by default when deploying with create2, https://github.com/Arachnid/deterministic-deployment-proxy.
+    address private constant CREATE2_FACTORY = 0x4e59b44847b379578588920cA78FbF26c0B4956C;
 
     /*//////////////////////////////////////////////////////////////////////////
                                  INTERNAL FUNCTIONS
@@ -118,20 +120,20 @@ abstract contract StdUtils {
     }
 
     /// @dev returns the address of a contract created with CREATE2 using the default CREATE2 deployer
-    function computeCreate2Address(bytes32 salt, bytes32 _initCodeHash) internal pure returns (address) {
-        return computeCreate2Address(salt, _initCodeHash, DEFAULT_CREATE2_DEPLOYER);
+    function computeCreate2Address(bytes32 salt, bytes32 initCodeHash) internal pure returns (address) {
+        return computeCreate2Address(salt, initCodeHash, CREATE2_FACTORY);
     }
 
     /// @dev returns the hash of the init code (creation code + no args) used in CREATE2 with no constructor arguments
     /// @param creationCode the creation code of a contract C, as returned by type(C).creationCode
-    function hashInitCodeNoConstructorArgs(bytes memory creationCode) public pure returns (bytes32) {
+    function hashInitCode(bytes memory creationCode) internal pure returns (bytes32) {
         return hashInitCode(creationCode, "");
     }
 
     /// @dev returns the hash of the init code (creation code + ABI-encoded args) used in CREATE2
     /// @param creationCode the creation code of a contract C, as returned by type(C).creationCode
     /// @param args the ABI-encoded arguments to the constructor of C
-    function hashInitCode(bytes memory creationCode, bytes memory args) public pure returns (bytes32) {
+    function hashInitCode(bytes memory creationCode, bytes memory args) internal pure returns (bytes32) {
         return keccak256(abi.encodePacked(creationCode, args));
     }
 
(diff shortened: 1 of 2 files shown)


karmacoma-eth:
upstreaming these little helper methods I found useful in our internal deploy scripts (https://github.com/showtime-xyz/showtime-contracts/blob/main/script/utils/Create2Helper.sol)

cc @mds1

src/StdUtils.sol
```diff
@@ -19,6 +19,7 @@ abstract contract StdUtils {
         57896044618658097711785492504343953926634992332820282019728792003956564819968;
     uint256 private constant UINT256_MAX =
         115792089237316195423570985008687907853269984665640564039457584007913129639935;
+    address private constant DEFAULT_CREATE2_DEPLOYER = 0x4e59b44847b379578588920cA78FbF26c0B4956C;
```
mds1, line 22: Nit, maybe we call this `CREATE2_FACTORY` to match the version in `Base.sol`, and we can add a comment explaining this is the default one used by forge?

https://github.com/foundry-rs/forge-std/blob/066ff16c5c03e6f931cd041fd366bc4be1fae82a/src/Base.sol#L30

src/StdUtils.sol
```diff
@@ -116,6 +117,24 @@ abstract contract StdUtils {
         return addressFromLast20Bytes(keccak256(abi.encodePacked(bytes1(0xff), deployer, salt, initcodeHash)));
     }
 
+    /// @dev returns the address of a contract created with CREATE2 using the default CREATE2 deployer
+    function computeCreate2Address(bytes32 salt, bytes32 _initCodeHash) internal pure returns (address) {
+        return computeCreate2Address(salt, _initCodeHash, DEFAULT_CREATE2_DEPLOYER);
```
mds1, line 122: Just removing the underscore prefix for consistency
```suggestion
    function computeCreate2Address(bytes32 salt, bytes32 initCodeHash) internal pure returns (address) {
        return computeCreate2Address(salt, initCodeHash, DEFAULT_CREATE2_DEPLOYER);
```

src/StdUtils.sol
```diff
@@ -116,6 +117,24 @@ abstract contract StdUtils {
...
+    /// @dev returns the hash of the init code (creation code + no args) used in CREATE2 with no constructor arguments
+    /// @param creationCode the creation code of a contract C, as returned by type(C).creationCode
+    function hashInitCodeNoConstructorArgs(bytes memory creationCode) public pure returns (bytes32) {
+        return hashInitCode(creationCode, "");
+    }
+
+    /// @dev returns the hash of the init code (creation code + ABI-encoded args) used in CREATE2
+    /// @param creationCode the creation code of a contract C, as returned by type(C).creationCode
+    /// @param args the ABI-encoded arguments to the constructor of C
+    function hashInitCode(bytes memory creationCode, bytes memory args) public pure returns (bytes32) {
```
mds1, line 134: These methods should all be `internal`, a few of them are `public`

src/StdUtils.sol
```diff
@@ -116,6 +117,24 @@ abstract contract StdUtils {
...
     }
 
+    /// @dev returns the address of a contract created with CREATE2 using the default CREATE2 deployer
+    function computeCreate2Address(bytes32 salt, bytes32 _initCodeHash) internal pure returns (address) {
+        return computeCreate2Address(salt, _initCodeHash, DEFAULT_CREATE2_DEPLOYER);
+    }
+
+    /// @dev returns the hash of the init code (creation code + no args) used in CREATE2 with no constructor arguments
+    /// @param creationCode the creation code of a contract C, as returned by type(C).creationCode
+    function hashInitCodeNoConstructorArgs(bytes memory creationCode) public pure returns (bytes32) {
```
mds1, line 127: Thoughts on just calling this `hashInitCode` and overloading that method? Main concern of that change is that it may lead to confusion for users who aren't too familiar with how to get the initcode hash for the contracts, whereas the current name is more explicit, but that could be handled by docs in the foundry book 🤔
```suggestion
    function hashInitCode(bytes memory creationCode) public pure returns (bytes32) {
```

mds1 (commented):
Thanks for this! Just a few nits on naming and stuff, curious to get your thoughts

src/StdUtils.sol
karmacoma-eth: makes sense, it was indeed on the wordy side

karmacoma-eth:
> Thanks for this! Just a few nits on naming and stuff, curious to get your thoughts

The suggestions make sense, thanks for the fresh pair of eyes!

karmacoma-eth:
suggestions implemented in [d5255d0](https://github.com/foundry-rs/forge-std/pull/276/commits/d5255d05ec20add56e1154cd2d673d8255181c17)

karmacoma-eth:
Side note: note sure what's up with the tests taking 46m in one of the CI runs? I was able to repro locally, seems like tests were running normally and then forge would hang out at 0% CPU with all threads waiting. 

I can't repro anymore and neither does CI, it suggests something related to the network?

mds1:
Yea seems related to fork tests, since subsequent runs are a lot faster by pulling from the cache instead. We have a lot of public RPCs in the CI to test the default RPCs in `StdChains`, so guessing they get throttled and it's waiting to retry requests. Not certain though, will need to look into it

mds1 (approved):
LGTM, thanks!

PR #277, opened 2023-01-21
refactor: rename "who" arg to "msgSender"

PaulRBerg:
Changing the `who` arg to `msgSender` would keep the util/ cheat signatures consistent between `StdUtils.sol` and `Vm.sol` (as per the changes introduced in #246).


mds1 (approved):
lgtm!

PR #278, opened 2023-01-21
Just a minor change to stay consistent with the helpful comment

FredCoen:
Just a minor change to stay consistent with the helpful comment, groups user called and foundry called functions together


mds1 (approved):
Nice catch, thanks!

PR #290, opened 2023-02-02
Rename `StdAssertions.sol` logs

Sabnock01:
This PR changes the `Expected` and `Actual` logs for `assertEq` functions in `StdAssertions.sol` to that of `Left` and `Right`, respectively. 

This is done to match the [assert_eq](https://doc.rust-lang.org/std/macro.assert_eq.html) macro in Rust and improve readability.

### DIFF
--- a/src/StdAssertions.sol
+++ b/src/StdAssertions.sol
@@ -28,8 +28,8 @@ abstract contract StdAssertions is DSTest {
     function assertEq(bool a, bool b) internal virtual {
         if (a != b) {
             emit log("Error: a == b not satisfied [bool]");
-            emit log_named_string("  Expected", b ? "true" : "false");
-            emit log_named_string("    Actual", a ? "true" : "false");
+            emit log_named_string("      Left", a ? "true" : "false");
+            emit log_named_string("     Right", b ? "true" : "false");
             fail();
         }
     }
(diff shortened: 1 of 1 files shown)


mds1:
Thanks @Sabnock01!

@d-xo would you accept a similar PR to ds-test upstream to keep the assertion logs consistent?

d-xo:
Hey, sorry for the delay.

Yes I would, think this is much nicer than what we have now actually :)

mds1:
@Sabnock01 Do you mind submitting the upstream PR to https://github.com/dapphub/ds-test? The once that's merged we can bump the submodule version in this PR

(cc @mattsse @gakonst for awareness in case this breaks any tests upstream in the main foundry repo)

Sabnock01:
@mds1 upstream PR to ds-test has been merged by d-xo. We're good to proceed with this whenever.

gakonst (approved):
Thx for hunting this down!

mds1:
Two things left before we can release:
1. There's still a few expected/actual changes in `StdAssertions` that were missed in this PR that we'll need to do in a follow up PR. I think the only ones are in the two `assertApproxEqAbsDecimal` methods and the two `assertApproxEqRelDecimal` methods, and a few comments still say expected/actual instead of left/right
2. Need to bump ds-test submodule

If anyone wants to tackle those in another PR that would be great, otherwise I should be able to get to it later this week

Sabnock01:
> Two things left before we can release:
> 
>     1. There's still a few expected/actual changes in `StdAssertions` that were missed in this PR that we'll need to do in a follow up PR. I think the only ones are in the two `assertApproxEqAbsDecimal` methods and the two `assertApproxEqRelDecimal` methods, and a few comments still say expected/actual instead of left/right
> 
>     2. Need to bump ds-test submodule
> 
> 
> If anyone wants to tackle those in another PR that would be great, otherwise I should be able to get to it later this week

I'll take care of it sometime today.

PR #291, opened 2023-02-03
Add flag disable default rpc

hexonaut:
Added a flag to disable falling back to default RPC URLs automatically. Public RPC endpoints are often rate-limited and may result in obscure errors when rate limits are hit. On our team we want to disable this behavior and instead spit out a very clear error that the user does not have the proper environment variable configured.

I've opened it as a draft because the new test is failing because the error is not matching even though I copied it exactly from the console. Is there hidden whitespace or an extra character or something? I don't have time to dig into this atm, so thought I'd submit most of the work.

Ping: @mds1

### DIFF
--- a/src/StdChains.sol
+++ b/src/StdChains.sol
@@ -67,7 +67,7 @@ abstract contract StdChains {
     // Maps from a chain ID to it's alias.
     mapping(uint256 => string) private idToAlias;
 
-    bool private useDefaultRpcUrls = true;
+    bool private fallbackToDefaultRpcUrls = true;
 
     // The RPC URL will be fetched from config or defaultRpcUrls if possible.
     function getChain(string memory chainAlias) internal virtual returns (Chain memory chain) {
@@ -158,7 +158,7 @@ abstract contract StdChains {
                 chain.rpcUrl = configRpcUrl;
             } catch (bytes memory err) {
                 string memory envName = string(abi.encodePacked(_toUpper(chainAlias), "_RPC_URL"));
-                if (useDefaultRpcUrls) {
+                if (fallbackToDefaultRpcUrls) {
                     chain.rpcUrl = vm.envOr(envName, defaultRpcUrls[chainAlias]);
                 } else {
                     chain.rpcUrl = vm.envString(envName);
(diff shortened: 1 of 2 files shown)


src/StdChains.sol
```diff
@@ -67,6 +67,8 @@ abstract contract StdChains {
     // Maps from a chain ID to it's alias.
     mapping(uint256 => string) private idToAlias;
 
+    bool private useDefaultRpcUrls = true;
```
mds1, line 70:
```suggestion
    bool private fallbackToDefaultRpcUrls = true;
```

src/StdChains.sol
```diff
@@ -171,6 +177,10 @@ abstract contract StdChains {
         return chain;
     }
 
+    function setUseDefaultRpcUrls(bool useDefault) internal {
```
mds1, line 180: Similarly 
```suggestion
    function setFallbackToDefaultRpcUrls(bool useDefault) internal {
```
(this file has a lot going on so just trying to lean towards keeping things explicit 😅)

src/StdChains.sol
hexonaut: changed

src/StdChains.sol
hexonaut: changed

test/StdChains.t.sol
```diff
@@ -144,4 +144,17 @@ contract StdChainsTest is Test {
         assertEq(modifiedChain.chainId, 999999999);
         assertEq(modifiedChain.rpcUrl, "https://modified.chain/");
     }
+
+    function testDontUseDefaultRpcUrl() public {
+        // Should error if default RPCs flag is set to false.
+        setUseDefaultRpcUrls(false);
+        vm.expectRevert(
```
mds1, line 151: @mattsse any idea why this test fails? the error messages appear identical, and it seems there's no hidden characters

mds1:
Thanks! This lgtm, will merge once we get the test issues resolved

mds1 (approved):
The new `testDontUseDefaultRpcUrl` test now passes, the CI failure is unrelated, from a flaky RPC call

PR #292, opened 2023-02-03
refactor(breaking): rename InvariantTest to StdInvariant and have Test inherit from it by default

### LINKED ISSUE #282: "InvariantTest" does not have a "vm"
No cheatcodes can currently be used in a test contract that inherits just from [`InvariantTest`](https://github.com/foundry-rs/forge-std/blob/0aa99eb8456693c015350c5e6c4f442ebe912f77/src/InvariantTest.sol).

I don't know if this is intended - but if it is, we should definitely mention what's the recommend way of adding the `vm` (in https://github.com/foundry-rs/book/pull/760), e.g. inherit from `Test`, or `TestBase`, define a `vm` yourself, etc.

mds1:
Closes #282

### DIFF
--- a/src/StdInvariant.sol
+++ b/src/StdInvariant.sol
@@ -3,7 +3,7 @@ pragma solidity >=0.6.2 <0.9.0;
 
 pragma experimental ABIEncoderV2;
 
-contract InvariantTest {
+contract StdInvariant {
     struct FuzzSelector {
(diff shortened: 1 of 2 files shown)


Sabnock01 (approved):
👍

gakonst (approved):

PR #294, opened 2023-02-06
chore: use coercion cheatcodes with stdJson helpers

### LINKED ISSUE #288: chore: update `StdJson.sol` to use the specific JSON parsing cheats
For example, the current `readUint` helper is like this:

```solidity
function readUint(string memory json, string memory key) internal pure returns (uint256) {
    return abi.decode(vm.parseJson(json, key), (uint256));
 }
```

But I think it should be:

```solidity
function readUint(string memory json, string memory key) internal pure returns (uint256) {
    return vm.parseJsonUint(json, key);
 }
```

@odyslam can you confirm that we should make this change / implement it if you have time?

odyslam:
Fixes #288

### DIFF
--- a/src/StdJson.sol
+++ b/src/StdJson.sol
@@ -69,23 +69,23 @@ library stdJson {
         return vm.parseJsonAddress(json, key);
     }
 
-    function readAddressArray(string memory json, string memory key) internal  returns (address[] memory) {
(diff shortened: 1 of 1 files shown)


mds1 (approved):
Thank you!

PR #295, opened 2023-02-06
Add Color Support to Console Logging

wirew0lf:
I've included a shortcut to be able to use console logging with colors. The syntax is the same to console.log() but instead uses console.color("color", ...) with the first argument being one of the following colors:

- #ff0000 `red`
- #00ff00 `green`
- #ffff00 `yellow`
- #0000ff `blue`
- #ff00ff `magenta`
- #00ffff `cyan`

Only downside is console.log supports a maximum of 32 bytes, the color codes consume part of this bytes, I'm currently looking for the best way to solve this, but meanwhile I think it would be benefitial to include this. Specially useful when preparing PoCs for demos and clients.


wirew0lf:
After some suggestions I've decided to do some rework on the console script:

- Changes were made to console2.sol as suggested and console.sol was reverted to original state.
- I've modified the library to use a similar pattern to the one used in StdStorage improving the syntax.

# New Syntax
In order to use this new library users first need to specify the `console2` library being used for the `Console` struct.
`using console2 for Console;`

Then the syntax will be `console.red().log();`. Where red  can be any one of the mentioned colors above. To use the standard color you can use `console.normal().log();`

Also it is worth noting that when you specify a color, if you just keep using console.log() without specifying any color the logs will be printed in the last specified color. For a better reference check the code and image below.

```solidity
import "forge-std/Test.sol";

contract ConsoleTest is Test {
    using console2 for Console;
    function testConsole() public {
        console.red().log("Test");
        console.log("This will still be red");        
        console.green().log(uint(1000));
        console.yellow().log(address(0));
        console.magenta().log(true);
        console.cyan().log("Test");
        console.normal().log("Normal Log");
    }
}
```

image

To bypass all this color stuff and just use the classical functions you can use:
`console2.log();`

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "lib/ds-test"]
 	path = lib/ds-test
 	url = https://github.com/dapphub/ds-test
+[submodule "lib/forge-std"]
```
mds1, line 4: Think this new submodule was added accidentally, let's delete it

src/Base.sol
```diff
@@ -21,6 +22,7 @@ abstract contract CommonBase {
 
     Vm internal constant vm = Vm(VM_ADDRESS);
     StdStorage internal stdstore;
+    Console internal console;
```
mds1, line 25: Hmm so this will technically be a breaking change for anyone who was importing the original `conosle.sol`, since there a collision with that library and this variable. But in solidity, you should never need the original `console.sol` (it's just for hardhat users) and should always prefer `console2`, so I think this is ok. 

It also means you can now always just use `console.log` instead of `console2.log`, but get all the benefits of `console2` which is nice.

Wdyt @gakonst @ZeroEkkusu?

src/console2.sol
```diff
@@ -1543,4 +1590,1373 @@ library console2 {
         _sendLogPayload(abi.encodeWithSignature("log(address,address,address,address)", p0, p1, p2, p3));
     }
 
+    // Log Functions with Color functionality.
+        function log(Console storage self, uint256 p0) internal view {
+        _sendLogPayload(abi.encodeWithSignature("log(string,uint256,string)", self._color, p0, CRESET));
```
mds1, line 1595: Should we reset the color after each `log` call? I suspect that's the normal UX that people would expect, i.e. change behavior:

```solidity
// From this:
console.red().log("Test");
console.log("This will still be red");

// To this:
console.red().log("Test");
console.log("This is back to the normal color");
```

src/console2.sol
```diff
@@ -1543,4 +1590,1373 @@ library console2 {
         _sendLogPayload(abi.encodeWithSignature("log(address,address,address,address)", p0, p1, p2, p3));
     }
 
+    // Log Functions with Color functionality.
```
mds1, line 1593: Were these codegen'd or did you do it manually by copying all the existing log methods and just e.g. multi-cursor editing them? Just curious

src/console2.sol
```diff
@@ -1,6 +1,10 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.4.22 <0.9.0;
 
+struct Console {
+    string _color;
+}
```
mds1, line 6: Wonder if we could also add support for bold/underline/italics? I think @brockelmore often uses escape codes in his console.logs so curious to get his thoughts on this PR too

src/console2.sol
wirew0lf: I originally developed it that way, resetting the color after each call. However it should be noted that due to the way console logging is implemented right now through a static call only 32 bytes of data can be logged for each line.

So the color set and reset will consume part of this 32 bytes. So less data can be logged on a single line. If we do it through 3 static calls so users can log 32 bytes even after setting colors there will be a breakline every time a color is set/reset so this should require some rework on forge/anvil.

However shouldn't be a big deal, so I agree on resetting the color after each call, I can modify it and push a new commit if you want me to do so.

src/console2.sol
wirew0lf: I copied the existing log functions and modified them through a regex

src/console2.sol
wirew0lf: Yeah definitely, wonder on how should I approach this in terms of syntax?

I can think of a few but would like to hear your opinions:

- console.red("bold").log(); -> If no font weight is specified it will use normal.
- console.red().bold().log(); -> Again, if bold() is not called it will just default to normal.
- console.color("red", "bold").log(); -> I personally prefer the ones above but, could be another option.

src/Base.sol
wirew0lf: Yeah, I thought of this when developing, but thought this would be the best approach in terms of user experience. Alternatively, this could be set to another word, eg: print. But I think this will be more confusing for users who are already used to do the console.log

src/console2.sol
mds1: > only 32 bytes of data can be logged for each line...So less data can be logged on a single line. 

Hmm I don't think this is correct? `_sendLogPayload` takes an arbitrary `bytes memory` so logs can be any length. Given that resetting the color after each call seems like it shouldn't be an issue

src/console2.sol
```diff
@@ -1543,4 +1590,1373 @@ library console2 {
         _sendLogPayload(abi.encodeWithSignature("log(address,address,address,address)", p0, p1, p2, p3));
     }
 
+    // Log Functions with Color functionality.
+        function log(Console storage self, uint256 p0) internal view {
```
mds1, line 1594:
```suggestion
    function log(Console storage self, uint256 p0) internal view {
```

mds1:
Discussed with @wirew0lf and decided to close this since there's a few ways to implement with various tradeoffs, some involving upstream changes to forge. We'll follow up in a separate issue to discuss first and make sure we get the best UX here

PR #305, opened 2023-02-21
Added deal cheat for ERC721 and ERC1155

wirew0lf:
I've created the deal cheatcode for ERC721 and ERC1155 tokens.

For ERC721 and it shares the same function signature as ERC20 i've modified the name to dealNft(), this could be merged into a single deal function by checking the interface of the given contract but I'm not sure it is the best option as it would increase gas usage and could lead to more errors for non-standard implementations.

Usage:
```solidity
// ERC721
dealNft(nftAddress, bob, tokenId);

//ERC1155
deal(1155address, bob, id, amount);
deal(1155address, bob, id, amount, true); // Will check if it is ERC1155Supply and adjust totalSupply.
```

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -549,8 +549,10 @@ abstract contract StdCheats is StdCheatsSafe {
         deal(token, to, give, false);
     }
 
-    function deal(address token, address to, uint256 id, uint256 give) internal virtual {
-        deal(token, to, id, give, false);
+    // Set the balance of an account for any ERC1155 token
+    // Use the alternative signature to update `totalSupply`
+    function dealERC1155(address token, address to, uint256 id, uint256 give) internal virtual {
+        dealERC1155(token, to, id, give, false);
     }
 
     function deal(address token, address to, uint256 give, bool adjust) internal virtual {
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -549,6 +549,10 @@ abstract contract StdCheats is StdCheatsSafe {
         deal(token, to, give, false);
     }
 
+    function deal(address token, address to, uint256 id, uint256 give) internal virtual {
+        deal(token, to, id, give, false);
+    }
+
     function deal(address token, address to, uint256 give, bool adjust) internal virtual {
         // get current balance
         (, bytes memory balData) = token.call(abi.encodeWithSelector(0x70a08231, to));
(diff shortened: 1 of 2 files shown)


src/StdCheats.sol
```diff
@@ -569,4 +573,47 @@ abstract contract StdCheats is StdCheatsSafe {
...
+            if (give < prevBal) {
+                totSup -= (prevBal - give);
+            } else {
+                totSup += (give - prevBal);
+            }
+            stdstore.target(token).sig(0xbd85b039).with_key(id).checked_write(totSup);
+        }
+    }
+
+    function dealNft(address token, address to, uint256 id) internal virtual {
```
mds1, line 599: Let's use `dealERC721` and `dealERC1155` as the method names, to make things more clear/explicit. ERC1155 is also an NFT standard so the current naming is ambiguous

src/StdCheats.sol
```diff
@@ -569,4 +573,47 @@ abstract contract StdCheats is StdCheatsSafe {
...
+        (, bytes memory toBalData) = token.call(abi.encodeWithSelector(0x70a08231, to));
+        uint256 toPrevBal = abi.decode(toBalData, (uint256));
+
+        // update balances
+        stdstore.target(token).sig(0x70a08231).with_key(abi.decode(ownerData, (address))).checked_write(--fromPrevBal);
+        stdstore.target(token).sig(0x70a08231).with_key(to).checked_write(++toPrevBal);
+
+        // update owner
+        stdstore.target(token).sig(0x6352211e).with_key(id).checked_write(to);
+    }
```
mds1, line 618: _(arbitrary comment location to start a thread)_

Needs `forge fmt`, also let's add some tests

mds1 (commented):
Thanks, this is handy!

src/StdCheats.sol
wirew0lf: Done! Changed method names to `dealERC721` and `dealERC1155`

src/StdCheats.sol
wirew0lf: Reformatted to match `forge fmt` and added tests for both methods.

src/StdCheats.sol
```diff
@@ -569,4 +575,50 @@ abstract contract StdCheats is StdCheatsSafe {
             stdstore.target(token).sig(0x18160ddd).checked_write(totSup);
         }
     }
+
+    function dealERC1155(address token, address to, uint256 id, uint256 give, bool adjust) internal virtual {
+        // get current balance
+        (, bytes memory balData) = token.call(abi.encodeWithSelector(0x00fdd58e, to, id));
+        uint256 prevBal = abi.decode(balData, (uint256));
```
mds1, line 582: Should we check if the call was successful here, like we do in `dealERC721`?

test/StdCheats.t.sol
```diff
@@ -103,6 +103,35 @@ contract StdCheatsTest is Test {
...
     }
 
+    function testDealERC1155Token() public {
+        BarERC1155 barToken = new BarERC1155();
+        address bar = address(barToken);
+        dealERC1155(bar, address(this), 0, 10000e18, false);
+        assertEq(barToken.balanceOf(address(this), 0), 10000e18);
+    }
+
+    function testDealERC1155TokenAdjustTS() public {
```
mds1, line 113:
```suggestion
    function testDealERC1155TokenAdjustTotalSupply() public {
```

wirew0lf: Should we modify this for the `testDealTokenAdjustTS()` function for the ERC20 deal cheatcode too @mds1 ?
https://github.com/foundry-rs/forge-std/blob/65761f05868d5dcea3e4b076da58c1c3f447940f/test/StdCheats.t.sol#L95

mds1: Yea that would be ideal IMO, just more clear/explicit that way. Sorry didn't realize that's where you got the name from!

src/StdCheats.sol
mds1: Got it, appreciate the explanation! Makes sense, I think it's ok to leave as-is then

test/StdCheats.t.sol
mds1: Ok, I just pushed this last change

PR #307, opened 2023-02-21
Bump ds-test and complete assertion log renaming

Sabnock01:
Finishes #290


mds1 (approved):
Thanks! CI failure unrelated

PR #310, opened 2023-02-23
feat(Vm): Add new `expectCall` variants

clabby:
# Overview

Sister PR to https://github.com/foundry-rs/foundry/pull/4413

Adds two new variants of the expectCall cheatcode that allows for asserting the amount of gas passed to a call.

```solidity
// Expect a call to an address with the specified msg.value, gas, and calldata.
function expectCall(address callee, uint256 msgValue, uint64 gas, bytes calldata data) external;

// Expect a call to an address with the specified msg.value and calldata, and a *minimum* amount of gas.
function expectCallMinGas(address callee, uint256 msgValue, uint64 minGas, bytes calldata data) external;
```


mds1 (approved):
Thanks!

PR #311, opened 2023-02-24
feat(StdAssertions): Add `assertEqCall`

0xPhaze:
Adds new assertions to StdAssertions.sol. These are useful for differential fuzz-testing calls.
Expects two calls to either return the same data or revert in the same way (if `strictRevertData = true`).

```solidity
function assertEqCall(
    address targetA,
    bytes memory callDataA,
    address targetB,
    bytes memory callDataB,
    bool strictRevertData
) internal virtual;
```

### DIFF
--- a/src/StdAssertions.sol
+++ b/src/StdAssertions.sol
@@ -322,4 +322,55 @@ abstract contract StdAssertions is DSTest {
             assertApproxEqRelDecimal(a, b, maxPercentDelta, decimals);
         }
     }
+
+    function assertEqCall(address target, bytes memory callDataA, bytes memory callDataB) internal virtual {
+        assertEqCall(target, callDataA, target, callDataB, true);
+    }
+
+    function assertEqCall(address targetA, bytes memory callDataA, address targetB, bytes memory callDataB)
+        internal
+        virtual
+    {
+        assertEqCall(targetA, callDataA, targetB, callDataB, true);
+    }
+
+    function assertEqCall(address target, bytes memory callDataA, bytes memory callDataB, bool strictRevertData)
+        internal
+        virtual
+    {
+        assertEqCall(target, callDataA, target, callDataB, strictRevertData);
+    }
+
+    function assertEqCall(
+        address targetA,
+        bytes memory callDataA,
+        address targetB,
+        bytes memory callDataB,
+        bool strictRevertData
+    ) internal virtual {
+        (bool successA, bytes memory returnDataA) = address(targetA).call(callDataA);
+        (bool successB, bytes memory returnDataB) = address(targetB).call(callDataB);
+
+        if (successA && successB) {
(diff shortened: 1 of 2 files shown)


test/StdAssertions.t.sol
```diff
@@ -607,6 +613,94 @@ contract StdAssertionsTest is Test {
...
+        bytes memory callDataB,
+        bytes memory revertDataA,
+        bytes memory revertDataB
+    ) external {
+        vm.assume(keccak256(revertDataA) != keccak256(revertDataB));
+
+        address targetA = address(new TestMockCall(revertDataA, SHOULD_REVERT));
+        address targetB = address(new TestMockCall(revertDataB, SHOULD_REVERT));
+
+        vm.expectEmit(false, false, false, true);
```
mds1, line 675: Any reason not to just make all of these params `true`? Typically I like defaulting to that unless there's reason you explicitly want to avoid testing a topic (using true even if there's no topic is ok)

src/StdAssertions.sol
```diff
@@ -322,4 +322,55 @@ abstract contract StdAssertions is DSTest {
...
+        if (successA && successB) {
+            assertEq(returnDataA, returnDataB, "Call return data does not match");
+        }
+
+        if (!successA && !successB && strictRevertData) {
+            assertEq(returnDataA, returnDataB, "Call revert data does not match");
+        }
+
+        if (!successA && successB) {
+            emit log("Error: Call reverted unexpectedly");
```
mds1, line 363: Maybe we change this and the branch below to
```suggestion
            emit log("Error: Calls were not equal");
```

The current message implies that A was intended to succeed because B did, but I don't think that's true (i.e. maybe they both should have reverted).

src/StdAssertions.sol
```diff
@@ -322,4 +322,55 @@ abstract contract StdAssertions is DSTest {
...
+            emit log_named_bytes("  Left call revert data", returnDataA);
+            emit log_named_bytes(" Right call return data", returnDataB);
+            fail();
+        }
+
+        if (successA && !successB) {
+            emit log("Error: Call did not revert");
```
mds1, line 370:
```suggestion
            emit log("Error: Calls were not equal");
```

mds1:
A few small comments, otherwise this looks great!

src/StdAssertions.sol
0xPhaze: I had first followed the convention assertEq(testValue, referenceValue), but then saw that forge-std changed this to Left/Right.

test/StdAssertions.t.sol
0xPhaze: Just copied it from some other section.

mds1 (approved):
Thanks @0xPhaze!

PR #313, opened 2023-02-27
Added StdStyle library to print logs with different colors and styles

wirew0lf:
After discussing with @mds1 and thanks to his help I've created an "styles" library that allows to use `console.log()` with different styles. As intended in #295.

### Syntax
Syntax is `StdStyle.[style_name]("Message")`. This must be used inside the standard console2.log(), and can be used with **string, uint256, int256, boolean and address** types. Moreover, they **can be nested** to apply multiple styles. Eg:
```solidity
  console2.log(StdStyle.red("Red Bold String Test"));
  console2.log(StdStyle.green(StdStyle.dim(uint256(10e18))));
  console2.log(StdStyle.yellow(StdStyle.bold(int256(-10e18))));
  console2.log(StdStyle.blue(StdStyle.underline(address(0))));
  console2.log(StdStyle.magenta(StdStyle.reverse(true)));
```
image

#### Colors
- #ff0000 `red`
- #00ff00 `green`
- #ffff00 `yellow`
- #0000ff `blue`
- #ff00ff `magenta`
- #00ffff `cyan`
image

#### Font
- bold
- dim
- italic
- underline
- reverse
image

### Custom Styles
If someone plans to use some "pre-defined" styles i.e: headers, errors, etc . Writing the full StdStyle.style syntax with nested styles multiple times can become tedious. To make this easier users can define custom styles in their test files with simple internal functions eg:

```solidity
    function testStyleCustom() public {
        console2.log(h1("Custom Style 1"));
        console2.log(h2("Custom Style 2"));
        console2.log(warning("Warning!"));
    }

    function h1(string memory a) private view returns (string memory) {
        return StdStyle.cyan(StdStyle.reverse(StdStyle.bold(a)));
    }

    function h2(string memory a) private view returns (string memory) {
        return StdStyle.magenta(StdStyle.bold(StdStyle.underline(a)));
    }

    function warning(string memory a) private view returns (string memory) {
        return StdStyle.yellow(StdStyle.bold(a));
    }
```
image

### DIFF
--- a/src/StdStyle.sol
+++ b/src/StdStyle.sol
@@ -24,19 +24,19 @@ library StdStyle {
     }
 
     function red(uint256 self) internal view returns (string memory) {
-        return string.concat(RED, vm.toString(self), RESET);
+        return red(vm.toString(self));
     }
 
     function red(int256 self) internal view returns (string memory) {
-        return string.concat(RED, vm.toString(self), RESET);
+        return red(vm.toString(self));
     }
 
     function red(address self) internal view returns (string memory) {
-        return string.concat(RED, vm.toString(self), RESET);
+        return red(vm.toString(self));
     }
 
     function red(bool self) internal view returns (string memory) {
-        return string.concat(RED, vm.toString(self), RESET);
+        return red(vm.toString(self));
     }
 
     function green(string memory self) internal view returns (string memory) {
@@ -44,19 +44,19 @@ library StdStyle {
     }
 
     function green(uint256 self) internal view returns (string memory) {
-        return string.concat(GREEN, vm.toString(self), RESET);
+        return green(vm.toString(self));
     }
 
     function green(int256 self) internal view returns (string memory) {
-        return string.concat(GREEN, vm.toString(self), RESET);
+        return green(vm.toString(self));
     }
 
     function green(address self) internal view returns (string memory) {
-        return string.concat(GREEN, vm.toString(self), RESET);
+        return green(vm.toString(self));
     }
 
     function green(bool self) internal view returns (string memory) {
-        return string.concat(GREEN, vm.toString(self), RESET);
+        return green(vm.toString(self));
     }
 
     function yellow(string memory self) internal view returns (string memory) {
@@ -64,19 +64,19 @@ library StdStyle {
     }
 
     function yellow(uint256 self) internal view returns (string memory) {
-        return string.concat(YELLOW, vm.toString(self), RESET);
+        return yellow(vm.toString(self));
     }
 
     function yellow(int256 self) internal view returns (string memory) {
-        return string.concat(YELLOW, vm.toString(self), RESET);
+        return yellow(vm.toString(self));
     }
 
     function yellow(address self) internal view returns (string memory) {
-        return string.concat(YELLOW, vm.toString(self), RESET);
+        return yellow(vm.toString(self));
     }
 
     function yellow(bool self) internal view returns (string memory) {
-        return string.concat(YELLOW, vm.toString(self), RESET);
+        return yellow(vm.toString(self));
     }
 
     function blue(string memory self) internal view returns (string memory) {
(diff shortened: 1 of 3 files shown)


src/StdStyle.sol
```diff
@@ -0,0 +1,241 @@
...
+    string constant UNDERLINE = "\u001b[4m";
+    string constant REVERSE = "\u001b[7m";
+    string constant RESET = "\u001b[0m";
+
+    function red(string memory self) internal view returns (string memory) {
+        return string.concat(RED, self, RESET);
+    }
+
+    function red(uint256 self) internal view returns (string memory) {
+        return string.concat(RED, vm.toString(self), RESET);
```
brockelmore, line 27: Why dont you just do:
```solidity
    function red(string memory self) internal view returns (string memory) {
        return string.concat(RED, self, RESET);
    }

    function red(uint256 self) internal view returns (string memory) {
        return red(vm.toString(self));
    }
```

That way in the future we only ever have to change the string implementation if we every want to change the logic, and all the others are just a conversion to string then the `red(string)` function

wirew0lf: Yeah, this is a better approach thanks for the suggestion @brockelmore. I've changed this in the last commit.

src/StdStyle.sol
```diff
@@ -0,0 +1,241 @@
...
+
+    function red(int256 self) internal view returns (string memory) {
+        return red(vm.toString(self));
+    }
+
+    function red(address self) internal view returns (string memory) {
+        return red(vm.toString(self));
+    }
+
+    function red(bool self) internal view returns (string memory) {
```
mds1, line 38: Seems we're missing `bytes32` and `bytes memory` overloads for each color/format, could we add those?

src/StdStyle.sol
```diff
@@ -0,0 +1,241 @@
...
+    string constant GREEN = "\u001b[92m";
+    string constant YELLOW = "\u001b[93m";
+    string constant BLUE = "\u001b[94m";
+    string constant MAGENTA = "\u001b[95m";
+    string constant CYAN = "\u001b[96m";
+    string constant BOLD = "\u001b[1m";
+    string constant DIM = "\u001b[2m";
+    string constant ITALIC = "\u001b[3m";
+    string constant UNDERLINE = "\u001b[4m";
+    string constant REVERSE = "\u001b[7m";
```
mds1, line 19: Would `INVERT` or `INVERSE` be a better name for this? IIUC it inverts the colors

src/StdStyle.sol
wirew0lf: This is intentional, I originally added `bytes32` and `bytes memory` too. However, if you then try to pass as string as literal i.e: `red("Test");` the compiler is not able to differentiate which signature to use and prints an error stating there are multiple candidates. Solution would be to use `red(string("Test"));` but I think this will be more tedious as user will use more often a string than bytes types + users can do `red(vm.toString(bytes32));`

However, other solution to overcome this could be `redBytes32()`and `redBytes()` wdty @mds1?

src/StdStyle.sol
wirew0lf: I thought of this, but I decided to follow the standard naming for ANSI/VT100:
- https://misc.flogisoft.com/bash/tip_colors_and_formatting
- https://www2.ccs.neu.edu/research/gpc/VonaUtils/vona/terminal/vtansi.htm

However I agree that `invert`or `inverse` could be better options as `reverse` may be confused with the text being reversed.

src/StdStyle.sol
mds1: Ah true, let's go with `redBytes32` and `redBytes` for each, since that matches e.g. `console.logBytes` and `console.logBytes32`

wirew0lf: Added `Bytes` and `Bytes32` types for each style

mds1:
Looks great, thanks!

lucas-manuel:
Just found this, nice 😎

Issue #252, opened 2022-12-06
Feature request: "expectEmit" cheatcode with no arguments


PaulRBerg:
From the Foundry Book on [best practices](https://book.getfoundry.sh/tutorials/best-practices):

> When testing events, prefer setting all `expectEmit` arguments to `true`, i.e. `vm.expectEmit(true, true, true, true)`.

My feature request is to add an `expectEmit` cheatcode that has no arguments. All of the topics (`checkTopic1`, `checkTopic2`, `checkTopic3`, and `checkData`) would be set to `true` by default.

ZeroEkkusu:
There was `emission` in https://github.com/foundry-rs/forge-std/pull/42.

mds1:
This idea has been discussed / gotten support in the foundry telegram a few times, so per https://github.com/foundry-rs/foundry/issues/3782 I'd support adding that cheat directly into foundry if you wanted to open the issue/PR there

mds1:
Closed by #320

PR #320, opened 2023-03-03
feat: add simple "expectEmit"

PaulRBerg:
Implements https://github.com/foundry-rs/forge-std/issues/252.

I know that this has been previously discussed in https://github.com/foundry-rs/forge-std/pull/42, but I think that this abstraction would be so helpful that it's worth reconsidering it.

FWIW: I have this function defined in like ~7 separate repos.

### DIFF
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -326,12 +326,22 @@ interface Vm is VmSafe {
     function expectRevert(bytes calldata revertData) external;
     function expectRevert(bytes4 revertData) external;
     function expectRevert() external;
+
+    // Prepare an expected log with all four checks enabled.
+    // Call this function, then emit an event, then call a function. Internally after the call, we check if
+    // logs were emitted in the expected order with the expected topics and data.
+    // Second form also checks supplied address against emitting contract.
+    function expectEmit() external;
+    function expectEmit(address) external;
+
     // Prepare an expected log with (bool checkTopic1, bool checkTopic2, bool checkTopic3, bool checkData).
     // Call this function, then emit an event, then call a function. Internally after the call, we check if
-    // logs were emitted in the expected order with the expected topics and data (as specified by the booleans)
+    // logs were emitted in the expected order with the expected topics and data (as specified by the booleans).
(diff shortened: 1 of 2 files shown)


mds1:
Wdyt of PR'ing this as a native cheat? It should be very simple to implement, and per https://github.com/foundry-rs/foundry/issues/3782 it'd be ideal to start moving things there by default when we can

PaulRBerg:
> Wdyt of PR'ing this as a native cheat? 

Done!

https://github.com/foundry-rs/foundry/pull/4479

> and per https://github.com/foundry-rs/foundry/issues/3782 it'd be ideal

I'm in favor of that proposal!

PaulRBerg:
Since I ended up implementing this as a native cheat in Foundry (https://github.com/foundry-rs/foundry/pull/4479), I refactored this branch to add the `expectEmit` cheat to `Vm.sol`, and then I force pushed to this branch.

src/Vm.sol
```diff
@@ -326,12 +326,21 @@ interface Vm is VmSafe {
...
     // Prepare an expected log with (bool checkTopic1, bool checkTopic2, bool checkTopic3, bool checkData).
     // Call this function, then emit an event, then call a function. Internally after the call, we check if
-    // logs were emitted in the expected order with the expected topics and data (as specified by the booleans)
-    function expectEmit(bool checkTopic1, bool checkTopic2, bool checkTopic3, bool checkData) external;
-    function expectEmit(bool checkTopic1, bool checkTopic2, bool checkTopic3, bool checkData, address emitter)
-        external;
+    // logs were emitted in the expected order with the expected topics and data (as specified by the booleans).
+    // Second form also checks supplied address against emitting contract.
+    function expectEmit(bool, bool, bool, bool) external;
+    function expectEmit(bool, bool, bool, bool, address) external;
```
mds1, line 342: Thanks! Will wait until the upstream PR is merged before merging this. Looks like the named parameters got dropped though?

PaulRBerg: Whoops! Will add back shortly.

PR #322, opened 2023-03-09
Allow forge-std to be installed as npm package

ian:
## What's this do? 

Adds support for forge-std to be used through a node package manager, via `npm add forge-std` or equivalent (yarn, pnpm, etc)

## How to install

After this PR is merged: 

`npm add forge-std`

Before this PR is merged:

We've temporarily published the repo as an npm package `forge-std-no-submodules`. To use, just add this to your package.json:

```
"forge-std@npm:forge-std-no-submodules": "latest"
```

No other changes are necessary, importing forge-std should work + look the same. You can nuke your `lib/forge-std` directory and .gitmodules files if you'd like.

## Rationale

Foundry is such an incredible tool for smart contract devs. It's honestly so far beyond the alternatives, I'm not sure what I'd do without it at this point. Kudos to you and team 👏

Many of us find the git submodule only approach rather frustrating, mainly due to the fact that the vast majority of our contract dependencies are being installed via npm or equivalent.

I'd really like to make the pitch for foundry to switch from git submodules to a node package manager, and can furnish many developer opinions to the matter. 

Would you be open to a discussion?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
-[submodule "lib/ds-test"]
-	path = lib/ds-test
-	url = https://github.com/dapphub/ds-test
+[submodule "lib/ds-test"]	
+	path = lib/ds-test	
+	url = https://github.com/dapphub/ds-test	
--- a/package.json
+++ b/package.json
@@ -12,5 +12,8 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/foundry-rs/forge-std.git"
+  },
+  "dependencies": {
+    "ds-test": "https://github.com/dapphub/ds-test"
   }
 }
--- a/src/StdAssertions.sol
+++ b/src/StdAssertions.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.2 <0.9.0;
 
-import {DSTest} from "ds-test/test.sol";
+import {DSTest} from "ds-test/src/test.sol";
 import {stdMath} from "./StdMath.sol";
 
 abstract contract StdAssertions is DSTest {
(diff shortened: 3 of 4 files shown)


mds1:
> Many of us find the git submodule only approach rather frustrating, mainly due to the fact that the vast majority of our contract dependencies are being installed via npm or equivalent.
> 
> I'd really like to make the pitch for foundry to switch from git submodules to a node package manager, and can furnish many developer opinions to the matter.
> 
> Would you be open to a discussion?

Switching to a better dependency model than submodules is being tracked in https://github.com/foundry-rs/foundry/issues/2618, so feel free to make your pitch there (though I think there's talks of using/building a more EVM-specific packaging tool)

--------

How does this PR enable `npm add forge-std` without adding a release process to publish this to npm?

It also seems this PR is broken as it's now looking for DSTest in `lib/ds-test/src/src/test.sol` instead of `lib/ds-test/src/test.sol`😅

Note that you can install forge-std as an npm package currently by specifying the full URL to a commit, like this: https://github.com/ethereum-optimism/optimism/blob/6e8369593b3ff59670cbc7a108d1692a49a595dd/packages/contracts-bedrock/package.json#L82

PaulRBerg:
@ian - if you can afford to switch to [PRBTest](https://github.com/PaulRBerg/prb-test), you will be able to pull the dependecy from [NPM](https://www.npmjs.com/package/@prb/test). I ship every release of PRBTest to the npm registry.

ian:
@PaulRBerg I ended up publishing https://www.npmjs.com/package/forge-std-no-submodules to get around the submodule issues.

mds1:
I'm going to close this PR since it's not clear how this actually resolves things per https://github.com/foundry-rs/forge-std/pull/322#issuecomment-1462841728, and because forge-std can be installed via npm already by passing in the github URL/commit currently. If there are any issues though feel free to open an issue

PR #331, opened 2023-03-22
Add Account helper struct to StdCheats

jameswenzel:
# Motivation

`makeAddrAndKey` is great for creating labeled accounts along with a private key for signing, but when working with multiple accounts, can get messy to track which key belongs to which account.

The `Account` struct contains both an address and its key. It's useful for packaging up information about arbitrary signers to be passed among different test methods.

Extremely basic use-case:

```solidity
function testFulfillOrder() public {
    Account memory offerer = makeAccount('offerer');
    // test setup logic...
    Order memory order = generateOrder(account);
    marketplace.fulfillOrder(order);
}

function generateOrder(Account memory account) internal returns (Order memory order) {
    // order generation logic...
    order.offerer = account.addr;
    order.signature = vm.sign(account.key, orderHash);
}
```

The `Account` struct is most useful when used with logic that requires an arbitrary number of accounts that need to make an arbitrary number of signatures, and assertions that require knowing an Account's address.

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -188,6 +188,11 @@ abstract contract StdCheatsSafe {
         string value;
     }
 
+    struct Account {
+        address addr;
+        uint256 key;
+    }
+
     function assumeNoPrecompiles(address addr) internal virtual {
         // Assembly required since `block.chainid` was introduced in 0.8.0.
         uint256 chainId;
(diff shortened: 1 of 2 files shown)


mds1 (approved):
Makes sense, I like it. Thanks!

PR #336, opened 2023-03-26
fix(StdAssertions): % is not displayed as %

0xdapper:


mds1:
Hmm @OliverNChalk wdyt here? I think you originally PR'd the logic/tests for these methods. Expressing e.g. `25%` as:
- 0.25, like we currently do
- 25%, like this PR changes to

are both valid. I think I lean towards merging this PR since it feels more intuitive/clear, but wanted to see if there's a reason you did it the other way.

Also @0xdapper looks like it needs `forge fmt`

OliverNChalk:
Hmm, I think I originally ported this from solmate. I think this PR looks good/correct from my PoV

mds1 (approved):
Thank you!

PR #339, opened 2023-03-29
Increment DS Test version

0xGuybrush:
PR to increment version of DS test (as suggested: https://github.com/foundry-rs/foundry/issues/4634)

* I've ran a `forge install` against DS Test and bumped to the latest commit on master there
* I've added some test cases into `StdAssertions.t.sol` and tried to match the conventions there
* Ran `forge fmt` and `forge test` (passes).

Please let me know any suggestions/amends.


mds1 (approved):
Thanks!

PR #343, opened 2023-04-08
changes for mapping slot API

clouds56:
corresponding PR https://github.com/foundry-rs/foundry/pull/4710


mds1:
Replaced by https://github.com/foundry-rs/forge-std/pull/449

PR #344, opened 2023-04-09
perf: use unchecked for stdMath.delta(uint256,uint256)

OliverNChalk:
I have a downstream codebase that uses `stdMath.delta`, so this gas golf would be valuable to me. If not wanted I can maintain a copy of this function ofc, just let me know.


brockelmore:
@mds1 Should we consider making all of these virtual functions though so that if a user wants to override they can?

PR #346, opened 2023-04-12
feat(stdCheats): Introduce `grow` and `drop` cheats

pedrommaiaa:
Sometimes, we need to adjust the amount of ether or tokens in an account without overwriting it completely. Currently, the only way to do this is as follows:

```Solidity
deal(address(this), address(this).balance + 1 ether);
deal(address(this), address(this).balance - 1 ether);
```

By using `grow` and `drop`, we can perform the same operation without having to consider the current amount of the specified account.

```Solidity
function testGrow() public {
    grow(address(this), 1 ether);
    grow(address(this), 1 ether);
    assertEq(address(this).balance, 2 ether);
}

function testDrop() public {
    grow(address(this), 2 ether);
    drop(address(this), 1 ether);
    assertEq(address(this).balance, 1 ether);
}
```

Enabling this functionality would enable users to avoid using mock tokens that expose mint and burn functionality, allowing them to create tests that better mimic real-world scenarios.

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -636,4 +636,72 @@ abstract contract StdCheats is StdCheatsSafe {
         // update owner
         stdstore.target(token).sig(0x6352211e).with_key(id).checked_write(to);
     }
+
+    // Increases the ether balance of an account.
+    function grow(address to, uint256 give) internal virtual {
+        vm.deal(to, to.balance + give);
+    }
+
+    // Increases the balance of an account for any ERC20 token, updating `totalSupply`.
+    function grow(address token, address to, uint256 give) internal virtual {
+        // update balance
+        (, bytes memory balData) = token.call(abi.encodeWithSelector(0x70a08231, to));
+        uint256 prevBal = abi.decode(balData, (uint256));
+        stdstore.target(token).sig(0x70a08231).with_key(to).checked_write(prevBal + give);
+
+        // update total supply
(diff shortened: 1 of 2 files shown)


mds1:
Aren't these basically just aliases to prevent a ` + balanceOf` style part of an expression? I'm hesitant to add more one line cheats like that, feels like it adds bloat with minimal value.  But it seems the functions are reimplemented here so maybe I'm misunderstanding something?

If we do implement them, let's come up with more intuitive names, like `dealWithAdjustment` or `adjustBalance` 😅

pedrommaiaa:
> Aren't these basically just aliases to prevent a ` + balanceOf` style part of an expression? I'm hesitant to add more one line cheats like that, feels like it adds bloat with minimal value. But it seems the functions are reimplemented here so maybe I'm misunderstanding something?
> 
> If we do implement them, let's come up with more intuitive names, like `dealWithAdjustment` or `adjustBalance` 😅

Yeah you can think of it as just aliases, what do you think about `dealAdd` and `dealSub`? I was trying to be concise with the names here but it can be up for discussion.

mds1:
> Yeah you can think of it as just aliases

We've been trying to avoid introducing bloat for things that are mainly just aliases for ~1 line helpers, since it affects compilation times, and slow compilation is a common complaint. (There are of course some existing 1 line helpers that have been around for a while). But these are easy enough to just implement locally with minimal effort, so would prefer to leave them out of forge-std. Thanks though! :)

PR #354, opened 2023-04-19
fix: removed unnecessary todo comment

dd0sxx:
Discovered this `todo` comment in `src/StdUtils.sol` that seems out of place. Unless this is here for a reason, I suggest removing it.


mds1 (approved):
Thanks!

PR #366, opened 2023-04-29
ci: sync release branch

PaulRBerg:
Adds a CI workflow for syncing the `v1` branch every time a new git tag that starts with `v1` is pushed to remote.

PR #371, opened 2023-05-05
ci: use "push.tags" webhook event

PaulRBerg:
Makes the workflow added in https://github.com/foundry-rs/forge-std/pull/366 more robust by using the [`push`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push) webhook event instead of [`release`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release), which doesn't work with releases made by other GitHub Actions workflows.

For more information, see https://github.com/anton-yurchenko/git-release/discussions/100.

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -1,14 +1,14 @@
 name: Sync Release Branch
 
 on:
-  release:
-    types:
-      - created
+  push:
+    tags:
+      - "v1.*"
 
 jobs:
   sync-release-branch:
     runs-on: ubuntu-latest
-    if: startsWith(github.event.release.tag_name, 'v1')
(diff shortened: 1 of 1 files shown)


mds1:
Is the `push` event fired when the release+tag is created from the GitHub UI? In that case I never pushed from my machine to the origin, so I don't know if this will trigger. From the docs it's not clear to me

PaulRBerg:
No, it is not fired when you create a GitHub release via the GitHub UI.

https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#push

In this case, I think it would be best to keep it as is, but FYI the workflow will stop working once Forge Std moves to an automated way of generating the releases themselves.

PR #373, opened 2023-05-06
feat(cheatcodes): Add `vm.prevrandao` to `Vm.sol`

Evalir:
Follow up to https://github.com/foundry-rs/foundry/pull/4856

### DIFF
--- a/src/console2.sol
+++ b/src/console2.sol
@@ -9,7 +9,19 @@ pragma solidity >=0.4.22 <0.9.0;
 library console2 {
     address constant CONSOLE_ADDRESS = address(0x000000000000000000636F6e736F6c652e6c6f67);
 
-    function _sendLogPayload(bytes memory payload) private view {
+    function _castLogPayloadViewToPure(
+        function(bytes memory) internal view fnIn
+    ) internal pure returns (function(bytes memory) internal pure fnOut) {
+        assembly {
+            fnOut := fnIn
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


src/Vm.sol
```diff
@@ -352,6 +352,8 @@ interface Vm is VmSafe {
     function fee(uint256 newBasefee) external;
     // Sets block.difficulty
     function difficulty(uint256 newDifficulty) external;
+    // Sets block.prevrandao
+    function prevrandao(bytes32 newPrevrandao) external;
```
mds1, line 356: Are these aliases for each other or will they revert depending on the EVM version you have selected in `foundry.toml`? If the latter, might be nice to document that in the comments here

Evalir: Ah, actually, that's a great point—these will fail silently depending on which EVM version you have. They've not been configured/changed to revert (and REVM does not fail either). We should probably update these to have this behavior

mds1: Sounds good, I'll hold off on merging this in the meantime then if you think that change will be ready soon

Evalir: Following up with https://github.com/foundry-rs/foundry/pull/4904

mds1: With https://github.com/foundry-rs/foundry/pull/4904 merged, @Evalir do you mind pushing a commit to tweak these comments so they're a bit more self-documenting?

Evalir:
Ah shoot closed this accidentally by updating my fork. Will reopen

PR #374, opened 2023-05-07
feat: pure `console.log`s

0xdapper:
h/t @0age https://twitter.com/z0age/status/1654922202930888704

Thought about creating a different console3.sol but there is no downside to just making everything pure.

### DIFF
--- a/src/console2.sol
+++ b/src/console2.sol
@@ -9,7 +9,19 @@ pragma solidity >=0.4.22 <0.9.0;
 library console2 {
     address constant CONSOLE_ADDRESS = address(0x000000000000000000636F6e736F6c652e6c6f67);
 
-    function _sendLogPayload(bytes memory payload) private view {
+    function _castLogPayloadViewToPure(
+        function(bytes memory) internal view fnIn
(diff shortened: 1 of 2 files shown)


mds1:
Awesome, thank you! Was hoping someone would PR this, really clever workaround

I'll take a closer look tomorrow or monday before merging, but can we revert the changes to `console.sol`? That one is left untouched to ensure compatibility with hardhat, whereas `console2.sol` fixes the original selector issues of `console.sol` + we can modify/extend it for forge as needed

0xdapper:
I don't think this should break the compatibility with hardhat

mds1:
I also don't think it should break the compatibility with hardhat, but I haven't tested to confirm, so would prefer to leave it unless there's a reason to change it. Not sure who uses `console.sol` or what assumptions hardhat makes about it. I guess one downside of changing it is that now anyone using `console.sol` in hardhat projects with minimal forge usage now get compiler warnings about function visibility 🤷

Aside from that this PR lgtm

mds1 (approved):
looks great, thank you 🫡

mds1 (approved):
looks great, thank you 🫡

PR #375, opened 2023-05-07
feat: add memory safe logging

Philogy:
Adds a new alternative logging library to `console` and `console2`, dubbed "safeconsole" that behaves almost identically to `console2` but leaves memory intact, making it useful for logging and debugging memory manipulating inline-assembly code. This is achieved by replacing string arguments with `bytes32` allowing you use the same string-syntax e.g. `safeconsole.log("My int: %d", 34);` without affecting memory (as long as your logging string is less than or equal to 32 characters long). The library also uses the [0age function pointer casting trick](https://twitter.com/z0age/status/1654922202930888704) to ensure that all logging functions are `pure` and don't require function visibility changes.

Carried over from my repo [forge-safe-log](https://github.com/Philogy/forge-safe-log) (renamed from "safelog" to avoid redundant sounding "safelog.log" writing). You'll also find the script that generated the library [here](https://github.com/Philogy/forge-safe-log/blob/main/script/generate-log-fns.py).

### DIFF
--- a/src/safeconsole.sol
+++ b/src/safeconsole.sol
@@ -24,8 +24,8 @@ library safeconsole {
     }
 
     function _memcopy(uint256 fromOffset, uint256 toOffset, uint256 length) private pure {
-        function(uint256, uint256, uint) internal view fnIn = _memcopyView;
-        function(uint256, uint256, uint) internal pure pureMemcopy;
+        function(uint256, uint256, uint256) internal view fnIn = _memcopyView;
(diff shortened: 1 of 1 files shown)


Philogy:
Changed the pragma version to fix the incompatibility issue. Pragma is now `>=0.6.2 <0.9.0`.

mds1 (approved):
This is great, thanks!

PR #377, opened 2023-05-11
feat(cheatcodes): Add `vm.prevrandao` to `Vm.sol`

Evalir:
Follow up to https://github.com/foundry-rs/foundry/pull/4856


mds1 (approved):
thanks!

PR #378, opened 2023-05-11
fix(`StdAssertions`): Expect all events to be emitted, without omitting the first one

Evalir:
This change stems from https://github.com/foundry-rs/foundry/pull/4920 — It will fail if we ignore the first `emit` of these assertions, as it expects to **strictly** match _all_ events performed in the next call and its subtree.

### DIFF
--- a/test/StdAssertions.t.sol
+++ b/test/StdAssertions.t.sol
@@ -686,11 +686,15 @@ contract StdAssertionsTest is Test {
         address targetB = address(new TestMockCall(returnDataB, SHOULD_REVERT));
 
         vm.expectEmit(true, true, true, true);
+        emit log("Error: Calls were not equal");
+        vm.expectEmit(true, true, true, true);
         emit log_named_bytes("  Left call return data", returnDataA);
         vm.expectEmit(true, true, true, true);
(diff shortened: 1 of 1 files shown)


mds1:
Sorry not sure I totally follow this change—it sounds like we'll need to `expectEmit` for _all_ events emitted in a call and it's subcalls now, and you can't only expect e.g. the third event?

Evalir:
Exactly, yes

mds1:
Hmm IIUC isn't this a big devex burden then? For example, let's say I have a fork test where the contract I'm developing locally, `MyContract` will execute a trade on uniswap and will `emit MyContractSwapWorked()` on a successful swap. In my tests, I can no longer just `vm.expectEmit()` for `emit MyContractSwapWorked()` and instead need to also expect all token transfer events, etc?

Evalir:
Yeah you're right. I think that then we should make it so that what you're saying is possible—shouldn't be a big change nevertheless.

mds1:
Awesome, yea that would be much better IMO. Thanks!

Evalir:
Closing—not needed anymore

PR #379, opened 2023-05-15
fix(tests): Conform to upcoming `expectRevert` behavior

Evalir:
Followup to https://github.com/foundry-rs/foundry/pull/4945

The new expectRevert behavior means that it will only check the next call, instead of checking a revert at the "test" level. Some of the tests in `forge-std` expected the latter behavior, but with the linked PR this will change. This updates the tests to comply to the new behavior.


test/StdChains.t.sol
```diff
@@ -3,6 +3,24 @@ pragma solidity >=0.7.0 <0.9.0;
 
 import "../src/Test.sol";
 
+contract StdChainsMock is Test {
+    function getChain_(string memory chainAlias) public returns (Chain memory) {
```
mds1, line 7: wdyt about `exposed_getChain`, etc for all methods? that's what we have recommended in the forge docs for tests of this style

Evalir: sounds good! will update here and in forge-std

Evalir:
Should be good to go now! @mds1

mds1 (approved):
Sorry missed that this was ready, thanks @Evalir! Looks like CI is passing—am I correct that this can be merged now (without waiting for v1 release) since this fix is compatible both with current and future (v1) expectRevert behavior?

Evalir:
Yep indeed can be merged now!

mds1 (approved):
Thank you!

PR #382, opened 2023-05-19
feat: add "boundPrivateKey" util

PaulRBerg:
This PR adds a utility for bounding private keys between 1 and the SECP256k1 order minus 1.

As shown by the discussions in https://github.com/foundry-rs/foundry/issues/3031, it is not obvious that private keys have to be bounded between 1 and the curve order minus 1.

Note: I used the console log-free `_bound` due to security reasons (avoiding accidental leaks of private keys).


mds1:
Looks like the tests are failing since it can't access the `SECP256K1_ORDER` var, otherwise lgtm!

PaulRBerg:
Whoops! Fixed now.

Thanks for the quick review.

mds1 (approved):
Thanks!

PR #383, opened 2023-05-21
perf: use staticcall instead of call

PaulRBerg:
Unless I'm missing something, `staticcall` should be used instead of `call` for querying the token balances.

Cc @brockelmore.

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -572,7 +572,7 @@ abstract contract StdCheats is StdCheatsSafe {
 
     function deal(address token, address to, uint256 give, bool adjust) internal virtual {
         // get current balance
-        (, bytes memory balData) = token.call(abi.encodeWithSelector(0x70a08231, to));
+        (, bytes memory balData) = token.staticcall(abi.encodeWithSelector(0x70a08231, to));
         uint256 prevBal = abi.decode(balData, (uint256));
 
         // update balance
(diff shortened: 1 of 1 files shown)


src/StdCheats.sol
```diff
@@ -593,15 +593,15 @@ abstract contract StdCheats is StdCheatsSafe {
...
+        (, bytes memory balData) = token.staticcall(abi.encodeWithSelector(0x00fdd58e, to, id));
         uint256 prevBal = abi.decode(balData, (uint256));
 
         // update balance
         stdstore.target(token).sig(0x00fdd58e).with_key(to).with_key(id).checked_write(give);
 
         // update total supply
         if (adjust) {
-            (, bytes memory totSupData) = token.call(abi.encodeWithSelector(0xbd85b039, id));
+            (, bytes memory totSupData) = token.staticcall(abi.encodeWithSelector(0xbd85b039, id));
```
mds1, line 604: Overall this PR seems like it should be ok: ERC-20 requires that `balanceOf` and `totalSupply` are view methods, and ERC-1155 requires that `balanceOf` is view also

Interestingly I didn't see the `totalSupply` specified in 1155 though—@wirew0lf any idea on if using a staticcall for the 1155 methods here might be problematic for some reason?

wirew0lf: Hey! yeah I agree it's better to use staticcall, good call. About the ERC115 @mds1 ERC1155 doesn't have a total supply by default, it is an extension [ERC1155Supply](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC1155/extensions/ERC1155Supply.sol).

On the cheatcode I acocunted for this and revert if adjust is set to true but contract is no ERC115Supply with the message `target contract is not ERC1155Supply`. But yeah on ERC1155Supply `totalSupply()` is also a view function so no problem by using staticcall.

mds1: Ah perfect, thanks @wirew0lf for the info and thanks @PaulRBerg for the PR!

mds1 (approved):

Issue #380, opened 2023-05-19
Add cheat/ util for checking addresses blacklisted by USDC/ USDT


PaulRBerg:
`testForkFuzz_` tests that run against USDC/ USDT break when the fuzzed address happens to be blacklisted.

This problem would be solved by a cheat/ util that assumes that the fuzzed address is not blacklisted. Here's a draft of what I have in mind:

```solidity
interface USDCLike {
    function isBlacklisted(address) external view returns (bool);
}

interface USDTLike {
    function isBlackListed(address) external view returns (bool);
}

function assumeNoBlacklisted(address token, address addr) internal virtual {
    if (token == 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48) {
        USDCLike usdc = USDCLike(token);
        vm.assume(!usdc.isBlacklisted(addr));
    } else if (token == 0xdAC17F958D2ee523a2206206994597C13D831ec7) {
        USDTLike usdt = USDTLike(token);
        vm.assume(!usdt.isBlackListed(addr));
    }
}
```

Of course, the function could be shortened by using a low-level call instead of using an interface; also, checking that `token` is USDC/ USDT is not strictly necessary, as we could check the success status of the low-level call, ignore reverts, and check if the return data is `false`.

mds1:
> tests that run against USDC/ USDT break when the fuzzed address happens to be blacklisted.

Were these blacklisted addresses in your code somewhere, or did forge generate them on its own? I'm assuming the former, since I don't know how they'd get into the forge dict otherwise, in which case it makes the argument for including this as a stdCheat a lot weaker (I'm not against it, but it does feel very specific to a use case so I'm hesitant to merge the PR)

PaulRBerg:
The blacklisted addresses were not in our code! Forge found them all by itself.

We started bumping into this scenario in our CI, so much so that we ended up writing our subroutine, and now we need the said subroutine in another repo, so I said that it might be a good thing to put it in Forge Std 😁

Here are some screenshots from February:

<img width="1109" alt="Screenshot 2023-05-24 at 9 48 34 PM" src="https://github.com/foundry-rs/forge-std/assets/8782666/c16a5dbc-4eb7-4843-ba74-faf49aca7b74">

<img width="954" alt="Screenshot 2023-05-24 at 9 48 49 PM" src="https://github.com/foundry-rs/forge-std/assets/8782666/dc628e7f-bb64-40b0-81b2-55f8c40ae0b4">

[...]

PR #384, opened 2023-05-21
feat: add "assumeNoBlacklisted" cheat

PaulRBerg:
Implements #380.

Btw, while writing the tests for this, I was looking at [`testAssumePayable`](https://github.com/foundry-rs/forge-std/blob/20872c5b1900526579159bdc6967f6b48c22e50e/test/StdCheats.t.sol#L319-L333) and wondering why it works?

I thought that `vm.expectRevert` only works for external reverts?

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -5,6 +5,7 @@ pragma experimental ABIEncoderV2;
 
 import {StdStorage, stdStorage} from "./StdStorage.sol";
 import {Vm} from "./Vm.sol";
+import {console2} from "./console2.sol";
 
 abstract contract StdCheatsSafe {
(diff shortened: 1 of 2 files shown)


mds1:
> Btw, while writing the tests for this, I was looking at [testAssumePayable](https://github.com/foundry-rs/forge-std/blob/20872c5b1900526579159bdc6967f6b48c22e50e/test/StdCheats.t.sol#L319-L333) and wondering why it works?
> 
> I thought that vm.expectRevert only works for external reverts?

It probably doesn't work right, this is a big footgun with expectRevert that will be fixed upstream soon, so that it can no longer be used on internal calls

PR #387, opened 2023-05-22
🐞 fix: assumePayable exclude 0x08

### LINKED ISSUE #386: `assumePayable` does not work for 0x08
According to https://www.evm.codes/precompiled?fork=shanghai, `0x08` will always returns `success` even when transfer fails. So we need modify the assumePayable function to check payable.

noyyyy:
close #386.

For minimum changes, just excludes 0x08

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -477,6 +477,7 @@ abstract contract StdCheatsSafe {
     // a cheat for fuzzing addresses that are payable only
     // see https://github.com/foundry-rs/foundry/issues/3631
     function assumePayable(address addr) internal virtual {
+        vm.assume(addr != address(8));
         (bool success,) = payable(addr).call{value: 0}("");
         vm.assume(success);
     }
--- a/test/StdCheats.t.sol
+++ b/test/StdCheats.t.sol
@@ -319,6 +319,10 @@ contract StdCheatsTest is Test {
     function testAssumePayable() external {
         // all should revert since these addresses are not payable
 
+        // precompiled contract 0x08
+        vm.expectRevert();
+        assumePayable(0x0000000000000000000000000000000000000008);
+
         // VM address
         vm.expectRevert();
         assumePayable(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);


mds1:
I don't follow why we want to exclude 0x08—it seems payable? The below does not revert

```solidity
function test_payablePrecompiles() external {
  (bool success,) = payable(address(8)).call{value: 1 ether}("");
  require(success, "call failed");
}
```

noyyyy:
> I don't follow why we want to exclude 0x08—it seems payable? The below does not revert
> 
> ```solidity
> function test_payablePrecompiles() external {
>   (bool success,) = payable(address(8)).call{value: 1 ether}("");
>   require(success, "call failed");
> }
> ```

I don't understant quitely either. But the following test will revert with `EvmError: Revert`.

```solidity
    function test_payablePrecompiles() external {
        payable(address(8)).transfer(1 ether);
    }
```

mds1:
That example fails because `.transfer` only forwards 2300 gas, which is insufficient gas for the precompile to execute, there it reverts. If you check this address on mainnet or goerli you will see successful ETH transfers to it, such as https://goerli.etherscan.io/tx/0x32bac49d3b8d4200946bdcc32b8f0c4f90bcce80757410f01477c9a1f871a748

Going to close this as a result since the address is payable, and if it were not the current check would fail :)

noyyyy:
> That example fails because `.transfer` only forwards 2300 gas, which is insufficient gas for the precompile to execute, there it reverts. If you check this address on mainnet or goerli you will see successful ETH transfers to it, such as https://goerli.etherscan.io/tx/0x32bac49d3b8d4200946bdcc32b8f0c4f90bcce80757410f01477c9a1f871a748
> 
> Going to close this as a result since the address is payable, and if it were not the current check would fail :)

Thanks for your answer. However, address 0x8 will lead to the fail of send ETH in fuzz test unless use `call` to send ETH rather than `transfer`. In most time, we will use `transfer` rather `call` as `transfer` is more safe.

For the original intention of `assumePayable`: https://github.com/foundry-rs/foundry/issues/3631, I still suggest excluding address 0x8 in `assumePayable`.

mds1:
You should not rely on the minimal gas forwarding of `transfer` for reentrancy protection, and in general you should not use `transfer`: https://consensys.net/diligence/blog/2019/09/stop-using-soliditys-transfer-now/

noyyyy:
> You should not rely on the minimal gas forwarding of `transfer` for reentrancy protection, and in general you should not use `transfer`: https://consensys.net/diligence/blog/2019/09/stop-using-soliditys-transfer-now/

This is really beyond my knowledge. Thanks for your time again!

PR #388, opened 2023-05-24
chore: add `Create2Factory` to `CommonBase`

### LINKED ISSUE #285: Make `CREATE2_FACTORY` internal (instead of private)
Here:
https://github.com/foundry-rs/forge-std/blob/4a79aca83f8075f8b1b4fe9153945fef08375630/src/StdUtils.sol#L24

With access to `CREATE2_FACTORY` in tests, I could do:
```solidity
vm.prank(CREATE2_FACTORY);
Contract c = new Contract{salt:""}(...args);
```

which would be much nicer than the current alternative: 
```solidity
bytes memory initCode = hashInitCode(type(Contract).creationCode,abi.encode(...args));
address ac = computeCreate2Address("",initCode);
``` 
About the above: 
1. it is more cumbersome, and 
2. it does not have the nice side-effect of actually creating an instance of `Contract`.

Sabnock01:
This PR moves the `CREATE2_FACTORY` address from `ScriptBase` to `CommonBase` to allow for its use in tests. Resolves #285

### DIFF
--- a/src/Base.sol
+++ b/src/Base.sol
@@ -9,6 +9,8 @@ abstract contract CommonBase {
     address internal constant VM_ADDRESS = address(uint160(uint256(keccak256("hevm cheat code"))));
     // console.sol and console2.sol work by executing a staticcall to this address.
     address internal constant CONSOLE = 0x000000000000000000636F6e736F6c652e6c6f67;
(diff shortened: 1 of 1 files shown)


mds1 (approved):
Thanks!

PR #389, opened 2023-05-24
add: destroyAccount cheatcode

joaquinlpereyra:
Introduces a new std cheatcode, `destroyAccount`, which mimics `SELFDESTRUCT` behavior. 

This is a companion PR to the one that introduces `resetNonce` to Forge itself: see the [Forge PR](https://github.com/foundry-rs/foundry/pull/5033) for rationale.

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -416,10 +416,11 @@ abstract contract StdCheatsSafe {
         (addr,) = makeAddrAndKey(name);
     }
 
-    // destroys an account inmediatly, sending the balance to beneficiary 
-    // destroying means: balance will be zero, code will be empty, nonce will be zero
-    // similar to selfdestruct but not identical: selfdestruct destroys code and nonce
-    // only after tx ends, this will run inmediatly
+    // Destroys an account immediately, sending the balance to beneficiary.
+    // Destroying means: balance will be zero, code will be empty, and nonce will be 0
+    // for EOAs or 1 for contract accounts.
+    // This is similar to selfdestruct but not identical: selfdestruct destroys code and nonce
+    // only after tx ends, this will run immediately.
     function destroyAccount(address who, address beneficiary) internal virtual {
         uint256 currBalance = who.balance;
         vm.etch(who, abi.encode());
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -416,6 +416,20 @@ abstract contract StdCheatsSafe {
         (addr,) = makeAddrAndKey(name);
     }
 
+    // destroys an account inmediatly, sending the balance to beneficiary 
+    // destroying means: balance will be zero, code will be empty, nonce will be zero
+    // similar to selfdestruct but not identical: selfdestruct destroys code and nonce
+    // only after tx ends, this will run inmediatly
+    function destroyAccount(address who, address beneficiary) internal virtual {
+        uint256 currBalance = who.balance;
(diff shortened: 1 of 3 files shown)


src/StdCheats.sol
```diff
@@ -416,6 +416,20 @@ abstract contract StdCheatsSafe {
         (addr,) = makeAddrAndKey(name);
     }
 
+    // destroys an account inmediatly, sending the balance to beneficiary 
+    // destroying means: balance will be zero, code will be empty, nonce will be zero
+    // similar to selfdestruct but not identical: selfdestruct destroys code and nonce
+    // only after tx ends, this will run inmediatly
```
mds1, line 422: Just some tweaks to make the comment a bit more readable/fix typos, I got confused at first since sometimes a line break was a new sentence, other times it wasn't
```suggestion
    // Destroys an account immediately, sending the balance to beneficiary.
    // Destroying means: balance will be zero, code will be empty, and nonce will be zero.
    // This is similar to selfdestruct but not identical: selfdestruct destroys code and nonce
    // only after tx ends, this will run immediately.
```

joaquinlpereyra:
@mds1 Commited your suggestions, thanks!

joaquinlpereyra:
Ah, actually, one of your suggestions was innacurate: because `destroyAccount` uses `etch` to set the code to zero _always_, the nonce will _always_ be zero after calling `destroyAccount`. Fixing it in another commit :)

mds1 (approved):
Thanks!

pcaversaccio:
Hey guys, just as a quick thought: currently calling `destroyAccount` also works for an EOA. I mean it doesn't make that much sense to reset the code there except for the nonce reset. So I would recommend using `destroyContract` and `resetEOA` (or similar) to separate semantics. Let me know your thoughts.

Also, just as a reference for anyone who wants to properly test `selfdestruct`, I recommend using it via `setUp`; an example [here](https://github.com/pcaversaccio/tornado-cash-exploit/blob/main/test/MetamorphicContract.t.sol).

mds1:
Hmm yea fair point. I'd be supportive of switching to `destroyContract` and `resetEOA`, but I also don't feel strongly here since this is definitely a "know what you're doing when using this" type of feature anyway. @joaquinlpereyra curious to hear what you think?

(Also nice pattern for selfdestruct tests, I like that)

joaquinlpereyra:
Mhm, honestly, I feel like the distinction is not very useful and just muddles the API. `destroyAccount`... destroys an account. If you destroy an EOA the `etch` will simply be a `NOOP` and everyone will be happy. 

And for the `setUp` hack, while it works, it it is a hack at best and limiting at worst. Using `setUp` to `SELFDESTRUCT` means you need to... well, `SELFDESTRUCT` in the `setUp`, which will run for _every_ test. That might (most probably...) not be what you want. I don't see a reason to use it now if `destroyAccount` is available.

pcaversaccio:
I actually would consider `destroyAccount` the hack (that's why it's called cheatcode at the end of the day...), and while I agree that it might be suboptimal that it's run in every test, it's the only way to correctly test the `selfdestruct` behaviour. Also, let's say you want to test an interaction with a selfdestructed contract within the same test (i.e. testing the "feature" of the `selfdestruct` opcode that you can interact until the end of the tx), this won't be possible with `destroyAccount`. At the end, as @mds1 said, it's a "know what you're doing when using this" type of feature and maybe not worth the effort.

PR #398, opened 2023-06-02
fix: improve DX

ZeroEkkusu:
## Motivation

- Improve the DX
- Fix warnings
- Clean up

## Solution

Various fixes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,9 @@ jobs:
       - name: Print forge version
         run: forge --version
 
-      # Backwards compatibility checks.
+      # Backwards compatibility checks:
+      # - the oldest and newest version of each supported minor version
+      # - versions with specific issues
       - name: Check compatibility with latest
         if: always()
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,6 +74,10 @@ jobs:
             exit 1
           fi
 
+      - name: Check compatibility with 0.6.12
+        if: always()
+        run: forge build --skip test --use solc:0.6.12
+
       - name: Check compatibility with 0.6.2
         if: always()
         run: |
--- a/src/StdChains.sol
+++ b/src/StdChains.sol
@@ -1,8 +1,6 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.2 <0.9.0;
 
-pragma experimental ABIEncoderV2;
-
 import {VmSafe} from "./Vm.sol";
 
 /**
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -194,7 +194,7 @@ abstract contract StdCheatsSafe {
     }
 
     // Checks that `addr` is not blacklisted by token contracts that have a blacklist.
-    function assumeNoBlacklisted(address token, address addr) internal virtual {
+    function assumeNoBlacklisted(address token, address addr) internal view virtual {
         // Nothing to check if `token` is not a contract.
         uint256 tokenCodeSize;
         assembly {
@@ -214,13 +214,8 @@ abstract contract StdCheatsSafe {
         vm.assume(!success || abi.decode(returnData, (bool)) == false);
     }
 
-    function assumeNoPrecompiles(address addr) internal virtual {
-        // Assembly required since `block.chainid` was introduced in 0.8.0.
-        uint256 chainId;
-        assembly {
-            chainId := chainid()
-        }
-        assumeNoPrecompiles(addr, chainId);
+    function assumeNoPrecompiles(address addr) internal pure virtual {
+        assumeNoPrecompiles(addr, _pureChainId());
     }
 
     function assumeNoPrecompiles(address addr, uint256 chainId) internal pure virtual {
@@ -514,6 +509,25 @@ abstract contract StdCheatsSafe {
     function assumePayable(address addr) internal virtual {
         (bool success,) = payable(addr).call{value: 0}("");
         vm.assume(success);
+    }
+
+    function _viewChainId() private view returns (uint256 chainId) {
+        // Assembly required since `block.chainid` was introduced in 0.8.0.
+        assembly {
+            chainId := chainid()
+        }
+
+        //Silence warnings in older Solc versions.
+        address(this);
+    }
+
+    function _pureChainId() private pure returns (uint256 chainId) {
+        function() internal view returns (uint256) fnIn = _viewChainId;
+        function() internal pure returns (uint256) pureChainId;
+        assembly {
+            pureChainId := fnIn
+        }
+        chainId = pureChainId();
     }
 }
 
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -519,14 +519,17 @@ abstract contract StdCheatsSafe {
         vm.assume(success);
     }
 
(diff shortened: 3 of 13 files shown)


mds1 (changes requested):
Agree with adding this method, just lots of comments around the details and other changes. 

Can we split this into multiple, more tightly scoped PRs? I think one PR for this cheat, and a second for the various cleanup would be a good split. That will make things easier to review, and result in a better diff/auto-generated-changelog for users checking out PRs to see new features

test/StdCheats.t.sol
```diff
@@ -505,3 +517,13 @@ contract RevertingContract {
         revert();
     }
 }
+
+contract CtWithConstructorArgs {
```
mds1, line 521: Does `Ct` stand for `Contract`? Let's use a more clear name here also
```suggestion
contract TestContractWithConstructorArgs {
```

test/StdCheats.t.sol
```diff
@@ -505,3 +517,13 @@ contract RevertingContract {
...
+    uint256 public immutable x;
+    bool public y;
+
+    constructor(uint256 _x, bool _y) payable {
```
mds1, line 525: We should have another test contract for a non-payable constructor

src/StdChains.sol
```diff
@@ -1,8 +1,6 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.2 <0.9.0;
 
-pragma experimental ABIEncoderV2;
```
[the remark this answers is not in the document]
ZeroEkkusu: It's required for unsupported types. However, it must be added in the end-file (`MyTest.sol`) to have effect - adding it to files in the middle (`StdChains.sol`) doesn't do anything when those files don't use the unsupported types.

It should be removed in some other files as well.

src/StdCheats.sol
```diff
@@ -673,4 +687,19 @@ abstract contract StdCheats is StdCheatsSafe {
         // update owner
         stdstore.target(token).sig(0x6352211e).with_key(id).checked_write(to);
     }
+
+    function etchify(bytes memory creationCode, address to) internal virtual {
```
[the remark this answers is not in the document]
mds1: hmm `etchWithCreationCode` maybe? I think `deployCodeTo` still works though:
- `deployCode` uses CREATE and goes to any address
- `deployCodeTo` uses cheats to place the code at a specific address, so by definition it cannot use CREATE

src/StdCheats.sol
```diff
@@ -215,13 +215,8 @@ abstract contract StdCheatsSafe {
...
 
-    function assumeNoPrecompiles(address addr) internal virtual {
-        // Assembly required since `block.chainid` was introduced in 0.8.0.
-        uint256 chainId;
-        assembly {
-            chainId := chainid()
-        }
-        assumeNoPrecompiles(addr, chainId);
+    function assumeNoPrecompiles(address addr) internal pure virtual {
+        assumeNoPrecompiles(addr, _pureChainId());
```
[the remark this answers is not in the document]
mds1: Ah interesting, thanks for the clarification there. So should we just keep the `pure` version around then since that can be used anywhere the `view` version is used?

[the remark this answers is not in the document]
ZeroEkkusu: `_viewChainId` is used to feed it to `_pureChainId`.

src/consoleSafe.sol
```diff
@@ -3,7 +3,7 @@ pragma solidity >=0.6.2 <0.9.0;
 
 /// @author philogy <https://github.com/philogy>
 /// @dev Code generated automatically by script.
-library safeconsole {
+library consoleSafe {
```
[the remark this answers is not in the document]
ZeroEkkusu: I'll revert it, but https://github.com/foundry-rs/forge-std/pull/375 is only 3 weeks old. So far, `<Name>Safe` has been the convention.

ZeroEkkusu:
I've removed the `0.6.12` backward compatibility check in the CI because there are no known issues there. @mds1 added it in https://github.com/foundry-rs/forge-std/pull/217/commits/fab939eae5d1c1cf5413b5ac397cd06d07ede9b0; I approved it anyway, not to block the PR.

I'll create a new PR for `deployCodeTo`.

test/StdCheats.t.sol
```diff
@@ -505,3 +517,13 @@ contract RevertingContract {
         revert();
     }
 }
+
+contract CtWithConstructorArgs {
+    uint256 public immutable x;
+    bool public y;
```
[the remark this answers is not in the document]
ZeroEkkusu: `constant` must be initialized immediately. I'll add `z` anyway.

.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: Print forge version
         run: forge --version
 
-      # Backwards compatibility checks.
+      # Backwards compatibility checks: 0.y.0 (or oldest), and releases with known issues are checked - do not remove those.
```
mds1, line 24: I thought we added the 0.6.12 check because the idea was to test the oldest and newest version of each supported minor version, since there are sometimes breaking changes in between. So we had 0.6.2 and 0.6.12 (0.6.2 is when interface inheritance was introduced, which is why we didn't support 0.6.0). Then we had 0.7.0 and 0.7.6 for the same reason. 

But regardless, since these checks are cheap (fast), we may as well just keep it to be safe and lean towards being more conservative?

src/StdCheats.sol
```diff
@@ -195,7 +195,7 @@ abstract contract StdCheatsSafe {
     }
 
     // Checks that `addr` is not blacklisted by token contracts that have a blacklist.
-    function assumeNoBlacklisted(address token, address addr) internal virtual {
+    function assumeNoBlacklisted(address token, address addr) internal view virtual {
```
mds1, line 198: Realizing now this method name has a typo, should be `assumeNotBlacklisted`. To avoid a breaking change we should probably just add that version as an alias. Unrelated to this PR so can handle it separately, just noting here

test/StdCheats.t.sol
```diff
@@ -370,7 +370,7 @@ contract StdCheatsTest is Test {
 
 contract StdCheatsMock is StdCheats {
     // We deploy a mock version so we can properly test expected reverts.
-    function assumeNoBlacklisted_(address token, address addr) external {
+    function assumeNoBlacklisted_(address token, address addr) external view {
         return assumeNoBlacklisted(token, addr);
```
mds1, line 374: Then we'd want to change the test name to `assumeNotBlacklisted` also

src/StdCheats.sol
ZeroEkkusu: Didn't make the change in this PR.

test/StdCheats.t.sol
ZeroEkkusu: See above.

src/StdCheats.sol
mds1: Pushed 2feefd2 to make this change

test/StdCheats.t.sol
mds1: Pushed 2feefd2 to make this change

mds1 (approved):
This is awesome, thanks so much @ZeroEkkusu! I added a few small commits (including one to alias `assumeNoBlacklisted` -> `assumeNotBlacklisted`, cc @PaulRBerg) 

CI is passing so I'm going to merge this to unblock #401

PR #403, opened 2023-06-17
ci: check for warnings

### LINKED ISSUE #402: ci: check for warnings
## Problem

If there are warnings, they are shown every time when contracts are compiled (e.g. `forge test`).

## Solution

Check for warnings in the CI backward compatibility check and the other relevant sections.

ZeroEkkusu:
## Motivation

Resolve #402 

## Solution

Fail CI if there's `Warning` in output.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,25 +22,65 @@ jobs:
         run: forge --version
 
       # Backwards compatibility checks.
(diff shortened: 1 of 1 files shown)


onbjerg (approved):
makes sense

mds1 (approved):
I like this, thanks!

PR #407, opened 2023-06-23
feat: Refactor `assumeNot*` std cheats

### LINKED ISSUE #405: `assumeNoPrecompiles` doesn't exclude VM address
https://github.com/foundry-rs/forge-std/blob/master/src/StdCheats.sol#L226-L247
The function should have a statement like `vm.assume(addr != address(vm))` because any interaction with the VM results in a revert.

Sabnock01:
Excludes the `vm` address `0x7109709ECfa91a80626fF3989D68f67F5b1DD12D` from the `assumeNoPrecompiles` cheat code. 

Resolves #405 

Note that `address(vm)` is used instead of `address(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D)` (such as in  `testAssumePayable`) so that it fits squarely into the current implementation of the cheat code and respective test with minimal confusion. Open to changing this though.

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -279,8 +279,14 @@ abstract contract StdCheatsSafe {
         assumeAddressIsNot(addressType4, addr);
     }
 
-    function _isPayable(address addr) private returns (bool) {
-        if (addr.code.length == 0) {
+    function _isPayable(address addr) private returns (bool) {        
+        uint256 size;
+        assembly {
+            size := extcodesize(addr)
+        }
+
+        if (size == 0) {
+            // return false if no code
             return false;
         } else {
             require(addr.balance < type(uint256).max, "balance exceeds max uint256");
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -233,7 +233,7 @@ abstract contract StdCheatsSafe {
         assumeNotBlacklisted(token, addr);
     }
 
-    function assumeAddressIsNot(AddressType addressType, address addr) internal virtual {
+    function assumeAddressIsNot(address addr, AddressType addressType) internal virtual {
         if (addressType == AddressType.Payable) {
             assumeNotPayable(addr);
         } else if (addressType == AddressType.NonPayable) {
@@ -248,8 +248,8 @@ abstract contract StdCheatsSafe {
     }
 
     function assumeAddressIsNot(address addr, AddressType addressType1, AddressType addressType2) internal virtual {
-        assumeAddressIsNot(addressType1, addr);
-        assumeAddressIsNot(addressType2, addr);
+        assumeAddressIsNot(addr, addressType1);
+        assumeAddressIsNot(addr, addressType2);
     }
 
     function assumeAddressIsNot(
@@ -258,9 +258,9 @@ abstract contract StdCheatsSafe {
         AddressType addressType2,
         AddressType addressType3
     ) internal virtual {
-        assumeAddressIsNot(addressType1, addr);
-        assumeAddressIsNot(addressType2, addr);
-        assumeAddressIsNot(addressType3, addr);
+        assumeAddressIsNot(addr, addressType1);
+        assumeAddressIsNot(addr, addressType2);
+        assumeAddressIsNot(addr, addressType3);
     }
 
     function assumeAddressIsNot(
@@ -270,16 +270,16 @@ abstract contract StdCheatsSafe {
         AddressType addressType3,
         AddressType addressType4
     ) internal virtual {
-        assumeAddressIsNot(addressType1, addr);
-        assumeAddressIsNot(addressType2, addr);
-        assumeAddressIsNot(addressType3, addr);
-        assumeAddressIsNot(addressType4, addr);
+        assumeAddressIsNot(addr, addressType1);
+        assumeAddressIsNot(addr, addressType2);
+        assumeAddressIsNot(addr, addressType3);
+        assumeAddressIsNot(addr, addressType4);
     }
 
-    // This function checks whether an address, `addr`, is payable. It returns true if addr has no code in which case
-    // it is an EOA. It will then send 1 wei to `addr` and check the success return value. There may be state changes
-    // depending on the fallback/receive logic implemented by `addr` which should be taken into account when
-    // this function is used.
+    // This function checks whether an address, `addr`, is payable. It works by sending 1 wei to
+    // `addr` and checking the `success` return value.
+    // NOTE: This function may result in state changes depending on the fallback/receive logic
+    // implemented by `addr`, which should be taken into account when this function is used.
     function _isPayable(address addr) private returns (bool) {
         require(
             addr.balance < UINT256_MAX,
@@ -298,6 +298,9 @@ abstract contract StdCheatsSafe {
         return success;
     }
 
+    // NOTE: This function may result in state changes depending on the fallback/receive logic
+    // implemented by `addr`, which should be taken into account when this function is used. See the
+    // `_isPayable` method for more information.
     function assumePayable(address addr) internal virtual {
         vm.assume(_isPayable(addr));
     }
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -221,7 +221,7 @@ abstract contract StdCheatsSafe {
         (success, returnData) = token.staticcall(abi.encodeWithSelector(0xe47d6060, addr));
         vm.assume(!success || abi.decode(returnData, (bool)) == false);
     }
-    
+
     // Checks that `addr` is not blacklisted by token contracts that have a blacklist.
     // This is identical to `assumeNotBlacklisted(address,address)` but with a different name, for
     // backwards compatibility, since this name was used in the original PR which has already has
@@ -306,6 +306,7 @@ abstract contract StdCheatsSafe {
 
     function assumeNoZeroAddress(address addr) internal virtual {
         vm.assume(addr != address(0));
+    }
 
     function assumeNoPrecompiles(address addr) internal pure virtual {
         assumeNoPrecompiles(addr, _pureChainId());
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -305,7 +305,7 @@ abstract contract StdCheatsSafe {
         // address), but the same rationale for excluding them applies so we include those too.
 
         // These should be present on all EVM-compatible chains.
-        vm.assume((addr < address(0x1) || addr > address(0x9)) && addr != address(vm));
+        vm.assume(addr < address(0x1) || addr > address(0x9));
 
         // forgefmt: disable-start
         if (chainId == 10 || chainId == 420) {
@@ -325,7 +325,9 @@ abstract contract StdCheatsSafe {
 
     function assumeNoForgeAddresses(address addr) internal virtual {
         // vm and console addresses
-        vm.assume(addr != 0x7109709ECfa91a80626fF3989D68f67F5b1DD12D && addr != 0x000000000000000000636F6e736F6c652e6c6f67);
+        vm.assume(
+            addr != 0x7109709ECfa91a80626fF3989D68f67F5b1DD12D && addr != 0x000000000000000000636F6e736F6c652e6c6f67
+        );
     }
 
     function readEIP1559ScriptArtifact(string memory path)
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -281,31 +281,21 @@ abstract contract StdCheatsSafe {
     // depending on the fallback/receive logic implemented by `addr` which should be taken into account when
     // this function is used.
     function _isPayable(address addr) private returns (bool) {
-        uint256 size;
-        assembly {
-            size := extcodesize(addr)
-        }
+        require(
+            addr.balance < UINT256_MAX,
+            "StdCheats _isPayable(address): Balance equals max uint256, so it cannot receive any more funds"
+        );
+        uint256 origBalanceTest = address(this).balance;
+        uint256 origBalanceAddr = address(addr).balance;
 
-        if (size == 0) {
-            // return true if no code
-            return true;
-        } else {
-            require(
-                addr.balance < UINT256_MAX,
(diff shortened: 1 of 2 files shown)


src/StdCheats.sol
```diff
@@ -228,7 +228,7 @@ abstract contract StdCheatsSafe {
         // address), but the same rationale for excluding them applies so we include those too.
 
         // These should be present on all EVM-compatible chains.
-        vm.assume(addr < address(0x1) || addr > address(0x9));
+        vm.assume((addr < address(0x1) || addr > address(0x9)) && addr != address(vm));
```
mds1, line 231: Should we also exclude the `console.log` address here, which is `0x000000000000000000636F6e736F6c652e6c6f67`?

Though, both vm and console aren't technically precompiles in the sense meant by this function, so perhaps we should instead go with something like [this](https://twitter.com/msolomon44/status/1622596868105732096):

```solidity
enum AddressType {
  ZeroAddress,
  Precompiles
  Payable
  NonPayable
  ForgeAddresses
  ...whatever else
}

function assumeAddressIsNot(AddressType addressType) internal { ... }
function assumeAddressIsNot(AddressType addressType1, AddressType addressType2) internal { ... }
// repeat for length of the enum
```

Sabnock01: I like this approach much better. Will rework.

Sabnock01: Actually I really like the array version. Less code and easier testing. 

Getting a lot of issues testing the overloaded one. 
```
[FAIL. Reason: The `vm.assume` cheatcode rejected too many inputs (65536 allowed)]
```

mds1: > Prefer the array version personally. Less code and easier to test

How is the UX with the array version? Main reason I though overloads would be better is:

```solidity
// This seems like good UX for devs
assumeAddressIsNot(1, 2, 3);

// This seems like bad UX (using a uint instead of enum for brevity in this example)
uint[] memory x = new uint[](3);
x[0] = 1;
x[1] = 2;
x[2] = 3;
assumeAddressIsNot(x);
```

> How do you propose to assert that an address is payable vs nonpayable and why include those in the enum exactly? Not positive if there's an optimal way to do this other than doing `call` on the address with some `msg.value` and checking `success`.

In the forge-std `assumePayable` method we just check `success` on a zero value call. But on second thought I'm not sure if our tests are sufficiently robust, e.g. if a contract has a non-payable `fallback` and no `receive`, that contract is not payable, and I think our `assumePayable` would wrongly return `true`. 

Better way is probably what you said, specifically: 
- Return true if address has no code
- Otherwise, ensure recipient's current ETH balance is less than maxUint, deal 1 wei to test contract, send it to recipient and check success, reset test contract and recipient balances to their original values.

> In both impl's, I have functions for each enum type that I check against in `assumeAddressIsNot` but both this and the overloaded version fail with too many `vm.assume` rejects when using a fuzz test. What do you think of this method of testing where instead of calling `assumeAddressIsNot` all the derivative functions are called?

If it helps, we can raise the `assume` limit for that test only using the [inline config](https://book.getfoundry.sh/reference/config/inline-test-config) approach. I think the most robust way would be something like the below, so if a new address type is added it's automatically tested

```solidity
function testAssumeAddressIsNot(address addr) external {
  for (uint8 i = 0; i < type(AddressType).max; i++) {
    assumeAddressIsNot(AddressType(i));
  }
  assertTrue(addr != address(0));
  assertTrue(
      addr < address(1) || (addr > address(9) && addr < address(0x4200000000000000000000000000000000000000))
            || addr > address(0x4200000000000000000000000000000000000800)        
    );
  assertTrue(addr != 0x7109709ECfa91a80626fF3989D68f67F5b1DD12D && addr != 0x000000000000000000636F6e736F6c652e6c6f67);

}
```

Sabnock01: Well `Payable` and `NonPayable` are mutually exclusive right, so we wouldn't need to iterate through the whole enum but instead n - 1?

mds1:
> That for loop is valid Solidity? Don't think there's a built-in way to get the length of an enum and it doesn't seem to work in either my forge test or chisel.

Hm it should work, see the [solidity docs](https://docs.soliditylang.org/en/latest/types.html#enums):
> Using `type(NameOfEnum).min` and `type(NameOfEnum).max` you can get the smallest and respectively largest value of the given enum.

---

> `Payable` and `NonPayable` are mutually exclusive right, so we wouldn't need to iterate through the whole enum but instead n - 1?

Ah good point. So maybe we put payable/nonpayable as indexes 0 and 1 in the enum, then we have:
- main test that iterates through indicies 2-n for the non-mutually-exclusive cases
- second test for payable
- third test for nonpayable

Wdyt of that? This way new values added to the end of the enum are automatically tested

> You still have to provide the address itself as an argument to `assumeAddressIsNot` right? That's the way `assumeNoBlacklisted`, `assumeNoPrecompiles`, and  `assumePayable` are.

Oops yes, sorry forgot to include that 😅

src/StdCheats.sol
mds1: > That for loop is valid Solidity? Don't think there's a built-in way to get the length of an enum and it doesn't seem to work in either my forge test or chisel.

Hm it should work, see the [solidity docs](https://docs.soliditylang.org/en/latest/types.html#enums):
> Using `type(NameOfEnum).min` and `type(NameOfEnum).max` you can get the smallest and respectively largest value of the given enum.

---

> `Payable` and `NonPayable` are mutually exclusive right, so we wouldn't need to iterate through the whole enum but instead n - 1?

Ah good point. So maybe we put payable/nonpayable as indexes 0 and 1 in the enum, then we have:
- main test that iterates through indicies 2-n for the non-mutually-exclusive cases
- second test for payable
- third test for nonpayable

Wdyt of that? This way new values added to the end of the enum are automatically tested

> You still have to provide the address itself as an argument to `assumeAddressIsNot` right? That's the way `assumeNoBlacklisted`, `assumeNoPrecompiles`, and  `assumePayable` are.

Oops yes, sorry forgot to include that 😅

Sabnock01: So even when setting `max_fuzz_rejects` to some absurdly high number like `10_000_000` which takes several minutes to complete, a fuzz test on `assumeNoNonPayable` still fails. 

So for both `assumeNoPayable` and `assumeNonPayable` I went with modified versions of the old test for the now deprecated `assumePayable`. 

Latest impl pushed.

mds1: Awesome sounds great, thanks! Lmk if this is ready for review or if you still have changes left—will be tied up today/tomorrow but should be able to review this thursday

src/StdCheats.sol
```diff
@@ -222,6 +233,84 @@ abstract contract StdCheatsSafe {
...
+
+            // reset balances
+            vm.deal(address(this), origBalanceTest);
+            vm.deal(addr, origBalanceAddr);
+
+            return success;
+        }
+    }
+
+    function assumeNoPayable(address addr) internal virtual {
```
mds1, line 302: Hmm, wondering if we should use "No" or "Not" in each of these: `assumeNoPrecompiles` makes sense, but it could have instead been "assumeNotPrecompile". Maybe the rule of thumb is "No" for plural and "Not" for singular? By that logic `assumeNoPayable`, `assumeNoNonPayable, `assumeNoZeroAddress` should all be "Not". I'm not great at grammar, so maybe I'm overcomplicating this 😅

Maybe we should use `assumePayable` and `assumeNotPayable` for those two, instead of the double negative of `assumeNoNonPayable`?

src/StdCheats.sol
```diff
@@ -249,6 +338,13 @@ abstract contract StdCheatsSafe {
         // forgefmt: disable-end
     }
 
+    function assumeNoForgeAddresses(address addr) internal pure virtual {
+        // vm and console addresses
+        vm.assume(
+            addr != 0x7109709ECfa91a80626fF3989D68f67F5b1DD12D || addr != 0x000000000000000000636F6e736F6c652e6c6f67
```
mds1, line 344:
```suggestion
            addr != address(vm) || addr != 0x000000000000000000636F6e736F6c652e6c6f67
```

src/StdCheats.sol
```diff
@@ -222,6 +233,84 @@ abstract contract StdCheatsSafe {
...
+
+    function _isPayable(address addr) private returns (bool) {
+        uint256 size;
+        assembly {
+            size := extcodesize(addr)
+        }
+
+        if (size == 0) {
+            // return false if no code
+            return false;
```
mds1, line 287: Addresses with no code are payable, e.g. EOAs. This hints we should add some unit tests for this method, maybe should make it `internal` to simplofy the testing. OTOH, the logic in the `else` branch works for EOAs too, so maybe we just remove the size check and make this branchless 🤔
```suggestion
            return true;
```

src/StdCheats.sol
```diff
@@ -222,6 +233,84 @@ abstract contract StdCheatsSafe {
...
+        } else {
+            require(addr.balance < UINT256_MAX, "balance exceeds max uint256");
```
mds1, line 289: this is the error string convention we use, I also clarified the error message
```suggestion
            require(addr.balance < UINT256_MAX, "StdCheats _isPayable(address): Balance equals max uint256, so it cannot receive any more funds");
```

src/StdCheats.sol
```diff
@@ -222,6 +233,84 @@ abstract contract StdCheatsSafe {
...
+            uint256 origBalanceTest = address(this).balance;
+            uint256 origBalanceAddr = address(addr).balance;
+            (bool success,) = payable(addr).call{value: 1}("");
```
mds1, line 292: This will fail if the test contract has zero ETH, and therefore may erroneously return `success = false` in that case, so we should `vm.deal(address(this), 1)` beforehand.

Also one side effect of this method is there may be state changes depending on the fallback/receive logic implemented by `addr`. We should make sure this is documented in comments here + in the book

test/StdCheats.t.sol
```diff
@@ -335,41 +335,52 @@ contract StdCheatsTest is Test {
...
+        // skip over Payable and NonPayable enums
+        for (uint8 i = 2; i < uint8(type(AddressType).max); i++) {
+            assumeAddressIsNot(AddressType(i), addr);
+        }
+        assertTrue(addr != address(0));
+        assertTrue(addr < address(1) || addr > address(9));
         assertTrue(
-            addr < address(1) || (addr > address(9) && addr < address(0x4200000000000000000000000000000000000000))
-                || addr > address(0x4200000000000000000000000000000000000800)
+            addr != 0x7109709ECfa91a80626fF3989D68f67F5b1DD12D || addr != 0x000000000000000000636F6e736F6c652e6c6f67
```
mds1, line 346: I think we should always have the `vm` var available, so we can use `address(vm)` here/below instead to make it a bit more readable

test/StdCheats.t.sol
```diff
@@ -335,41 +335,52 @@ contract StdCheatsTest is Test {
...
-    function testAssumePayable() external {
-        // We deploy a mock version so we can properly test the revert.
-        StdCheatsMock stdCheatsMock = new StdCheatsMock();
+    function testAssumeNoNonPayable() external {
         // all should revert since these addresses are not payable
 
         // VM address
         vm.expectRevert();
-        stdCheatsMock.exposed_assumePayable(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);
+        assumeNoNonPayable(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);
```
mds1, line 368: This will break in forge v1 since JUMPs to internal functions are not supported for `vm.expectRevert` (the fact that they ever were supported was an accident, and they are not well supported—only the first assertion in this test will actually run, then execution silently halts, so it's a bad footgun 😅). So let's make sure to keep the original approach of making sure `expectRevert` is only every used with external calls (hence the `stdCheatsMock.exposed_assumePayable` method). More info in the foundry changelog and https://github.com/foundry-rs/foundry/pull/4945

Let's also add some success cases here that are epected to not revert, e.g. EOAs and a mock payable contract

mds1 (commented):
Looks great overall, thank you! Lots of small comments around details / some nits

Sabnock01:
> Looks great overall, thank you! Lots of small comments around details / some nits

Cool. I like all of these suggestions. It's a big PR to be fair.

src/StdCheats.sol
Sabnock01: Yeah, that's a good point.

src/StdCheats.sol
```diff
@@ -193,6 +196,14 @@ abstract contract StdCheatsSafe {
         uint256 key;
     }
 
+    enum AddressType {
```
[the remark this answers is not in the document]
Sabnock01: How do you imagine that one fitting in with it taking an extra `token` param?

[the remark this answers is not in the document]
mds1: ah good point forgot about that, ok let's ignore this comment 😅

src/StdCheats.sol
Sabnock01: Changed them all to singular and `Not*`.

Sabnock01: One thing I did notice when adjusting the tests for `Payable`/`NotPayable` is that in the `forge test` context, the console address has no code while the vm has a fallback and sending 1 wei results in the payable check returns true.

test/StdCheats.t.sol
Sabnock01: Acknowledged and done.

Sabnock01:
Let me know what you make of these adjustments.

mds1:
This looks great, thanks a lot @Sabnock01! I pushed one commit with a few small tweaks, lmk if you have any feedback/objections there, and afterwards I'll merge

Sabnock01:
> This looks great, thanks a lot @Sabnock01! I pushed one commit with a few small tweaks, lmk if you have any feedback/objections there, and afterwards I'll merge

These are good. Should have caught more of them honestly but was a pleasure working with you on this one.

mds1:
Likewise, thanks for the quality PR! '🙌

I'll be putting out a new forge-std release with this shortly

PR #415, opened 2023-07-07
Add checked_write_signed for int256

0xb10ckdev:
There is no `checked_write` function for `int256` param.
There are only `checked_write` functions with `address`, `uint256`, `bool` and `bytes32` parameters now.
Actually, for `int256` variables, the below code can be used.
```
int256 value = -100;
stdstore
    .target(address(addr))
    .sig("var()")
    .checked_write(uint256(value));
```
But it might be better if it has `checked_write` function for `int256` params since the above code is not good to understand.

### DIFF
--- a/src/StdStorage.sol
+++ b/src/StdStorage.sol
@@ -239,7 +239,7 @@ library stdStorage {
         checked_write(self, bytes32(amt));
     }
 
-    function checked_write_signed(StdStorage storage self, int256 val) internal {
+    function checked_write_int(StdStorage storage self, int256 val) internal {
         checked_write(self, bytes32(uint256(val)));
     }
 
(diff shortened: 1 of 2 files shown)


src/StdStorage.sol
```diff
@@ -239,6 +239,10 @@ library stdStorage {
         checked_write(self, bytes32(amt));
     }
 
+    function checked_write_signed(StdStorage storage self, int256 val) internal {
```
mds1, line 242: Any objections to this name change to directly reflect the type? This makes it more analogous to how we handle this elsewhere (e.g. `console.logBytes`)

```suggestion
    function checked_write_int(StdStorage storage self, int256 val) internal {
```

mds1 (approved):
Thank you!

PR #416, opened 2023-07-08
feat(cheats): ffi

ZeroEkkusu:
## Motivation

When I started using Foundry, I didn't find `ffi` intuitive. I thought I could give it a command, but, I was supposed to initialize an array first, assign values, etc.

This cheat makes it work the way I thought it did. I'll let you decide if it's worth adding.

## Solution

**Before**

```solidity
string[] memory inputs = new string[](3);

inputs[0] = "echo";
inputs[1] = "-n";
inputs[2] = "hello world";

bytes memory res = vm.ffi(inputs);
```

**After**

```solidity
bytes memory res = ffi("echo -n", raw("hello world"));
```

Note: `raw` passes the text as-is.

<br>
<br>

**Before**

```solidity
string[] memory inputs = new string[](4);

inputs[0] = "yarn";
inputs[1] = "ts-node";
inputs[2] = "path/to/script.ts";
inputs[3] = vm.toString(
    abi.encode(
        user,
        data
    )
);

bytes memory res = vm.ffi(inputs);
```

**After**

```solidity
bytes memory res = ffi(
    "yarn ts-node path/to/script.ts",
    vm.toString(abi.encode(user, data))
);
```

Note: `ffi` accepts up to five arguments. Both examples used two.

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -578,6 +578,8 @@ abstract contract StdCheatsSafe {
         }
 
         out = vm.ffi(ffiInput_fs);
+
+        ffiInput_fs = new string[](0);
     }
 
     function raw(string memory text) internal pure virtual returns (string memory) {
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -11,6 +11,10 @@ abstract contract StdCheatsSafe {
 
     bool private gasMeteringOff;
 
+    string[] private ffiInput_fs;
+    // keccak256("forge-std-rawIndicator")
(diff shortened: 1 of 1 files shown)


mds1:
This is pretty neat and definitely nicer UX, but in the spirit of https://github.com/foundry-rs/foundry/issues/3782 and gradually migrating towards that I'd rather implement this natively in forge and minimize addition of new StdCheats

ZeroEkkusu:
Alright!

ZeroEkkusu:
Alright!

PR #421, opened 2023-07-17
demonstrate error in json name

DavidOrchard:


mds1:
It looks like this is just a test of a native cheat code, not a forge-std stdCheat. Going to close this as a result, but feel free to add this test upstream to in the foundry repo

PR #423, opened 2023-07-18
feat(`vm`): add keyExists

Evalir:
Upstream from https://github.com/foundry-rs/foundry/pull/5431

### DIFF
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -259,6 +259,9 @@ interface VmSafe {
     function parseJsonBytes32(string calldata, string calldata) external returns (bytes32);
     function parseJsonBytes32Array(string calldata, string calldata) external returns (bytes32[] memory);
 
+    // Checks if a key exists in a JSON or TOML object.
+    function keyExists(string calldata, string calldata) external view returns (bool);
+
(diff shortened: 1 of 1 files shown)


src/Vm.sol
```diff
@@ -259,6 +259,9 @@ interface VmSafe {
     function parseJsonBytes32(string calldata, string calldata) external returns (bytes32);
     function parseJsonBytes32Array(string calldata, string calldata) external returns (bytes32[] memory);
 
+    // Checks if a key exists in a JSON or TOML object.
+    function keyExists(string calldata, string calldata) external view returns (bool);
```
mds1, line 263: We try to always used named inputs/outputs here for clarity, do you mind updating? (looks like the above `parseJson*` methods are also missing named params if you wanted to fix those also 😅

Evalir: oh yeah just followed what was above 😅 will update them all!

mds1: hey @Evalir just bumping this

Evalir:
@mds1 should be g2g!

mds1 (approved):
Thanks!

PR #424, opened 2023-07-19
docs: new readme

ZeroEkkusu:
## Motivation

Prettify and update readme.

## Solution

Check out the [new readme](https://github.com/foundry-rs/forge-std/tree/19de2f08b029f9b6f36d9b8d9617cdeef1365924).

<br>

*Note Any explanations you deem useful should be moved to the book.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,250 +1,170 @@
-# Forge Standard Library • [![CI status](https://github.com/foundry-rs/forge-std/actions/workflows/ci.yml/badge.svg)](https://github.com/foundry-rs/forge-std/actions/workflows/ci.yml)
+# 🏛️ Forge Standard Library &nbsp; [![CI status](https://github.com/foundry-rs/forge-std/actions/workflows/ci.yml/badge.svg)](https://github.com/foundry-rs/forge-std/actions/workflows/ci.yml)
 
-Forge Standard Library is a collection of helpful contracts and libraries for use with [Forge and Foundry](https://github.com/foundry-rs/foundry). It leverages Forge's cheatcodes to make writing tests easier and faster, while improving the UX of cheatcodes.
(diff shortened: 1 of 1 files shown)


mds1:
It looks like a lot of info and sample code was removed, is this upstreamed to the book yet? We should hold off on merging until it's upstreamed somewhere, otherwise it will be harder for users to find / harder to point them to

mds1:
I think it's useful to lean towards having too much information, so closing as stale for now, but happy to revisit an updated readme or reopen this when ready

ZeroEkkusu:
The current readme still suffers from the two issues solved here:

- The documentation in the book is out-of-date, and people aren't even aware that certain modules exist.
- Even less modules are showcased in the readme, making no mention of the others, giving the impression that's all Forge Std has to offer.

I suggest the following:

If this readme *looks* good to you, a `Read more` button can be added next to the sections that have more information. These descriptions will live after everything in the readme, and on the click of the button, the readme will scroll down to the header with the current information. It will also have the ↩️ for quickly navigating back.

Let me know what you think.

mds1:
Agreed that both the book documentation and forge-std README are missing a lot of information, and discoverability of things in forge-std has been a problem. Ideally we only need to maintain it in one place, which probably should be the book so everything users need is in one place. 

I'm not sure I follow how your suggestion solves the issues since it contains less information than currently, but I do see two possible plans for solving those issues, cc @Evalir:
1. First get the book updated to fully document forge-std, require book updates as part of PRs here going forward, then update the README to give a brief overview and link out to the book
2. Given https://github.com/foundry-rs/foundry/issues/3782, gradually improve docs by first migrating things from forge-std to forge and documenting them in the book at that time. Once fully migrated upstream, then update the README/book accordingly with what's left in forge-std

PR #425, opened 2023-07-19
fix: restore inheritance order

ZeroEkkusu:
## Motivation

https://github.com/foundry-rs/foundry/pull/5404

## Solution

- Remove warning note
- Restore inheritance order


mds1 (approved):
CI failure unrelated, https://github.com/foundry-rs/foundry/pull/5428 broke the `ignore` check which will be fixed upstream. Thanks!

PR #430, opened 2023-08-11
Fix assumeNotForgeAddress in StdCheats.sol

ashhanai:
`assumeNotForgeAddress` allowed any address to pass the assumption check. Unless `address(vm) == 0x000000000000000000636F6e736F6c652e6c6f67` (which is not), the condition always pass.

- Fixed by changing logical `or` to `and` in the condition. 
- Add check for Create2Deployer address

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -341,8 +341,8 @@ abstract contract StdCheatsSafe {
     }
 
     function assumeNotForgeAddress(address addr) internal pure virtual {
-        // vm and console addresses
-        vm.assume(addr != address(vm) && addr != 0x000000000000000000636F6e736F6c652e6c6f67);
+        // vm, console, and Create2Deployer addresses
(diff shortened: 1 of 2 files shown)


src/StdCheats.sol
```diff
@@ -341,8 +341,8 @@ abstract contract StdCheatsSafe {
     }
 
     function assumeNotForgeAddress(address addr) internal pure virtual {
-        // vm and console addresses
-        vm.assume(addr != address(vm) || addr != 0x000000000000000000636F6e736F6c652e6c6f67);
+        // vm, console, and Create2Deployer addresses
+        vm.assume(addr != address(vm) && addr != 0x000000000000000000636F6e736F6c652e6c6f67 && addr != 0x4e59b44847b379578588920cA78FbF26c0B4956C);
     }
```
mds1, line 346: Thanks! Do you mind adding a test for this also? You'll see existing tests like `testAssumeNotPrecompile` that you can base if on

mds1 (approved):
Thanks!

PR #433, opened 2023-08-13
chore: conform test naming to style guide

### LINKED ISSUE #273: Rename the test functions to adhere to the Foundry best practices
As suggested in the [Best Practices](https://book.getfoundry.sh/tutorials/best-practices) section of the Foundry Book, and as discussed in more detail [here](https://github.com/foundry-rs/book/issues/761), it is recommended to name the test functions like this:

- `test_Description`
- `testFuzz_Description`
- `test_Revert[If|When]_Description`
- `testFork_Description`
- `testForkFuzz_Revert[If|When]_Description`

Doing this would also make the test lintable by [scopelint](scopelint) when the tool is production-ready.

The tests in this repo do not, however, conform to this best practice yet.

Sabnock01:
This PR puts the `forge-std` test naming conventions in line with the [style guide](https://book.getfoundry.sh/tutorials/best-practices#tests)

Some liberties have been taken with respect to the names in `StdAssertions.t.sol` but open to any feedback.

I did not add any test names with the `Revert[If|When]` modifier since none of the tests appear to pass or fail conditionally. Correct me if this is erroneous. 

Closes #273

### DIFF
--- a/test/StdAssertions.t.sol
+++ b/test/StdAssertions.t.sol
@@ -422,13 +422,17 @@ contract StdAssertionsTest is Test {
                                     APPROX_EQ_ABS_DECIMAL(INT)
     //////////////////////////////////////////////////////////////////////////*/
 
-    function testFuzz_AssertApproxEqAbsDecimal_Int_Pass(int256 a, int256 b, uint256 maxDelta, uint256 decimals) external {
+    function testFuzz_AssertApproxEqAbsDecimal_Int_Pass(int256 a, int256 b, uint256 maxDelta, uint256 decimals)
+        external
+    {
         vm.assume(stdMath.delta(a, b) <= maxDelta);
 
         t._assertApproxEqAbsDecimal(a, b, maxDelta, decimals, EXPECT_PASS);
     }
 
-    function testFuzz_AssertApproxEqAbsDecimal_Int_Fail(int256 a, int256 b, uint256 maxDelta, uint256 decimals) external {
+    function testFuzz_AssertApproxEqAbsDecimal_Int_Fail(int256 a, int256 b, uint256 maxDelta, uint256 decimals)
+        external
+    {
         vm.assume(stdMath.delta(a, b) > maxDelta);
 
         vm.expectEmit(false, false, false, true);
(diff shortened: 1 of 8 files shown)


mds1:
> I did not add any test names with the `Revert[If|When]` modifier **since none of the tests appear to pass or fail conditionally.** 

Sorry what do you mean by this bolded part? Also just to clarify, the `Revert[If|When]` convention doesn't affect execution, it just signals you're testing that something should revert. This is useful because for any time you want to filter happy path vs. failure cases, such as excluding reverting tests when generating gas reports

test/StdAssertions.t.sol
```diff
@@ -31,21 +31,21 @@ contract StdAssertionsTest is Test {
                                     ASSERT_FALSE
     //////////////////////////////////////////////////////////////////////////*/
 
-    function testAssertFalse_Pass() external {
+    function test_AssertFalse() external {
         t._assertFalse(false, EXPECT_PASS);
     }
 
-    function testAssertFalse_Fail() external {
+    function testFail_AssertFalse() external {
```
mds1, line 38: `testFail` is a (deprecated, antipattern) prefix that tells forge a revert is expected, which is why the tests that were renamed to `testFail*` are failing. So we should rename these ones :)

mds1:
This is great, thanks @Sabnock01! Just about ready to merge, but want to bump the question in https://github.com/foundry-rs/forge-std/pull/433#issuecomment-1678224641 first

Sabnock01:
> This is great, thanks @Sabnock01! Just about ready to merge, but want to bump the question in https://github.com/foundry-rs/forge-std/pull/433#issuecomment-1678224641 first

Ah, no problem. I was trying to say that specific syntax didn't appear necessary for any of the current tests but I don't recall specifically why I worded it the way I did.

mds1:
Ah got it. Yea of course it’s not strictly necessary, but since we’re updating to follow the best practices doc conventions we may as well use that one too, for tests that are focused on checking revert conditions. Would you mind using that where applicable?

Sabnock01:
> Ah got it. Yea of course it’s not strictly necessary, but since we’re updating to follow the best practices doc conventions we may as well use that one too, for tests that are focused on checking revert conditions. Would you mind using that where applicable?

Sure, can do that.

mds1 (approved):
Thanks so much!

PR #434, opened 2023-08-14
chore(`StdCheats`): deprecate changePrank

Evalir:
Leftover work from https://github.com/foundry-rs/foundry/pull/4826#issuecomment-1543962477 — deprecates `changePrank` before we delete it completely.

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -4,6 +4,7 @@ pragma solidity >=0.6.2 <0.9.0;
 pragma experimental ABIEncoderV2;
 
 import {StdStorage, stdStorage} from "./StdStorage.sol";
+import {console2} from "./console2.sol";
 import {Vm} from "./Vm.sol";
 
 abstract contract StdCheatsSafe {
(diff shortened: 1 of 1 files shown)


src/StdCheats.sol
```diff
@@ -692,6 +693,7 @@ abstract contract StdCheats is StdCheatsSafe {
     }
 
     function changePrank(address msgSender) internal virtual {
+        console2.log("changePrank is deprecated. Please use vm.startPrank instead.");
```
mds1, line 696: Instead of importing console2 do you mind doing something this? It helps reduce compilation times in some cases, and compilation times are becoming a bottleneck for a lot of users :)  https://github.com/foundry-rs/forge-std/blob/c70ef65c9fe1c6255a83f6363f4c766fb67ba844/src/StdUtils.sol#L187-L197

Evalir: for sure! done in https://github.com/foundry-rs/forge-std/pull/434/commits/e6fae92e0adcc4c0308b90be9dba72b5b61db369

mds1 (approved):
Thanks!

PR #437, opened 2023-08-22
feat: base goerli and base chaindata setup to stdchains default rpcs

giuseppecrj:
Updates StdChains to have base and base goerli as default rpcs during initialization


mds1:
Can you also update the `testRpcs()` test to include this chain? That test is currently commented out since a lot of default RPCs are flaky, but might bring it back one day so would be good to keep it in sync (though to be fair there's probably other chains missing)

giuseppecrj:
@mds1 I added the tests for base and base_goerli as well as the missing ones for moonbeam and moonriver. Thanks!

mds1 (approved):
Thank you!

PR #439, opened 2023-08-23
feat: add a log helper library to easily log entire structs directly

DrakeEvans:
This is meant to be a placeholder draft PR to discuss a file which allows the logging of the entire struct.  

This adds a log(name: String) function to all the structs defined in forge-std so that they can easily be logged to the console during testing.

### DIFF
--- a/src/forgeStdLogHelper.sol
+++ b/src/forgeStdLogHelper.sol
@@ -1,0 +1,336 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.4.22 <0.9.0;
+
+import {console2 as console } from "./console2.sol";
+import {StdChains} from "./StdChains.sol";
+import {StdCheatsSafe} from "./StdCheats.sol";
+import {StdInvariant} from "./StdInvariant.sol";
+import {StdStorage} from "./StdStorage.sol";
+import {VmSafe} from "./Vm.sol";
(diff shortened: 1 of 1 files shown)


mds1:
This is interesting, and I like the idea. My main objection with adding this as-is to forge-std is it's impact on compile time. With via-ir, this library takes ~3s to compile for me:

```sh
$ hyperfine 'solc $(forge re) src/forgeStdLogHelper.sol --via-ir'
Benchmark 1: solc $(forge re) src/forgeStdLogHelper.sol --via-ir
  Time (mean ± σ):      3.007 s ±  0.079 s    [User: 2.838 s, System: 0.126 s]
  Range (min … max):    2.929 s …  3.209 s    10 runs
```

Compile times are becoming the speed bottleneck for devs these days, especially with via-ir (related: https://github.com/foundry-rs/forge-std/issues/225).

As a result, and in the interest of https://github.com/foundry-rs/foundry/issues/3782, I'd prefer to implement this (and more generally, all console.log functionality) as native cheats upstream in forge. We can add a lot of `vm.print()` signatures to the cheatcode interface and remove the need for all the solidity console.log libraries we currently have here. (logs have existing meaning in the EVM, hence why I'd name it `vm.print` instead of `vm.log`)

mds1:
Closing, and created https://github.com/foundry-rs/foundry/issues/7773 to track

PR #441, opened 2023-08-25
Update Vm.sol with exists, isFile and isDir

ruvaag:


ruvaag:
@Evalir PR for upstream changes, tagging for vis. There is no associated issue in this repo, but let me know if we need one for completeness.

src/Vm.sol
```diff
@@ -541,4 +541,10 @@ interface Vm is VmSafe {
     function transact(uint256 forkId, bytes32 txHash) external;
     // Suspends execution of the main thread for `duration` milliseconds
     function sleep(uint256 duration) external;
+    // Returns true if the given path points to an existing entity, else returns false
+    function exists(string calldata path) external returns (bool);
+    // Returns true if the path exists on disk and is pointing at a regular file, else returns false
+    function isFile(string calldata path) external returns (bool);
+    // Returns true if the path exists on disk and is pointing at a directory, else returns false
+    function isDir(string calldata path) external returns (bool);
```
Evalir, line 549: let's also name the return vars

Evalir (approved):
lgtm, albeit up to @mds1 for final review / approve! just a smol nit:

src/Vm.sol
ruvaag: Ack - fixed with reflexive named params for max clarity

mds1 (approved):
Thanks! I moved the new cheats from the `Vm` interface to the `VmSafe` interface next to the other filesystem cheats

PR #443, opened 2023-08-29
Update Vm.sol with tryFfi

jrcarlos2000:
ref : https://github.com/foundry-rs/foundry/blob/bff4ed912bb023d7bf9b20eda581aa4867a1cf70/testdata/cheats/Vm.sol#L139

### DIFF
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -54,6 +54,12 @@ interface VmSafe {
         uint256 privateKey;
     }
 
+    struct FfiResult {
+        int32 exit_code;
+        bytes stdout;
+        bytes stderr;
+    }
+
     // Derives a private key from the name, labels the account with that name, and returns the wallet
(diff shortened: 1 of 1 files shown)


src/Vm.sol
```diff
@@ -75,6 +81,8 @@ interface VmSafe {
     function getNonce(address account) external view returns (uint64 nonce);
     // Performs a foreign function call via the terminal
     function ffi(string[] calldata commandInput) external returns (bytes memory result);
+    // Performs a foreign function call via terminal and returns the exit code, stdout, and stderr
+    function tryFfi(string[] calldata) external returns (FfiResult memory);
```
mds1, line 85:
```suggestion
    function tryFfi(string[] calldata commandInput) external returns (FfiResult memory result);
```

mds1 (approved):
Thank you!

PR #445, opened 2023-08-29
feat: invariant target interfaces

bernard-wagner:
See https://github.com/foundry-rs/foundry/pull/5605


mds1:
Now that https://github.com/foundry-rs/foundry/pull/5605 is merged can this be taken out of draft, or does it need anything else?

mds1 (approved):
Thanks!

PR #446, opened 2023-08-29
fix(stdStorage): wrong slot id may returned by `find()` when slot value == 1337

paco0x:
When finding the target slot ID of a function signature, `find()` would do a staticcall and iterate over all the possible slots, then set the slot to a hardcoded value `hex"1337"` to test if the slot is desired.

But if there are multiple slot reads in the function call and the desired slot value is set to `hex"1337"` coincidentally, a wrong slot id may be returned.

Minimal reproduce contract:

```solidity
import "forge-std/Test.sol";
import "forge-std/StdStorage.sol";

contract StorageTest {
    bytes32 public a = hex"abcd";
    bytes32 public b = hex"1337";

    function extraSload() public view returns (bytes32) {
        // trigger read on slot `a`, and make a staticcall to make sure compiler doesn't optimize this SLOAD away
        assembly {
            pop(staticcall(gas(), sload(a.slot), 0, 0, 0, 0))
        }
        return b;
    }

    function noExtraSload() public view returns (bytes32) {
        return b;
    }
}

contract StdStorageTest is Test {

    function test_ExtraSload() public {
        StorageTest test = new StorageTest();
        // this will fail!
        assertEq(1, stdstore.target(address(test)).sig(test.extraSload.selector).find());
    }

    function test_NoExtraSload() public {
        StorageTest test = new StorageTest();
        assertEq(1, stdstore.target(address(test)).sig(test.noExtraSload.selector).find());
    }
}
```

Test result:
```bash
❯ forge test -vv --mp test/StdStorageTest.sol
[⠆] Compiling...
No files changed, compilation skipped

Running 2 tests for test/StdStorageTest.sol:StdStorageTest
[FAIL. Reason: Assertion failed.] test_ExtraSload() (gas: 201767)
Logs:
  Error: a == b not satisfied [uint]
        Left: 1
       Right: 0

[PASS] test_NoExtraSload() (gas: 192642)
Test result: FAILED. 1 passed; 1 failed; 0 skipped; finished in 3.54ms
Ran 1 test suites: 1 tests passed, 1 failed, 0 skipped (2 total tests)

Failing tests:
Encountered 1 failing test in test/StdStorageTest.sol:StdStorageTest
[FAIL. Reason: Assertion failed.] test_ExtraSload() (gas: 201767)

Encountered a total of 1 failing tests, 1 tests succeeded
```

The `test_ExtraSload()` failed to return the correct slot ID in test.

This patch fixes the issue by replacing the hardcoded new value with bitwise not, prevent from conflicting with the target slot value.

### DIFF
--- a/src/StdStorage.sol
+++ b/src/StdStorage.sol
@@ -71,9 +71,9 @@ library stdStorageSafe {
                 if (prev != fdat) {
                     continue;
                 }
-                bytes32 newVal = ~prev;
+                bytes32 new_val = ~prev;
                 // store
-                vm.store(who, reads[i], newVal);
+                vm.store(who, reads[i], new_val);
                 bool success;
                 {
                     bytes memory rdat;
@@ -81,7 +81,7 @@ library stdStorageSafe {
                     fdat = bytesToBytes32(rdat, 32 * field_depth);
                 }
 
-                if (success && fdat == newVal) {
+                if (success && fdat == new_val) {
                     // we found which of the slots is the actual one
                     emit SlotFound(who, fsig, keccak256(abi.encodePacked(ins, field_depth)), uint256(reads[i]));
                     self.slots[who][fsig][keccak256(abi.encodePacked(ins, field_depth))] = uint256(reads[i]);
--- a/src/StdStorage.sol
+++ b/src/StdStorage.sol
@@ -68,16 +68,21 @@ library stdStorageSafe {
                 if (prev == bytes32(0)) {
                     emit WARNING_UninitedSlot(who, uint256(reads[i]));
                 }
+                if (prev != fdat) {
+                    continue;
+                }
+                bytes32 newVal = bytes32(~uint256(prev));
                 // store
-                vm.store(who, reads[i], bytes32(hex"1337"));
+                vm.store(who, reads[i], newVal);
                 bool success;
-                bytes memory rdat;
+                bytes32 dat;
                 {
+                    bytes memory rdat;
                     (success, rdat) = who.staticcall(cald);
-                    fdat = bytesToBytes32(rdat, 32 * field_depth);
+                    dat = bytesToBytes32(rdat, 32 * field_depth);
                 }
 
-                if (success && fdat == bytes32(hex"1337")) {
+                if (success && dat == newVal) {
                     // we found which of the slots is the actual one
                     emit SlotFound(who, fsig, keccak256(abi.encodePacked(ins, field_depth)), uint256(reads[i]));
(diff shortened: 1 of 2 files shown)


src/StdStorage.sol
```diff
@@ -68,16 +68,21 @@ library stdStorageSafe {
                 if (prev == bytes32(0)) {
                     emit WARNING_UninitedSlot(who, uint256(reads[i]));
                 }
+                if (prev != fdat) {
+                    continue;
+                }
```
mds1, line 73: Why do we need to add this check? (I'm not too familiar with the internals of this portion)

src/StdStorage.sol
```diff
@@ -68,16 +68,21 @@ library stdStorageSafe {
...
+                bytes32 newVal = bytes32(~uint256(prev));
```
mds1, line 74: I think the casts are unnecessary here
```suggestion
                bytes32 newVal = ~prev;
```

src/StdStorage.sol
```diff
@@ -68,16 +68,21 @@ library stdStorageSafe {
...
                 // store
-                vm.store(who, reads[i], bytes32(hex"1337"));
+                vm.store(who, reads[i], newVal);
                 bool success;
-                bytes memory rdat;
+                bytes32 dat;
```
mds1, line 78: Wondering if these changes were required for any reason too? Based on the PR description I'd have thought the only diff is change from testing `hex"1337"` to testing `~prev`

paco0x: I was thinking of a very edge case that the `fdat` variable would be changed to something new but not equal to `~prev`, and the following steps in the loop would be affected since there's a `prev ~= fdat` check.

But it'll be a very weird case if this happens (function return value depends on multiple slot values), I'll just revert this back.

src/StdStorage.sol
paco0x: When finding the target slot, we should ensure the slot value == function return value, just as we did in the first if block. This could reduce some unnecessary SSTOREs and checks in the loop.

Another reason is that only checking the slot value after modifying it to a new value isn't enough. Consider this case:

```
contract Test {
    bytes public a  = hex"1";
    bytes public b = ~a;
}
```

If we're finding the slot ID of b and we're iterating through `[a, b]`, setting `a` to `~a` would result in the equation to slot value b (function return value). And `find()` would just return the slot ID of a instead of b. Require `prev == fdat` could help to correct it in this case.

brockelmore:
Nice! lgtm, but i haven’t wrapped my head around if there are any other edge cases for inverting the read value instead of using specific value. But my intuition is that this is a strict improvement and likely edge cases are only reduced by this.

mds1 (approved):
Thank you!

PR #447, opened 2023-08-29
chore: make json parse cheatcodes pure

vdrg:
Hi folks, I believe the json parsing cheatcodes classify as pure as they are not affected by other calls and don't read the state. Please let me know if I'm missing anything.


mds1 (approved):
Thank you!

PR #458, opened 2023-09-25
chore: camelCase for exitCode

PaulRBerg:
Pardon my OCD but I am getting a Solhint warning in my copy of `Vm` (in PRBTest) because of `exit_code`:

> warning  Variable name must be in mixedCase  var-name-mixedcase

### DIFF
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -55,7 +55,7 @@ interface VmSafe {
     }
 
     struct FfiResult {
-        int32 exit_code;
+        int32 exitCode;
         bytes stdout;
         bytes stderr;
     }


mds1:
Ah, should have caught this when it was originally merged. This is technically a breaking change since it's already in a release, but the release was recent and from a [sourcegraph query](https://sourcegraph.com/search?q=context:global+tryFfi+lang:Solidity+-repo:%5Egithub%5C.com/foundry-rs/foundry%24+-repo:%5Egithub%5C.com/foundry-rs/forge-std%24+&patternType=standard&sm=0&groupBy=repo) it seems vulcan are the only users (clearly that search isn't perfect since PRBTest isn't there).

Given that, we can sneak this into a v1.7.0 once @vdrg and @gnkz from vulcan acknowledge

vdrg:
@mds1 Thanks for letting us know! Acknowledged, we'll update forge-std once this is merged.

mds1:
Great, planning to merge+release early next week since I'm out this week—will keep you updated :)

PaulRBerg:
Also happy to update PRBTest to include the fix. I don't mind the breaking change since I've only added `exit_code` this week.

mds1 (approved):

PR #460, opened 2023-09-25
perf: mark parseJsonKeys as pure

PaulRBerg:
Same as https://github.com/foundry-rs/forge-std/pull/447 for `parseJsonKeys`.


mds1 (approved):
Thanks!

Issue #463, opened 2023-09-29
chore: ensure logical and consistent groupings of cheats in `Vm.sol`


mds1:
Ref https://github.com/foundry-rs/forge-std/issues/459

In general, `Vm.sol` is grouped logically (e.g. all filesystem cheats in one place), but we should:
1. Make sure all groupings are correct and fix any out-of-place cheats
2. Add comments to more clearly separate sections so it's more clear

ruvaag:
I can take this up

mds1:
Thanks @ruvaag, though I already opened a PR for this in https://github.com/foundry-rs/forge-std/pull/464 :)

PR #464, opened 2023-09-29
chore: reorder Vm for logical consistency

mds1:
Closes https://github.com/foundry-rs/forge-std/issues/463

@PaulRBerg would appreciate a quick review here if you have the time! No rush though, I also need to give it a review myself before merging

cc @vdrg too since you're also big consumers of Vm, in case this affects you in any way (though it shouldn't affect you)

### DIFF
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -90,13 +90,17 @@ interface VmSafe {
 
     // Starts recording all map SSTOREs for later retrieval.
     function startMappingRecording() external;
+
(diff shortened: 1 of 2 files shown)


mds1:
Merging to unblock #465, but feedback welcome which we can fix in a follow up PR

PaulRBerg:
Had a quick now and this is definitely an improvement over the previous version. Thanks @mds1.

Issue #469, opened 2023-10-11
Provide mock ERC-20 and ERC-721 contracts


PaulRBerg:
In OpenZeppelin v5, they have made the `ERC20` and` ERC721` contracts abstract:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/793d92a3331538d126033cbacb1ee5b8a7d95adc/contracts/token/ERC20/ERC20.sol#L34

This means that for testing purposes we all need to define a dummy ERC-20 contract like this:

```solidity
// SPDX-License-Identifier: GPL-3.0-or-later
pragma solidity >=0.8.20;

import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";

contract ERC20Mock is ERC20 {
    constructor(string memory name, string memory symbol) ERC20(name, symbol) { }
}
```

The same requirement applies to ERC-721.

It would be helpful if Forge Std provided these mocks for users so they don't end up re-implemented all over GitHub.

andreivladbrg:
I can work on this, but I don't think it would be a good decision for `forge-std` to have `openzeppelin` installed as a dependency.

The only way to make it work is to copy and paste the `ERC20` and `ERC721` contracts into a `mock` dir.

wdyt?

PaulRBerg:
Great point @andreivladbrg; let's see what @mds1 and other Forge Std contributors have to say.

mds1:
This seems like a good idea. Agreed with the requirement of not adding and dependencies to forge-std.

Similar to the interfaces directory, these should not be imported anywhere by default to avoid impacting compilation times, so they can live as standalone contracts in `src/mocks/`

One consideration is the tokens added should support solidity 0.6.2 through 0.8.x. This means we'll have to implement checked math in the token directly, so we likely just can't copy an existing token out of the box. The easiest is probably to copy solmate's because it's a flat file and easy to read, then edit it to add checked math. The file should have comments indicating it's not for use in production, and include a comment with permalink to the original source it was based on.

What devex do you think this better here?
1. Include permissionless `mint(address to, uint256 amount)` and `burn(address from, uint256 amount)` methods to make it clear it's a mock
2. Keep it as a standard token with no mint/burn functionality and use `deal` to set balances. With this approach, users will likely want to use the `deal(address token, address to, uint256 give, bool adjust)` version with the last arg as `true` to ensure total supply is updated, otherwise it will stay at zero

I think I lean towards (1) but curious to hear opinions.

[...]

PR #470, opened 2023-10-17
feat: add mock contracts for ERC20 and ERC721

andreivladbrg:
Closes https://github.com/foundry-rs/forge-std/issues/469

### DIFF
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -4,6 +4,8 @@ pragma solidity >=0.6.2 <0.9.0;
 pragma experimental ABIEncoderV2;
 
 import {IMulticall3} from "./interfaces/IMulticall3.sol";
+import {ERC20Mock} from "./mocks/ERC20Mock.sol";
+import {ERC721Mock} from "./mocks/ERC721Mock.sol";
 import {VmSafe} from "./Vm.sol";
 
 abstract contract StdUtils {
@@ -131,6 +133,21 @@ abstract contract StdUtils {
     /// @dev returns the address of a contract created with CREATE2 using the default CREATE2 deployer
     function computeCreate2Address(bytes32 salt, bytes32 initCodeHash) internal pure returns (address) {
         return computeCreate2Address(salt, initCodeHash, CREATE2_FACTORY);
+    }
+
+    /// @dev returns a initilized mock ERC20 contract
+    function deployERC20Mock(string memory name, string memory symbol, uint8 decimals)
+        internal
+        returns (ERC20Mock mock)
+    {
+        mock = new ERC20Mock();
+        mock.initialize(name, symbol, decimals);
+    }
+
+    /// @dev returns a initilized mock ERC721 contract
+    function deployERC721Mock(string memory name, string memory symbol) internal returns (ERC721Mock mock) {
+        mock = new ERC721Mock();
+        mock.initialize(name, symbol);
     }
 
     /// @dev returns the hash of the init code (creation code + no args) used in CREATE2 with no constructor arguments
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -135,7 +135,7 @@ abstract contract StdUtils {
         return computeCreate2Address(salt, initCodeHash, CREATE2_FACTORY);
     }
 
-    /// @dev returns an initilized mock ERC20 contract
+    /// @dev returns an initialized mock ERC20 contract
     function deployMockERC20(string memory name, string memory symbol, uint8 decimals)
         internal
         returns (MockERC20 mock)
@@ -144,7 +144,7 @@ abstract contract StdUtils {
         mock.initialize(name, symbol, decimals);
     }
 
-    /// @dev returns an initilized mock ERC721 contract
+    /// @dev returns an initialized mock ERC721 contract
     function deployMockERC721(string memory name, string memory symbol) internal returns (MockERC721 mock) {
         mock = new MockERC721();
         mock.initialize(name, symbol);
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -4,8 +4,8 @@ pragma solidity >=0.6.2 <0.9.0;
 pragma experimental ABIEncoderV2;
 
 import {IMulticall3} from "./interfaces/IMulticall3.sol";
-import {ERC20Mock} from "./mocks/ERC20Mock.sol";
-import {ERC721Mock} from "./mocks/ERC721Mock.sol";
+import {MockERC20} from "./mocks/MockERC20.sol";
+import {MockERC721} from "./mocks/MockERC721.sol";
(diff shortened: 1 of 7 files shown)


mds1:
Thanks for this! Only skimmed the code but lgtm so far, will take a closer look once we get CI passing.

It seems we'll need to make a few changes so it compiles with older versions, see the run [here](https://github.com/foundry-rs/forge-std/actions/runs/6552104405/job/17986006102?pr=470) or the screenshot below. The immutable fix is easy, we can just remove that

The chain ID is a bit trickier but we found a good way to do that in `StdCheats.sol`, which you can copy. In the code you'd call the `_pureChainId()` method: https://github.com/Tudmotu/forge-std/blob/5dcccff9d9fd2ea1b2ef540308313ed31377a982/src/StdCheats.sol#L615-L636


For the constructor, I'm not sure if there's a solution offhand: We want to make sure there's no warnings regardless of the solc version used (you can of course test this easily by changing the `solc_version` in the config file), and my guess is that adding the `public` visibility modifier to the constructor for 0.6.12 will result in a warning for later solc versions. @mattsse if we used `ignored_error_codes` in forge-std's `foundry.toml` is that respected in upstream repos without affecting them? In other words, if we silence warning code 4937 in forge-std, does that both (1) hide forge-std's emission of that warning from upstream users, and (2) not hide upstream user's own emission of that warning? If so, that might be a suitable solution. If not, we might have to have two versions of each contract

image

andreivladbrg:
@mds1, just wanted to let you know that I've pushed a commit to fix the incompatibilities with older solc versions (there were more 😁).

Regarding the constructor, it's indeed a tricky one. I'm not entirely sure how to address it yet. For now, I've set the `public` modifier, but there is a warning when compiling (tho it runs).

Another idea, aside from having two versions of the contracts, is to remove the constructor. We could then hard-code the metadata to default values like "Mock ERC20", "MOCK", decimals to 18, and add a `changeMetadata` method that allows adjustments to these values later on.

wdyt?

mds1:
Here's a constructor workaround we used to use in forge-std which should work here

```solidity
/// @dev To hide constructor warnings across solc versions due to different constructor visibility requirements and
/// syntaxes, we put the constructor in a private method and assign an unused return value to a variable. This
/// forces the method to run during construction, but without declaring an explicit constructor.
uint256 private CONSTRUCTOR = _constructor();

function _constructor() private returns (uint256) {
  // logic here
  return 0;
}
```

andreivladbrg:
> Here's a constructor workaround we used to use in forge-std which should work here

It is a smart way of doing this, but I don't think it is entirely useful here, given that we also need to pass arguments in the constructor (name, symbol).

>remove the constructor and have an initialize method that can only be called once, like proxies use. Then in StdUtils we can add deployMockERC20 and deployMockERC721 helper functions to simplify the deploy + initialization process.

I ended up doing this way.

src/StdUtils.sol
```diff
@@ -133,6 +135,21 @@ abstract contract StdUtils {
         return computeCreate2Address(salt, initCodeHash, CREATE2_FACTORY);
     }
 
+    /// @dev returns a initilized mock ERC20 contract
```
mds1, line 138:
```suggestion
    /// @dev returns an initialized mock ERC20 contract
```

src/StdUtils.sol
```diff
@@ -133,6 +135,21 @@ abstract contract StdUtils {
...
+    function deployERC20Mock(string memory name, string memory symbol, uint8 decimals)
+        internal
+        returns (ERC20Mock mock)
+    {
+        mock = new ERC20Mock();
+        mock.initialize(name, symbol, decimals);
+    }
+
+    /// @dev returns a initilized mock ERC721 contract
```
mds1, line 147:
```suggestion
    /// @dev returns an initialized mock ERC721 contract
```

src/StdUtils.sol
```diff
@@ -133,6 +135,21 @@ abstract contract StdUtils {
         return computeCreate2Address(salt, initCodeHash, CREATE2_FACTORY);
     }
 
+    /// @dev returns a initilized mock ERC20 contract
+    function deployERC20Mock(string memory name, string memory symbol, uint8 decimals)
+        internal
+        returns (ERC20Mock mock)
+    {
+        mock = new ERC20Mock();
```
mds1, line 143: Let's rename everything to put the "Mock" part as a prefix, e.g. `MockERC20` and `MockERC721`—this is the naming convention solmate uses, and it's my preference as well. For example:
```suggestion
    function deployMockERC20(string memory name, string memory symbol, uint8 decimals)
        internal
        returns (MockERC20 mock)
    {
        mock = new MockERC20();
```

src/mocks/ERC20Mock.sol
```diff
@@ -0,0 +1,216 @@
...
+
+    /*//////////////////////////////////////////////////////////////
+                               INITIALIZE
+    //////////////////////////////////////////////////////////////*/
+
+    /// @dev A bool to track whether the contract has been initialized.
+    bool private initialized;
+
+    /// @dev To hide constructor warnings across solc versions due to different constructor visibility requirements and
+    /// syntaxes, we add a initialization function that can be called only once.
```
mds1, line 53:
```suggestion
    /// syntaxes, we add an initialization function that can be called only once.
```

src/mocks/ERC721Mock.sol
```diff
@@ -0,0 +1,221 @@
...
+
+    /*//////////////////////////////////////////////////////////////
+                               INITIALIZE
+    //////////////////////////////////////////////////////////////*/
+
+    /// @dev A bool to track whether the contract has been initialized.
+    bool private initialized;
+
+    /// @dev To hide constructor warnings across solc versions due to different constructor visibility requirements and
+    /// syntaxes, we add a initialization function that can be called only once.
```
mds1, line 61:
```suggestion
    /// syntaxes, we add an initialization function that can be called only once.
```

src/mocks/ERC20Mock.sol
mds1, line 1: Would you mind adding the test suites from Solmate also and using those here? They're pretty thorough, and from a code review of your contracts it seems to correct, but having tests would add confidence + protect us from introducing any bugs in future PRs/changes to this code

mds1 (commented):
Good call about the constructor, `initialize` approach used here LGTM

src/mocks/ERC20Mock.sol
andreivladbrg: > Would you mind adding the test suites from Solmate also and using those here 

Sure, just added them

> protect us from introducing any bugs in future PRs/changes to this code

Correct

andreivladbrg:
@mds1 Addressed all your comments, lmk if there is anything else

mds1 (approved):
This is great, thank you!

Would you mind adding a PR to the book's forge standard library reference documenting these mocks and deploy utils?

PR #473, opened 2023-10-31
fix: rely less on foundry internals

DaniPopes:
Rely less on internal Foundry error messages for our own testing.
Unfortunately it looks like we still rely on the "RPC URL not found" error message to when defaulting to environment variables in `StdChains.getChainWithUpdatedRpcUrl`.

Unblocks https://github.com/foundry-rs/foundry/pull/6131

cc @mattsse @mds1

### DIFF
--- a/src/StdChains.sol
+++ b/src/StdChains.sol
@@ -160,9 +160,17 @@ abstract contract StdChains {
                     chain.rpcUrl = vm.envString(envName);
                 }
                 // distinguish 'not found' from 'cannot read'
-                bytes memory notFoundError =
+                bytes memory oldNotFoundError =
                     abi.encodeWithSignature("CheatCodeError", string(abi.encodePacked("invalid rpc url ", chainAlias)));
-                if (keccak256(notFoundError) != keccak256(err) || bytes(chain.rpcUrl).length == 0) {
+                // TODO: use `abi.encodeWithSelector(Vm.CheatcodeError)` on next Vm update
+                bytes memory newNotFoundError = abi.encodeWithSignature(
+                    "CheatcodeError(string)", string(abi.encodePacked("invalid rpc url: ", chainAlias))
+                );
+                bytes32 errHash = keccak256(err);
+                if (
+                    (errHash != keccak256(oldNotFoundError) && errHash != keccak256(newNotFoundError))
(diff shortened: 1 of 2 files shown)


mattsse (approved):
I think this is reasonable

src/StdChains.sol
```diff
@@ -160,9 +160,17 @@ abstract contract StdChains {
                     chain.rpcUrl = vm.envString(envName);
                 }
                 // distinguish 'not found' from 'cannot read'
-                bytes memory notFoundError =
+                bytes memory oldNotFoundError =
                     abi.encodeWithSignature("CheatCodeError", string(abi.encodePacked("invalid rpc url ", chainAlias)));
-                if (keccak256(notFoundError) != keccak256(err) || bytes(chain.rpcUrl).length == 0) {
+                // TODO: use `abi.encodeWithSelector(Vm.CheatcodeError)` on next Vm update
```
mds1, line 165: Can we resolve this TODO on this PR or is something blocking it?

src/StdChains.sol
```diff
@@ -160,9 +160,17 @@ abstract contract StdChains {
                     chain.rpcUrl = vm.envString(envName);
                 }
                 // distinguish 'not found' from 'cannot read'
-                bytes memory notFoundError =
+                bytes memory oldNotFoundError =
```
mds1, line 163:
```suggestion
                // Distinguish 'not found' from 'cannot read'
                // The upstream error thrown by forge for failing cheats changed so we check both the old and new versions
                bytes memory oldNotFoundError =
```

src/StdChains.sol
DaniPopes: would need to add the error type to Vm

mds1: Is something upstream blocking that addition or can we make that change in this PR? One potential issue is I'm not sure offhand is solidity 0.6.2 supports that syntax

DaniPopes: Yeah that's also a problem, custom errors are in 0.8.4. I guess this is fine for now

gakonst (approved):
good w me

src/StdChains.sol
mds1: Yea I think this is ok, thanks for updating

PR #474, opened 2023-11-07
vm: add `loadAllocs(string)` cheatcode

tynes:
Adds the `loadAllocs(string)` cheatcode to be usable as part of the
`Vm` interface. This was added to foundry in https://github.com/foundry-rs/foundry/pull/6207.
The cheatcode can be used to read a JSON file from disk that sets EVM
state directly.


mds1 (approved):
Thank you!

PR #479, opened 2023-11-15
feat: update `computeCreateAddress`

qiweiii:
As mentioned in https://github.com/foundry-rs/foundry/issues/6270#issuecomment-1810429601, updating Vm.sol and use vm call in `StdUtils`.

### DIFF
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -174,7 +174,7 @@ abstract contract StdUtils {
         (bool status,) = address(CONSOLE2_ADDRESS).staticcall(abi.encodeWithSignature("log(string)", p0));
         status;
     }
-    
+
     function console2_log_StdUtils(string memory p0, uint256 p1) private view {
         (bool status,) = address(CONSOLE2_ADDRESS).staticcall(abi.encodeWithSignature("log(string,uint256)", p0, p1));
         status;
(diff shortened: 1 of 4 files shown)


src/StdUtils.sol
```diff
@@ -186,6 +170,11 @@ abstract contract StdUtils {
 
     // Used to prevent the compilation of console, which shortens the compilation time when console is not used elsewhere.
 
+    function console2_log(string memory p0) private view {
```
mds1, line 173: Looks like this fails to compile because `StdCheats.sol` has a function with the same signature, so it clashes when both this contract and `StdCheats` are inherited from in `Test.sol`. Perhaps we namespace all of these internal `console2_log` functions, e.g. this one becomes `function console2_log_StdUtils(string memory p0) private view`. Open to other suggestions too

qiweiii: sgtm

mds1:
Thank you! I think this should be good to go now, I'll give it a sanity check and merge on Monday :)

dkuppitz:
Has this been tested in the wild? This update broke all my deployment scripts. The new implementation just reverts with "Invalid data".

```
    ├─ [0] console::log(computeCreate2Address is deprecated. Please use vm.computeCreate2Address instead.) [staticcall]
    │   └─ ← ()
    ├─ [0] VM::890c283b(61e8d2e604d55934ecf61dc1730de3018a02b583ef76d80409b3a57b66cfe0582be6702d5f2862666ef322dad9569a07c0baf85b4c8deff2e9db92634def1a1f) [staticcall]
    │   └─ ← 0x0bc44503...
    └─ ← "Invalid data"
    ```

dkuppitz:
Nvm, I didn't check the actual changes. Realized just now that I had to do a `foundryup` to get it working.

PR #490, opened 2024-01-10
Mark envOr functions as view

anikaraghu:
I tried using `vm.envOr` in a view function and noticed it wasn't marked as `view`. Checked with foundry devs and looks like this was just an oversight.

Fix a couple compile warnings that came up as a result

### DIFF
--- a/src/StdChains.sol
+++ b/src/StdChains.sol
@@ -148,7 +148,7 @@ abstract contract StdChains {
 
     // lookup rpcUrl, in descending order of priority:
     // current -> config (foundry.toml) -> environment variable -> default
-    function getChainWithUpdatedRpcUrl(string memory chainAlias, Chain memory chain) private returns (Chain memory) {
+    function getChainWithUpdatedRpcUrl(string memory chainAlias, Chain memory chain) private view returns (Chain memory) {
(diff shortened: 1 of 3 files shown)


mds1:
Apologies for the churn here, pasting my response from our other chat for reference. Will close this one as a result, since there should be an upstream foundry PR first

> ah, sorry I hinted at this above but wasn't explicit. The source of truth for cheatcodes is now in [this foundry file](https://github.com/foundry-rs/foundry/blob/master/crates/cheatcodes/spec/src/vm.rs), and we autogenerate the forge-std Vm.sol file from that with [this script](https://github.com/foundry-rs/forge-std/blob/master/scripts/vm.py)
> 
> so the right approach [here](https://github.com/foundry-rs/foundry/blob/68c3663e0dd0261bbb14284b492ce0b0e703006b/crates/cheatcodes/spec/src/vm.rs#L894-L978) is add the view modifier into the solidity definitions here, then once that is merged in foundry, run that python script to open the corresponding forge-std PR.
>
> some more docs on this here: https://github.com/foundry-rs/foundry/blob/68c3663e0dd0261bbb14284b492ce0b0e703006b/docs/dev/cheatcodes.md#cheatcodes-implementation

PR #496, opened 2024-01-16
Add signP256 cheatcode

wilsoncusack:
Follow up to https://github.com/foundry-rs/foundry/pull/6797, adds function definition for signP256 to Vm.sol

### DIFF
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -418,8 +418,11 @@ interface VmSafe {
     /// Performs an Ethereum JSON-RPC request to the current fork URL.
     function rpc(string calldata method, string calldata params) external returns (bytes memory data);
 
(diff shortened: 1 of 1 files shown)


wilsoncusack:
@mds1 
image

mds1:
@DaniPopes I'm looking at https://github.com/foundry-rs/foundry/pull/6785 and it aded a new group called `Base64`. Feels like overkill for that to have it's own group. What do you think about removing that group and moving the base64 cheats to the `Utilities` group?

wilsoncusack:
Does it make sense to leave the byes64 issue to a separate PR?

mds1:
Replaced by https://github.com/foundry-rs/forge-std/pull/497

PR #498, opened 2024-01-17
fix: handle USDT blacklisting address(0) and address(1) edge case

zerosnacks:
After a [failing test](https://github.com/foundry-rs/foundry/actions/runs/7547585011/job/20547871860#step:12:883) in Foundry I noticed USDT has blacklisted [address(0)](https://phalcon.blocksec.com/explorer/tx/eth/0x26e2265c66d7de39bdf0570a74118bbeec782ea11b10b9b075f3ef3866d70722) and [address(1)](https://phalcon.blocksec.com/explorer/tx/eth/0xd8ea4a26e5354f8660df8dd24bc46b7f260865cbc5f9e5069665934a0265d6d8) (unclear why), causing fuzz tests with edge bias to revert.

To reduce the likelihood of failing tests I suggest skipping `address(0)` and the precompiles.

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -206,7 +206,10 @@ abstract contract StdCheatsSafe {
     }
 
     // Checks that `addr` is not blacklisted by token contracts that have a blacklist.
-    function assumeNotBlacklisted(address token, address addr) internal view virtual {
+    function assumeNotBlacklisted(address token, address addr) internal virtual {
+        // `USDT` blacklists `address(0)` and `address(1)` for unknown reasons.
+        assumeAddressIsNot(addr, AddressType.ZeroAddress, AddressType.Precompile, AddressType.ForgeAddress);
+
         // Nothing to check if `token` is not a contract.
         uint256 tokenCodeSize;
         assembly {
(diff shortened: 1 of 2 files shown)


mds1:
Not sure I understand the fix—if those addresses are blacklisted by USDT, wouldn't they be caught by the existing call to `isBlackListed(address)`?
```solidity
// 4-byte selector for `isBlackListed(address)`, used by USDT.
(success, returnData) = token.staticcall(abi.encodeWithSelector(0xe47d6060, addr));
vm.assume(!success || abi.decode(returnData, (bool)) == false);
```

zerosnacks:
> Not sure I understand the fix—if those addresses are blacklisted by USDT, wouldn't they be caught by the existing call to `isBlackListed(address)`?
> 
> ```solidity
> 
> // 4-byte selector for `isBlackListed(address)`, used by USDT.
> 
> (success, returnData) = token.staticcall(abi.encodeWithSelector(0xe47d6060, addr));
> 
> vm.assume(!success || abi.decode(returnData, (bool)) == false);
> 
> ```

You are correct, I misinterpreted the purpose of the test. The fuzz test is expected to throw when encountering a blacklisted address, which it does.

It makes sense to test the implementation like this but it does lead to random tests failing, especially because of the edge bias.

I'll close this PR.

PR #510, opened 2024-02-08
Inherit interfaces in mock tokens

arr00:
Currently `MockERC20` and `MockERC721` don't inherit their respective interfaces. This PR fixes that.

### DIFF
--- a/src/mocks/MockERC20.sol
+++ b/src/mocks/MockERC20.sol
@@ -1,17 +1,11 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.2 <0.9.0;
 
+import {IERC20} from "../interfaces/IERC20.sol";
+
 /// @notice This is a mock contract of the ERC20 standard for testing purposes only, it SHOULD NOT be used in production.
 /// @dev Forked from: https://github.com/transmissions11/solmate/blob/0384dbaaa4fcb5715738a9254a7c0a4cb62cf458/src/tokens/ERC20.sol
-contract MockERC20 {
-    /*//////////////////////////////////////////////////////////////
-                                 EVENTS
-    //////////////////////////////////////////////////////////////*/
-
-    event Transfer(address indexed from, address indexed to, uint256 amount);
-
-    event Approval(address indexed owner, address indexed spender, uint256 amount);
-
+contract MockERC20 is IERC20 {
     /*//////////////////////////////////////////////////////////////
                             METADATA STORAGE
     //////////////////////////////////////////////////////////////*/
--- a/src/mocks/MockERC721.sol
+++ b/src/mocks/MockERC721.sol
@@ -1,19 +1,11 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.2 <0.9.0;
 
+import {IERC721} from "../interfaces/IERC721.sol";
+
 /// @notice This is a mock contract of the ERC721 standard for testing purposes only, it SHOULD NOT be used in production.
 /// @dev Forked from: https://github.com/transmissions11/solmate/blob/0384dbaaa4fcb5715738a9254a7c0a4cb62cf458/src/tokens/ERC721.sol
-contract MockERC721 {
-    /*//////////////////////////////////////////////////////////////
-                                 EVENTS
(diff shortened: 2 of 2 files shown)


mds1:
Thanks for your PR :) 

What is the benefit of having them inherit from the interfaces? This feels like a nontrivial diff that adds code for unclear benefit, since these ERC interfaces are finalized and static, and the mocks are already compliant. I'd like to keep everything in forge-std as simple as possible so it's easy to read/maintain and to reduce compilation times.

I'm also not a fan of making the variables private since that limits devs from inheriting the mock and adding e.g. new methods to set approvals

arr00:
It is unintuitive for there to be an interface as well as an implementation of that interface in the same repo but them not be tied together at all. I came across this when using the interface and the mock token in a situation shown below:

```
IERC20 fromToken;
fromToken = IERC20(address(new MockERC20()));
```

mds1:
> It is highly preferred for the below to work:
>
> ```
> IERC20 fromToken;
> fromToken = new MockERC20();
> ```

That would be really nice but I don't think that is supported since solidity doesn't have structural typing: https://forum.soliditylang.org/t/making-interfaces-structural/1790

```solidity
import {MockERC20} from "../src/mocks/MockERC20.sol";
import {IERC20} from "../src/interfaces/IERC20.sol";
import {Test} from "../src/Test.sol";

contract ATest is Test {
    IERC20 public token;

    function setUp() public {
      IERC20 fromToken;
      fromToken = new MockERC20();
    }
}
```

```
$ forge build
[⠢] Compiling...
[⠊] Compiling 26 files with 0.8.23
[⠒] Solc 0.8.23 finished in 648.53ms
Error: 
Compiler run failed:
Warning (3420): Source file does not specify required compiler version! Consider adding "pragma solidity ^0.8.23;"
--> test/A.t.sol

Error (7407): Type contract MockERC20 is not implicitly convertible to expected type contract IERC20.
  --> test/A.t.sol:19:19:
   |
19 |       fromToken = new MockERC20();
   |                   ^^^^^^^^^^^^^^^
```

arr00:
@mds1 Consider running that test in the context of that PR. It works.

src/mocks/MockERC20.sol
```diff
@@ -1,36 +1,54 @@
...
 /// @dev Forked from: https://github.com/transmissions11/solmate/blob/0384dbaaa4fcb5715738a9254a7c0a4cb62cf458/src/tokens/ERC20.sol
-contract MockERC20 {
+contract MockERC20 is IERC20 {
     /*//////////////////////////////////////////////////////////////
-                                 EVENTS
+                            METADATA STORAGE
     //////////////////////////////////////////////////////////////*/
 
-    event Transfer(address indexed from, address indexed to, uint256 amount);
+    string private _name;
```
mds1, line 13: Let's change all `private` vars to `internal` so devs have more control when inheriting from this contract and making custom mock tokens

src/mocks/MockERC20.sol
```diff
@@ -1,36 +1,54 @@
...
                               ERC20 STORAGE
     //////////////////////////////////////////////////////////////*/
 
-    uint256 public totalSupply;
+    uint256 private _totalSupply;
+
+    mapping(address => uint256) private _balances;
 
-    mapping(address => uint256) public balanceOf;
+    mapping(address => mapping(address => uint256)) private _allowances;
```
mds1, line 39: Nit, let's keep the names the same and just add the underscore 
```suggestion
    mapping(address => uint256) private _balanceOf;

    mapping(address => mapping(address => uint256)) private _allowance;
```

mds1 (commented):
Oops you're right, I was on the wrong branch. Ok, agreed that is a much better UX!

Two small changes then we can merge

src/mocks/MockERC20.sol
arr00: Agree

src/mocks/MockERC20.sol
arr00: Sg

arr00:
@mds1 let me know if any other changes are needed

mds1 (approved):
Thank you!

PR #514, opened 2024-02-16
Use native assertions

klkvr:
Same as #503 with only difference being removal of assertion tests which were duplicating tests in https://github.com/foundry-rs/foundry/blob/master/testdata/cheats/Assert.t.sol


mds1:
Let me know if I'm wrong, but I think we still want to hold off on merging to avoid breaking `forge init`, until we #505 is ready and we bundle both into a release?

klkvr:
yep, we shouldn't just merge it to master without a release, just created a PR in advance

mds1 (approved):
Thank you!

PR #515, opened 2024-02-24
test: remove test marked as internal

DaniPopes:
I don't think there's a way to make this pass either, so I'm removing it.

### DIFF
--- a/test/StdChains.t.sol
+++ b/test/StdChains.t.sol
@@ -43,30 +43,38 @@ contract StdChainsTest is Test {
         assertEq(getChain("sepolia").rpcUrl, "https://sepolia.infura.io/v3/b9794ad1ddf84dfb8c34d6bb5dca2001");
     }
 
+    // Named with a leading underscore to clarify this is not intended to be run as a normal test,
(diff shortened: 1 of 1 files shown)


test/StdChains.t.sol
```diff
@@ -43,11 +43,6 @@ contract StdChainsTest is Test {
         assertEq(getChain("sepolia").rpcUrl, "https://sepolia.infura.io/v3/b9794ad1ddf84dfb8c34d6bb5dca2001");
     }
 
-    function testFuzz_Rpc(string memory rpcAlias) internal {
-        string memory rpcUrl = getChain(rpcAlias).rpcUrl;
-        vm.createSelectFork(rpcUrl);
-    }
```
mds1, line 49: Ah this got accidentally renamed in https://github.com/foundry-rs/forge-std/pull/433. It should be kept and renamed back to `testRpc`, where it's used in the commented out `testRpcs` test below. That test is commented out since testing public RPCs is slow and flaky

mds1 (approved):

Issue #522, opened 2024-03-14
`DSTest` removal in v1.8.0 seems to break the `forge` contract sizer filter


CruzMolina:
Bumping from v1.7.6 to v1.8.0 seems to break the way `forge build --sizes` filters out test contracts.

Test contracts that import & inherit `forge-std/Test.sol` now appear in the `forge build --sizes` output.

mds1:
Thanks for reporting! Fixed in https://github.com/foundry-rs/forge-std/releases/tag/v1.8.1

PR #523, opened 2024-03-15
fix: add `IS_TEST = true` to Test

mds1:
Closes https://github.com/foundry-rs/forge-std/issues/522

This used to come from [ds-test](https://github.com/dapphub/ds-test/blob/e282159d5170298eb2455a6c05280ab5a73a4ef0/src/test.sol#L38C5-L38C32), and v1.8.0 of forge-std removed that dependency

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -3,7 +3,7 @@ fs_permissions = [{ access = "read-write", path = "./"}]
 
 [rpc_endpoints]
 # The RPC URLs are modified versions of the default for testing initialization.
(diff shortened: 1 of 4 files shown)


mattsse (approved):
this works, we use this as an additional check to identify test contracts on the abi level, 

https://github.com/foundry-rs/foundry/blob/42da94276892f63afefd0dc743e862b058a4b4c2/crates/common/src/compile.rs#L255-L257

we do the same for Script.sol

klkvr (approved):

PR #524, opened 2024-03-18
fix(StdInvariant): correct `targetArtifactSelector` type

DaniPopes:
The return type of `targetArtifactSelector` is `(string, bytes4[])[]`.
See: <https://github.com/foundry-rs/foundry/blob/db8ea58ba607075ce0f39bb975a38da6b4889ea1/crates/evm/evm/src/executors/invariant/mod.rs#L377-L379>

### DIFF
--- a/src/StdInvariant.sol
+++ b/src/StdInvariant.sol
@@ -9,6 +9,11 @@ abstract contract StdInvariant {
         bytes4[] selectors;
     }
 
+    struct FuzzArtifactSelector {
+        string artifact;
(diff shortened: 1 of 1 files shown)


mds1 (approved):
Nice catch. Was `targetArtifactSelector`  basically a no-op and not working at all before? This is a breaking change to the interface, but it's ok if it wasn't doing anything.

Ideally we can point users to somewhere explaining how to fix the breaking change as well, even if it's just fleshing out the PR description here more

PR #526, opened 2024-03-19
ci(github): update CI workflow

sambacha:
update to correct foundry github action repo  
add identifier to compat checks

Also, shouldn't the Via-IR compilation time checks use the latest solc?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,10 +11,10 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


mds1 (approved):
Thank you!

> Also, shouldn't the Via-IR compilation time checks use the latest solc?

The intent of those is mainly to help prevent accidentally re-introducing regressions in build time that previously occurred when using that version. It's possible improvements to solc in future 0.8.x versions would no longer have resulted in those regressions, so I think it's more conservative to stick with the version known to be susceptible to the performance hits we saw

PR #532, opened 2024-03-25
Revert "ci(github): update CI workflow"

mds1:
Reverts foundry-rs/forge-std#526

> The workflow is not valid. .github/workflows/ci.yml (Line: 38, Col: 13): The identifier '0.8.0' is invalid. IDs may only contain alphanumeric characters, '_', and '-'. IDs must start with a letter or '_' and and must be less than 100 characters. .github/workflows/ci.yml (Line: 49, Col: 13): The identifier '0.7.6' is invalid. IDs may only contain alphanumeric characters, '_', and '-'. IDs must start with a letter or '_' and and must be less than 100 characters.

@sambacha any chance you can reopen with fixed identifiers? not sure why this didn't fail on the PR itself

PR #533, opened 2024-03-26
ci(workflow): upgrade actions and compiler checks

sambacha:
This addresses https://github.com/foundry-rs/forge-std/pull/532

Action lint check: 

https://rhysd.github.io/actionlint/#eJztl1tv0zAUx9/zKY66PWwP6QWhdgpimjYJ2ANiglY8Rm7iNKaunfmSUjG+O8dOlmVZQaC1EpXWl8b2uf1/x+pJBVnRCK6ugyCQIgoA1lItMy7Xccp0QUySR/DjJ+4XlvNY0VtLtYn8WufuG2CuiEhyqqsVQAgrog1VQZDSjFhu/ImyojLQOeU8gjnReRB8k3N/OreMp9U5GuoQawE7t8LYkBODKStXQ4tWGqsxKZDEMCn0AEtIltKai/J10JgIL+9aaEM4h3fSilRt6lOoA2TVbqj04P7RSMmTnDBxUY4a6zUzedSsAEqqNHOFCrbIDd90s94oJgxGVwt6b9t4Oxr1URjeH9anR3BJkuWaqFRDIlfYBDZnnJkNeIkNgCPMZHIKkqcICIhIQdC1e6wDgsyAkiQHbYtCKkNTWDEhVacaF6fe0V4k6IImLGMJMK2x3x1dV66KTmXerdUp92EZNoevyUafnD4WfteiiB0rrHl7fFLR8BcBmeglK8CFO23ZMtST5BJ6x5VXD+5goWgB4S30vhIlmFj03jgoouUFXafzc1y8v55+mF3Gn2bTm9m099j8OzMwam1lrNvb3zIY9s/6wwcEaQTlMD6Lhzumgs94eUHjNY18yoPDNOmPu5gm8XivmDDlAWJ6cpsme75NkwO8TeP+6FWX0zhu7+0BlEt6gKS2gNo3p/8M0xGUjIRMeUIMx5YbloataD1h+x2cHynRVtGn5jhgp05u89M/mrh57oP/GalDz1TsZsNosu21YBvTRAqj8H1H+41Bq5zB1cOzq6iP6Lszol3bPwq8JBjomSJjfOOju1fqStuh2i+JYsWzGxprH2a3aqvSdq51J72t9O6hxQ81/oVwU/83efkT0fh/tsJj0du8qk6UZem8stULu8f+1RRFoxUxBofSNmek5u6yswx+AQrwXZo=

## Questions:

Shouldn't foundry test be using a defined seed value?   
What about checking different EVM versions across SOLC versions?  
Why not perform gas checks and persist gas snapshots as an artifact, so any regressions can be identified?   

## Simplify

Doing something like this would make adding EVM versions to the testing easier

```yaml
jobs:
  build:
    runs-on: ubuntu-latest
    strategy:
      matrix:
     # These values match the names of test groups that will be run
        test-group:
          [
            0.8.0,
            0.7.6,
            0.7.0,
            0.6.12,
            0.6.2,
          ]
      fail-fast: false
    steps:
	# (...)
	# (...)
      - name: Check compatibility with ${{ matrix.test-group }}
        id: compat
        if: always()
        run: |
          output=$(forge build --skip test --use solc:/${{ matrix.test-group }}/)
          if echo "${output}" | grep -q "Warning"; then
              echo "${output}"  >> "$GITHUB_OUTPUT"
              exit 1
          fi

      - name: Pass or fail
        run: |
          if [[ ${{ steps.compat.outputs.output }} == 0 ]]; then exit 0; else exit 1; fi
```

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -8,14 +8,16 @@ on:
 jobs:
   sync-release-branch:
     runs-on: ubuntu-latest
-    if: startsWith(github.event.release.tag_name, 'v1')
+    if: ${{ github.repository_owner == 'foundry-rs' && startsWith(github.event.release.tag_name, 'v1') }}
     steps:
       - name: Check out the repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0
           ref: v1
 
+      # the email is derived from the bots user id
+      # found here: https://api.github.com/users/github-actions%5Bbot%5D
       - name: Configure Git
         run: |
           git config user.name github-actions[bot]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,20 +1,25 @@
 name: CI
 
+
 on:
-  workflow_dispatch:
+  workflow_dispatch: {}
   pull_request:
   push:
     branches:
       - master
 
+defaults:
+  run:
+    shell: bash
+
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Install Foundry
-        uses: onbjerg/foundry-toolchain@v1
+        uses: foundry-rs/foundry-toolchain@v1
         with:
           version: nightly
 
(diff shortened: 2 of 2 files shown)


.github/workflows/sync.yml
```diff
@@ -8,14 +8,16 @@ on:
 jobs:
   sync-release-branch:
     runs-on: ubuntu-latest
-    if: startsWith(github.event.release.tag_name, 'v1')
+    if: ${{ github.repository_owner == 'foundry-rs' && startsWith(github.event.release.tag_name, 'v1') }}
```
mds1, line 11: Presumably this change is to prevent the job running on forks?

mds1:
Superceded by https://github.com/foundry-rs/forge-std/pull/550 which is a bit smaller diff

PR #537, opened 2024-04-08
Add new cheatcodes to Vm.sol

kamuik16:
Following up from https://github.com/foundry-rs/foundry/pull/7539.

Adding new cheatcodes to Vm.sol.


mds1:
Thanks! There's a few other cheats since this PR and I'll add them all in one PR

PR #538, opened 2024-04-09
feat: add new maxes

PaulRBerg:
Useful for fuzzing timestamps in Foundry tests and bounding them to max uint40 and the maximum value that a Unix timestamp can have (year 2038 problem).

e.g. in practice:

https://github.com/sablier-labs/v2-core/blob/76e29fee33befaeb54a82473c28cf182e6f8855d/test/integration/fuzz/lockup-linear/createWithDurations.t.sol#L57

https://github.com/sablier-labs/v2-core/blob/76e29fee33befaeb54a82473c28cf182e6f8855d/test/utils/Constants.sol#L6-L13


src/StdUtils.sol
```diff
@@ -18,10 +18,12 @@ abstract contract StdUtils {
     address private constant CONSOLE2_ADDRESS = 0x000000000000000000636F6e736F6c652e6c6f67;
     uint256 private constant INT256_MIN_ABS =
         57896044618658097711785492504343953926634992332820282019728792003956564819968;
+    uint40 internal constant MAX_UNIX_TIMESTAMP = 2_147_483_647; // 2^31 - 1
```
mds1, line 21: Hm why is this one relevant for solidity devs? I'd think UINT32_MAX would be more appropriate, since I don't think may devs are storing timestamps as signed ints (this value equals `type(int32).max`)

mds1:
Sorry for the slow response here. Thinking about this more, I'd prefer to close this PR. The existing constants are there because they're needed by StdUtils functions, and these are not used anywhere in the StdUtils methods.

`Base.sol` is where constants are defined for user consumption, and even then I'd prefer to keep that minimal, and only add values that are not exposed in modern solidity (e.g. as you know `type([u]intN).max` is supported, making the constants not needed for most users anymore).

Going to close optimistically as part of the triage I'm doing now, but of course still open to further discussion even if the PR is closed :)

PR #540, opened 2024-04-12
feat: Add support for polygon amoy testnet

kcyeu:
As Polygon is deprecating Mumbai testnet and preferring the new testnet, Amoy. We are adding supports for Amoy accordingly in the implementation and test file.

Reference:
1. [Polygon Blog]((https://polygon.technology/blog/introducing-the-amoy-testnet-for-polygon-pos))

### DIFF
--- a/src/StdChains.sol
+++ b/src/StdChains.sol
@@ -214,9 +214,6 @@ abstract contract StdChains {
         setChainWithDefaultRpcUrl("arbitrum_nova", ChainData("Arbitrum Nova", 42170, "https://nova.arbitrum.io/rpc"));
         setChainWithDefaultRpcUrl("polygon", ChainData("Polygon", 137, "https://polygon-rpc.com"));
         setChainWithDefaultRpcUrl(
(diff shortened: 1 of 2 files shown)


src/StdChains.sol
```diff
@@ -216,6 +216,9 @@ abstract contract StdChains {
         setChainWithDefaultRpcUrl(
             "polygon_mumbai", ChainData("Polygon Mumbai", 80001, "https://rpc-mumbai.maticvigil.com")
         );
```
mds1, line 218: Should we remove mumbai, or is it not fully shut down yet and we should keep it for now?

mds1: Per https://polygon.technology/blog/introducing-the-amoy-testnet-for-polygon-pos I've removed Mumbai:

> No Goerli = broken Mumbai. The community consensus across the Polygon ecosystem has been to deprecate Mumbai in tandem with Goerli.

mds1 (approved):

PR #544, opened 2024-04-20
chore: fix bound log

totomanov:
Make the logs for `bound(uint256 x, uint256 min, uint256 max)` and `bound(int256 x, int256 min, int256 max)` consistent.


mattsse (approved):

baiwfg2:
@totomanov May I ask why to print this ? I want to disable this print no matter how verbose I'm using in `forge test`. How should I acheive that ?

PR #560, opened 2024-05-16
chore: remove duplicate interface

stevennevins:
Removes a duplicate interface.

I was running into:
Screenshot 2024-05-16 at 3 23 39 PM

with a newly created foundry repo that only had `Counters.sol` in the src directory. It seems to have been caused by the duplicate IERC721TokenReceiver interfaces

### DIFF
--- a/src/mocks/MockERC721.sol
+++ b/src/mocks/MockERC721.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.6.2 <0.9.0;
 
-import {IERC721Metadata} from "../interfaces/IERC721.sol";
+import {IERC721Metadata, IERC721TokenReceiver} from "../interfaces/IERC721.sol";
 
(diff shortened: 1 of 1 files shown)


stevennevins:
@mds1 tagging for visibility

mds1:
How do I reproduce the error? The below sequence only has `Counter.sol` and runs successfully, without the provided compiler error.
```
mkdir prj
cd prj
forge init
forge build
```

stevennevins:
https://github.com/stevennevins/ci-workflows/actions/runs/9114751802/job/25059446816

```
npm install

forge build

npx wagmi generate
```

It's more that this causes an issue with the minimal foundry repo when creating a SDK with wagmi.  This can be handled on the wagmi config side by excluding the MockERC721 in the options, but removing the duplicate interface seemed like a good option.

mds1 (approved):
Makes sense, thank you. I agree we don't need to duplicate the definition of the `IERC721TokenReceiver` interface in forge-std

PR #573, opened 2024-06-26
feat: deprecate console2

DaniPopes:
Since https://github.com/foundry-rs/foundry/pull/6504, Foundry treats the two types of `console` as the same in both traces and emitted logs. There is no reason for `console2` to exist anymore, so it should be removed.

To achieve this I also had to make the `console` functions `pure` by just copying over the previous `console2` logic.

To avoid a breaking change, we can just re-export `console as console2` in `console2.sol`.


src/console2.sol
```diff
@@ -1,1558 +1,4 @@
...
-        _sendLogPayload(abi.encodeWithSignature("log(address,address,address,bool)", p0, p1, p2, p3));
-    }
-
-    function log(address p0, address p1, address p2, address p3) internal pure {
-        _sendLogPayload(abi.encodeWithSignature("log(address,address,address,address)", p0, p1, p2, p3));
-    }
-
-}
\ No newline at end of file
+import {console as console2} from "./console.sol";
```
mattsse, line 4: this should be sufficient I think

mds1: Tested locally and this did not break my console2 usage

mds1 (approved):

hellwolf:
what about console2.log for int256?

```
    function log(string memory p0, int256 p1) internal pure {
        _sendLogPayload(abi.encodeWithSignature("log(string,int256)", p0, p1));
    }
```

?

mds1:
Thanks for flagging @hellwolf, just put out v1.9.1 which resolves this: https://github.com/foundry-rs/forge-std/releases/tag/v1.9.1

hellwolf:
awesome, thanks folks!

PR #577, opened 2024-07-02
fix: add missing console logs

DaniPopes:
Ref https://github.com/foundry-rs/forge-std/pull/573#issuecomment-2203315762

I copy pasted the console2 functions to console and replaced `int256` with `int`.
This is the entire diff.

### DIFF
--- a/src/console.sol
+++ b/src/console.sol
@@ -186,6 +186,10 @@ library console {
         _sendLogPayload(abi.encodeWithSignature("log(uint)", p0));
     }
 
+    function log(int p0) internal pure {
+        _sendLogPayload(abi.encodeWithSignature("log(int)", p0));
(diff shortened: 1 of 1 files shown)


mds1 (approved):

hellwolf:
well, the thing is, it now compiles, but I am not sure if it actually prints anything. At least not with "forge 0.2.0 (5ac78a9 2024-06-02T00:19:17.377492500Z)"

DaniPopes:
You're right, it's not hooked up, please open an issue in https://github.com/foundry-rs/foundry

hellwolf:
That's not idea, it could be a footgun, should it be rolled back, for now?

I will report a bug or propose a fix.

PR #584, opened 2024-07-22
feat: add boundLog to enable log-uniform sampling

guil-lambert:
Using `bound` to generate random inputs is heavily skewed towards the upper end of the range [see below]. To explore a more balanced distribution of values, users should have the option to sample in log space instead. 

This PR adds `boundLog(uint256 x, uint8 min, uint8 max)`, which generates a random number between [`2**min`, `2**(max+1)-1`] that is uniformly distributed in log space (ie. data generated follows a [log-uniform distribution](https://en.wikipedia.org/wiki/Reciprocal_distribution))

Note: contrarily to `bound`, `boundLog` WILL NOT return `x` when the input `x` is within the provided [`2**min`, `2**(max+1)-1`] bounds. 

## 
`bound` vs `boundLog` sampling between `1` and `2^32`, `N=10**6`
image

### DIFF
--- a/src/StdUtils.sol
+++ b/src/StdUtils.sol
@@ -87,6 +87,27 @@ abstract contract StdUtils {
         console2_log_StdUtils("Bound result", vm.toString(result));
     }
 
+    function _boundLog(uint256 x, uint8 min, uint8 max) internal pure virtual returns (uint256 result) {
+        require(min <= max, "StdUtils boundLog(uint256,uint8,uint8): Max is less than min.");
+
+        // If x is between min and max, DO NOT return x directly. This is to ensure that the sampling remains uniform in log space
+
+        // select an exponent between [min, max]
+        uint256 range = uint256(max) - uint256(min) + 1;
+        uint256 m0 = min + (x % range);
+
+        // randomize the input value, use it to generate a number between 2 ** m0 and 2 **(m0+1)-1
+        x = uint256(keccak256(abi.encode(x)));
+        uint256 m1 = x % 2 ** max;
+        result = 2 ** m0 + (m1 >> (max - m0));
+
+    }
+
+    function boundLog(uint256 x, uint8 min, uint8 max) internal pure virtual returns (uint256 result) {
+        result = _boundLog(x, min, max);
+        console2_log_StdUtils("Bound (Log) result", result);
+    }
+
     function boundPrivateKey(uint256 privateKey) internal pure virtual returns (uint256 result) {
         result = _bound(privateKey, 1, SECP256K1_ORDER - 1);
     }
--- a/test/StdUtils.t.sol
+++ b/test/StdUtils.t.sol
@@ -201,6 +201,37 @@ contract StdUtilsTest is Test {
     }
 
     /*//////////////////////////////////////////////////////////////////////////
+                                     BOUND LOG
+    //////////////////////////////////////////////////////////////////////////*/
+
+    function test_boundLog() public {
+
+        for (uint256 i = 0; i <= 255; ++i) {
+            assertEq(boundLog(0, uint8(0), uint8(i)), 2**0 + ((uint256(keccak256(abi.encode(uint256(0)))) % 2**i) >> i));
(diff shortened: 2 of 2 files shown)


mds1:
cc @grandizzy @klkvr for thoughts

grandizzy:
@mds1 I think it makes sense to have such directly in foundry and have strategies generate values in bounded range. What I can see here as a potential issue is that we will need config to specify limits per fuzzed param so could lead to more complex config, e.g. for a fuzzed test like `testFuzz(uint256 a, uint256 b)` we need to specify a min/max for `a` and  another one (or none) for `b`
Since we already have such per named param config for fixtures we could extend it to include bounds too (ref https://book.getfoundry.sh/forge/fuzz-testing#fuzz-test-fixtures)

mds1:
Yea the UX for this was always the trickiest, in the past we also considered just using the inline config which could be a good approach:
```solidity
/// forge-config: default.fuzz.range.x = [5, type(uint256).max];
/// forge-config: default.fuzz.distribution.x = logarithmic;
/// forge-config: default.fuzz.range.y = [1, 10];
/// forge-config: default.fuzz.distribution.x = uniform;
function testFoo(uint256 x, uint256 y) public {
  // --- snip ---
}
```

How would you specify distribution type with fixtures? Though, I am hesitant to expose distribution 
type at all because it adds more complexity to fuzz config. I would prefer to add some logic to forge to detect when this is needed and just accordingly. For example if fuzzing over a really large space like uint256, logarithmic uniformity is preferable and should be the default. If fuzzing a uint8 standard uniform sampling is preferable.

grandizzy:
> How would you specify distribution type with fixtures? 

Yeah, wasn't thinking to use fixtures for distribution but maybe having a similar way to current fixtures for all configuration of named parameter in Solidity

```Solidity
struct Uint256ParamConfig {
    uint256 min;
    uint256 max;
    uint256[] fixtures;
    uint256[] excluded;
}

Uint256ParamConfig public configAmount;
```
(basically including fixtures in same config and also superseeding assume and bound cheatcodes per named param). Though probably inline config is better UX here considering that we need specific struct per type...

> I would prefer to add some logic to forge to detect when this is needed and just accordingly. For example if fuzzing over a really large space like uint256, logarithmic uniformity is preferable and should be the default. If fuzzing a uint8 standard uniform sampling is preferable.

Yep, agree, maybe we should start an issue in foundry to track / implement these?

mds1:
> Yep, agree, maybe we should start an issue in foundry to track / implement these?

That sounds good to me, would you mind creating that? :)

07Vaishnavi-Singh:
> > How would you specify distribution type with fixtures?
> 
> Yeah, wasn't thinking to use fixtures for distribution but maybe having a similar way to current fixtures for all configuration of named parameter in Solidity
> 
> ```solidity
> struct Uint256ParamConfig {
>     uint256 min;
>     uint256 max;
>     uint256[] fixtures;
>     uint256[] excluded;
> }
> 
> Uint256ParamConfig public configAmount;
> ```
> 
> (basically including fixtures in same config and also superseeding assume and bound cheatcodes per named param). Though probably inline config is better UX here considering that we need specific struct per type...
> 
> > I would prefer to add some logic to forge to detect when this is needed and just accordingly. For example if fuzzing over a really large space like uint256, logarithmic uniformity is preferable and should be the default. If fuzzing a uint8 standard uniform sampling is preferable.
> 
> Yep, agree, maybe we should start an issue in foundry to track / implement these?

Hey I am working on this issue and trying to solve it using the struct 
```
struct ParamConfig {
        uint256 min;
        uint256 max;
        DistributionType distributionType;
        uint256[] fixtures;
        uint256[] excluded;
    } 

```
and I am fetching the from another function that is checking the unit type of the value. Is it important to keep the exact same struct to solve the issue?

Theghost256 (approved):

mds1:
Closing in favor of https://github.com/foundry-rs/foundry/pull/9154

Hugoo:
Related to this issue: https://github.com/foundry-rs/foundry/issues/8788

PR #588, opened 2024-07-26
chore: fix comment

riyueguang:


mds1:
Thanks. This file is autogenerated from https://github.com/foundry-rs/foundry/blob/df5f45c8c1a7e138203f1f479940f1500752b3dc/crates/cheatcodes/spec/src/vm.rs#L1517 — you will need to make the PR there, and then it will propagate here next we generate `Vm.sol`

PR #591, opened 2024-07-30
🐛 Fix `logMemory` in `safeconsole`

Philogy:
The `safeconsole.logMemory` method did not work as expected because it was calling the `logBytes(bytes)` method on the console pre-compile and not `log(bytes)`.

I furthermore marked all assembly blocks as "memory safe" as it doesn't change memory and to prevent it from conflicting with certain viaIR optimizations.

### DIFF
--- a/src/safeconsole.sol
+++ b/src/safeconsole.sol
@@ -11,6 +11,7 @@ library safeconsole {
     function _sendLogPayload(uint256 offset, uint256 size) private pure {
         function(uint256, uint256) internal view fnIn = _sendLogPayloadView;
         function(uint256, uint256) internal pure pureSendLogPayload;
+        /// @solidity memory-safe-assembly
         assembly {
             pureSendLogPayload := fnIn
         }
@@ -18,6 +19,7 @@ library safeconsole {
     }
 
     function _sendLogPayloadView(uint256 offset, uint256 size) private view {
+        /// @solidity memory-safe-assembly
         assembly {
             pop(staticcall(gas(), CONSOLE_ADDR, offset, size, 0x0, 0x0))
         }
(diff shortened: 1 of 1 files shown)


mds1:
Sorry where is `log(bytes)` defined? I do not see that method in the console libraries

Philogy:
> Sorry where is log(bytes) defined? I do not see that method in the console libraries

@mds1 https://github.com/foundry-rs/forge-std/blob/2cbff0602d340503dba9828ab6981053704d1384/src/console.sol#L53-L55

mds1:
That signature you linked is `logBytes(bytes)` which is `0xe17bf956`, but the one in the PR is `log(bytes)` which is `0x0be77f56`

Philogy:
> That signature you linked is `logBytes(bytes)` which is `0xe17bf956` (what was originally in `console.sol`), but the PR changes it `log(bytes)` (`0x0be77f56`) and I don't see a `function log(bytes)` defined anywhere

So the library function is called `logBytes(bytes)` (which is also why I initially put that) but it dispatches out (the actual call to the underlying "console pre-compile" address) `log(bytes)`.

You can try it yourself locally by testing the code from my PR:

```solidity
pragma solidity ^0.8.0;

import {Test} from "../src/Test.sol";
import {safeconsole} from "../src/safeconsole.sol";

/// @author philogy <https://github.com/philogy>
contract safeconsoleTest is Test {
    struct A {
        uint256 a;
        uint256 b;
    }

    function test_logMemory() public pure {
        A memory a = A(uint256(keccak256("a")), uint256(keccak256("b")));
        safeconsole.logMemory(0x00, 0x100);
    }
}
```

Here also a link to foundry's actual definition of the underlying console interface: https://github.com/foundry-rs/foundry/blob/26a7559758c192911dd39ce7d621a18ef0d419e6/crates/evm/abi/src/console/mod.rs#L18-L19

mds1:
I see now, thank you!

mds1 (approved):

Philogy:
No worries, happens, thanks for being so responsive!

PR #594, opened 2024-08-06
cheats: fix assume not precompile

tynes:
After the cancun hardfork, the point evaluation precompile was added at `address(10)` which the `assumeNotPrecompile` implementation did not take into account. Pectra is scheduled to include BLS precompiles and likely more precompiles will be introduced in other future hardforks. We might as well reserve the least significant byte for Ethereum precompiles, as the address space is large and there is no need to make calls from these addresses, and in return code will not break on new EVM specs.

This was causing test failures in the Optimism test suite when we updated the EVM version to cancun.

### DIFF
--- a/src/StdCheats.sol
+++ b/src/StdCheats.sol
@@ -322,8 +322,8 @@ abstract contract StdCheatsSafe {
         // Note: For some chains like Optimism these are technically predeploys (i.e. bytecode placed at a specific
         // address), but the same rationale for excluding them applies so we include those too.
 
-        // These should be present on all EVM-compatible chains.
-        vm.assume(addr < address(0x1) || addr > address(0x9));
+        // These are reserved by Ethereum and may be on all EVM-compatible chains.
+        vm.assume(addr < address(0x1) || addr > address(0x100));
 
         // forgefmt: disable-start
         if (chainId == 10 || chainId == 420) {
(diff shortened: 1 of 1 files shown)


src/StdCheats.sol
```diff
@@ -322,8 +322,8 @@ abstract contract StdCheatsSafe {
         // Note: For some chains like Optimism these are technically predeploys (i.e. bytecode placed at a specific
         // address), but the same rationale for excluding them applies so we include those too.
 
-        // These should be present on all EVM-compatible chains.
-        vm.assume(addr < address(0x1) || addr > address(0x9));
+        // These are reserved by Ethereum and may be on all EVM-compatible chains.
+        vm.assume(addr < address(0x1) || addr > address(0x100));
```
mds1, line 326: I agree with the larger upper bound so this is more future-proof. Do you have a source for the `0x100` value? I've only seen https://eips.ethereum.org/EIPS/eip-1352 which reserves through `0xffff`.

tynes: The genesis for testnets includes 1 wei for the precompiles in the range `address(0)` to `address(0xff)`. See https://github.com/eth-clients/holesky/blob/main/metadata/genesis.json

I meant to follow this standard, will update the PR to use `address(0xff)` rather than `address(0x100)`

tynes: Updated in [dab1435](https://github.com/foundry-rs/forge-std/pull/594/commits/dab1435989f46223db6f63e9ac9821c157bd74ed)

mds1 (approved):
Thank you!

PR #602, opened 2024-09-11
feat: setArbitraryStorage, copyStorage, mockFunction and assumeNoRevert

grandizzy:
re https://github.com/foundry-rs/foundry/pull/8807

### DIFF
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -1437,6 +1437,9 @@ interface VmSafe {
     /// If the condition is false, discard this run's fuzz inputs and generate new ones.
     function assume(bool condition) external pure;
 
+    /// Discard this run's fuzz inputs and generate new ones if next call reverted.
+    function assumeNoRevert() external pure;
+
     /// Writes a breakpoint to jump to in the debugger.
     function breakpoint(string calldata char) external;
 
@@ -1546,6 +1549,9 @@ interface VmSafe {
     /// Compute the address a contract will be deployed at for a given deployer address and nonce.
     function computeCreateAddress(address deployer, uint256 nonce) external pure returns (address);
 
+    /// Utility cheatcode to copy storage of `from` contract to another `to` contract.
+    function copyStorage(address from, address to) external;
+
     /// Returns ENS namehash for provided string.
     function ensNamehash(string calldata name) external pure returns (bytes32);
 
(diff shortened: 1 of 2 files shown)


mds1:
Thanks! Can you link to the upstream PR(s) in the description for traceability?

src/Vm.sol
```diff
@@ -1571,6 +1577,9 @@ interface VmSafe {
     /// Unpauses collection of call traces.
     function resumeTracing() external view;
 
+    /// Utility cheatcode to set arbitrary storage for given target address.
+    function setArbitraryStorage(address target) external;
```
mds1, line 1581: Do we have more docs for this in the foundry book? From this docstring alone it's not clear when/why to use this function or what it does exactly

src/Vm.sol
```diff
@@ -1699,6 +1708,14 @@ interface Vm is VmSafe {
     /// Calldata match takes precedence over `msg.value` in case of ambiguity.
     function mockCall(address callee, uint256 msgValue, bytes calldata data, bytes calldata returnData) external;
 
+    /// Whenever a call is made to `callee` with calldata `data`, this cheatcode instead calls
+    /// `target` with the same calldata. This functionality is similar to a delegate call made to
+    /// `target` contract from `callee`.
+    /// Can be used to substitute a call to a function with another implementation that captures
+    /// the primary logic of the original function but is easier to reason about.
+    /// If calldata is not a strict match then partial match by selector is attempted.
+    function mockFunction(address callee, address target, bytes calldata data) external;
```
mds1, line 1717: Curious what the use case is for this? 

> If calldata is not a strict match then partial match by selector is attempted.

What does this mean, why is there calldata matching here? I thought it would just be `address(target).call(data)`

grandizzy:
> Thanks! Can you link to the upstream PR(s) in the description for traceability?

yep, added

src/Vm.sol
grandizzy: yes, I an creating a book PR with details and examples fir each of them

src/Vm.sol
grandizzy: one could add different mocks for a function called with different params (e.g. mock `Counter.count(1)` with `Counter1.count()` while `Counter.count(2)` mocked with `Counter2.count()`) or a catch all mock for `Counter.count`

mds1 (approved):

Issue #603, opened 2024-09-17
Is using mock contracts for testing necessary?


milosdjurica:
In test cases when revert is expected, tests are following this pattern: 
1. Deploy `MockContract`
2. Call method on that `MockContract` that is supposed to revert.

As far as I saw, mocks are used only for revert cases.

Quick example : 

https://github.com/foundry-rs/forge-std/blob/beb836e33f9a207f4927abb7cd09ad0afe4b3f9f/test/StdChains.t.sol#L86-L94

### Why do we need to use mocks to properly test the revert? 

Can't we just test revert like this? -> 

```js

    function test_RevertIf_ChainNotFound() public {
        vm.expectRevert("StdChains getChain(string): Chain with alias \"does_not_exist\" not found.");
        getChain("does_not_exist");
    }

```

I have tried changing tests to 2nd approach and tests are passing. I am not sure if I am missing something. 
I will open the PR with changes on the `StdChains.t.sol` so you can get a closer look on what do I mean. Please let me know your thoughts on this :)

mds1:
We intentionally use mocks because `vm.expectRevert` is intended to be used on the next call (or contract creation). If you remove the mocks and directly call the internal functions, you JUMP to that function. This was accidentally supported by foundry, but not a lot of people rely on it so it's not easy to remove. It's a big footgun because if you have two consecutive `expectRevert` calls with jumps, the test stops executing. So:

```solidity
// This one is ok
vm.expectRevert("StdChains getChain(string): Chain with alias \"does_not_exist\" not found.");
getChain("does_not_exist");

// All code below here never executes! This is dangerous
vm.expectRevert("StdChains getChain(string): Chain with alias \"does_not_exist\" not found.");
getChain("does_not_exist");
```

I checked the forge book quickly and didn't see this documented, cc @klkvr @grandizzy @Evalir for more info

PR #604, opened 2024-09-17
Removing mock contracts in tests

milosdjurica:
Corresponds to the issue #603 .

If those changes indeed should be implemented, I am open to change other test files too.


mds1:
Thanks! But closing per https://github.com/foundry-rs/forge-std/issues/603#issuecomment-2358649553 :)

PR #605, opened 2024-09-18
feat: add keyExists to stdJson and stdToml

mds1:
Hey @zerosnacks @klkvr can I get a quick review here?

One thing I'm wondering: why are the `keyExists` cheats `view` but the other JSON/TOML parsing cheats are `pure`?


mattsse (approved):
these are fine, right @klkvr ?

zerosnacks (approved):
Proposed API LGTM

I don't think there is a clear reason for marking these as `view`

klkvr (approved):
lgtm

yeah, we can probably make it `pure`

PR #615, opened 2024-10-04
Update Vm.sol with vm.mockCalls methods

Tudmotu:
Added methods for `vm.mockCalls` in `Vm` interface, see https://github.com/foundry-rs/foundry/pull/9024.

### DIFF
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -1592,16 +1592,22 @@ interface VmSafe {
     /// Returns a random `address`.
     function randomAddress() external returns (address);
 
-    /// Returns an random `bool`.
+    /// Returns a random `bool`.
     function randomBool() external view returns (bool);
 
-    /// Returns an random byte array value of the given length.
+    /// Returns a random byte array value of the given length.
     function randomBytes(uint256 len) external view returns (bytes memory);
 
-    /// Returns an random `int256` value.
+    /// Returns a random fixed-size byte array of length 4.
+    function randomBytes4() external view returns (bytes4);
+
+    /// Returns a random fixed-size byte array of length 8.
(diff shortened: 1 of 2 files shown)


src/Vm.sol
```diff
@@ -1747,6 +1747,16 @@ interface Vm is VmSafe {
...
 
+    /// Mocks multiple calls to an address, returning specified data for each call.
+    /// Calldata can either be strict or a partial match, e.g. if you only
+    /// pass a Solidity selector to the expected calldata, then the entire Solidity
+    /// function will be mocked.
+    function mockCalls(address callee, bytes calldata data, bytes[] calldata returnData) external;
+
+    /// Mocks multiple calls to an address with a specific `msg.value`, returning specified data for each call.
+    /// Calldata match takes precedence over `msg.value` in case of ambiguity.
+    function mockCalls(address callee, uint256 msgValue, bytes calldata data, bytes[] calldata returnData) external;
```
zerosnacks, line 1758: Small note, you can Vm.sol is generated by running `scripts/vm.py` directly from source. If you want to add additional comments please add them in the original source as otherwise it won't persist

Tudmotu: Ok, got it. I wasn't sure if I'm supposed to run it or if it runs automatically. Running it adds a couple of changes that are unrelated to the `mockCalls` feature. Hope it's ok to include them.

zerosnacks: That's OK, not a problem

Tudmotu:
I made a it of a mess with the commits, so I squashed them now.

zerosnacks:
Thanks! LGTM

PR #616, opened 2024-10-04
feat: add modifier to run `afterTest` in tests that inherit StdAfterTest

yash-atreya:
Ref: https://github.com/foundry-rs/forge-std/issues/617 

Implementing this directly in foundry would add unnecessary complexity to `ContractRunner`. See https://github.com/foundry-rs/foundry/pull/8985#issuecomment-2393165001 

This approach is cleaner and easier. 

Implemented the modifier and `afterTest()` fn that needs to be overridden in a separate abstract contract `StdAfterTest`; and not in existing `StdCheatsSafe` as this would break existing tests. Hence, making this feature opt-in.

### DIFF
--- a/src/StdAfterTest.sol
+++ b/src/StdAfterTest.sol
@@ -1,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.6.2 <0.9.0;
+
+// Inheritted by test contracts where afterTest() is needed
+abstract contract StdAfterTest {
+    // This modifier needs to applied to the test after which we want to run the afterTest() function
+    modifier runAftetTest() {
+        _;
+        afterTest();
+    }
+
+    // This is needs to be overriden by the Test contract that inherits this
+    function afterTest() public virtual;
+}


grandizzy:
makes sense to me, would be great to have @mds1 input too

src/StdAfterTest.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.6.2 <0.9.0;
+
+// Inheritted by test contracts where afterTest() is needed
+abstract contract StdAfterTest {
```
mds1, line 5: We will need to inherit from this here: https://github.com/foundry-rs/forge-std/blob/4e5f46a8014eb0a40526e88af6c89001096e4656/src/Test.sol#L30

src/StdAfterTest.sol
```diff
@@ -0,0 +1,14 @@
...
+    // This modifier needs to applied to the test after which we want to run the afterTest() function
+    modifier runAftetTest() {
+        _;
+        afterTest();
+    }
+
+    // This is needs to be overriden by the Test contract that inherits this
+    function afterTest() public virtual;
```
mds1, line 13: Can we add some comments to document how this works exactly and what it's intended for? Does it run only after successful tests, or also failed ones? If both, I can see a use case like cleaning up tmp files. What's the use case for only running after successful test cases?

yash-atreya: `afterTest()` won't run after tests that revert.

yash-atreya: Makes sense.
Closing this and adding an example impl to the linked issue.

PR #619, opened 2024-10-04
Add `--from` option to vm.py, to generate Vm.sol from json file

Tudmotu:
~~Defaults to "foundry-rs" and "master" respectively to maintain backwards-compatibility.~~

Added the `--from` option which lets you specify a local json file to generate `Vm.sol` from.

This PR uses `argparse` to implement the argument handling.

When running `vm.py -h` will display help information:
image

Running the script with no args should behave same as before.


mds1:
Can you describe your use case for this? Also cc @DaniPopes who wrote the script originally

DaniPopes:
It would be much easier to have just one argument that's the path of `cheatcodes.json`

Tudmotu:
@mds1 yes, my use case is that I like to open the Foundry Book and Forge-Std PRs before the main PR on Foundry gets merged. The reason I do it like this is because I have very little time to actually write code so when I have the opportunity, I try to get everything done all at once.

This means that in order to update the Vm interface using `vm.py`, I need to manually edit the `CHEATCODES_JSON_URL` to point to my feature branch on my Foundry fork.

Adding these params lets me run e.g. `./scripts/vm.py tudmotu vm.mockCalls` and it'll be updated based on my feature branch.

Tudmotu:
> It would be much easier to have just one argument that's the path of `cheatcodes.json`

 You mean provide the entire URL on the command line?

DaniPopes:
No, a path to a file, defaulting to fetching from the default source and keeping it in memory

Tudmotu:
> No, a path to a file, defaulting to fetching from the default source and keeping it in memory

Oh, so for example `vm.py --from /path/to/my/cheatcodes.json`?

mds1:
Thanks @DaniPopes

@Tudmotu I can review + merge once that change is made!

Tudmotu:
@mds1, @DaniPopes , updated the PR with the `--from` flag

DaniPopes (approved):

PR #639, opened 2024-12-09
feat: new cheatcode `assumeUnusedAddress`

xenide:
Sometimes during fuzzing we want to test with a wide range of addresses, but these address need to be empty and not already have contracts deployed. This cheatcode would come in handy for that. 

Any suggestions if other overloads are necessary for extended use cases?


DaniPopes (approved):
Thanks!

Skippyy96:
Thank you

PR #643, opened 2025-01-09
test: remove `testFail*` tests

zerosnacks:
In preparation to deprecate `testFail*`: https://github.com/foundry-rs/foundry/pull/9574

- Removes `deployMockERC721` and `deployMockERC20` from `StdUtils` as their tests includes large amount of `testFail*` tests and is generally outdated. I haven't been able to find the `deploy*` methods being used in any public project on Github.
- Implements workaround for other cases

Alternatively we could upstream Solady's mocks and tests (which have been updated to not use `testFail*` but this feels out of place and would be incomplete (ERC4626, ERC6909, etc.. https://github.com/Vectorized/solady/tree/main/test/utils/mocks). I think it is most common for people to rely on the mocks their library comes with (be it Solmate, Solady or OpenZeppelin) or write their own wrapper around the token implementation.


zerosnacks:
cc @mds1 if you would like to retain the deploy utility methods I could port over Solady's test suite which does not have testFail* methods

PR #644, opened 2025-01-11
feat: add ERC7540 and ERC7575 interfaces

hieronx:
This adds the interfaces for the 2 finalized extensions to ERC4626, [ERC7540](https://eips.ethereum.org/EIPS/eip-7540) and [ERC7575](https://eips.ethereum.org/EIPS/eip-7575). See https://ethereum.org/en/developers/docs/standards/tokens/erc-4626/#introduction for details on these extensions.

Adding these would be helpful for repositories that rely on the interfaces in `forge-std`.

ERC7540 interfaces are split by async deposit, async redeem, and fully async vaults, per the ERC specification. This also ensures the ERC-165 ID matches for each vault type.

The ERC7575 interface specifies all ERC4626 methods, rather than extending `IERC4626.sol`, because:
1. ERC20 methods need to be excluded for ERC7575.
2. This ensures `IERC7575.interfaceId` is the expected ID as specified in https://eips.ethereum.org/EIPS/eip-7575#erc-165-support, since `.interfaceID`  does not account for inherited interfaces.


hieronx:
Done @zerosnacks !

hieronx:
@zerosnacks CI is fixed now as well

DaniPopes (approved):

zerosnacks:
I don't see any blockers, merging this.

PR #646, opened 2025-01-21
OP Networks Add

ElliotFriedman:
Add: zora, race, metal, binary, orderly and mode testnets and mainnet configs.

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -1,5 +1,7 @@
 [profile.default]
 fs_permissions = [{ access = "read-write", path = "./"}]
+optimizer = true
+optimizer_runs = 200
 
 [rpc_endpoints]
 # The RPC URLs are modified versions of the default for testing initialization.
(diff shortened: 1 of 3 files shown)


mds1:
Hey @zerosnacks @grandizzy, looks like the foundryup changes broke CI here—could you PR in a fix?

grandizzy:
yeah, we probably have to add optimizer = true and optimizer_runs = 200 in foundry.toml to keep same setup

grandizzy:
> > yeah, we probably have to add optimizer = true and optimizer_runs = 200 in foundry.toml to keep same setup
> 
> Done, that seems to fix the problem, at least locally.
> 
> It's hard to understand what downstream effects this will have if any of users of the library.
> 
> Would this force compile of tests in downstream repos with these optimizer settings?

That's correct. This was the default in foundry until https://github.com/foundry-rs/foundry/pull/9657 which disables optimizer by default, so I don't think there'll be any issue.

zerosnacks (approved):

zerosnacks:
Thanks @ElliotFriedman!

PR #649, opened 2025-01-31
Fix DEFAULT_TEST_CONTRACT

CodeSandwich:
The test contracts have a different address than the one in the constant. I tried printing `address(this)` in 2 separate tests and they were all showing this address. It makes sense because the `msg.sender` and `tx.origin` were always `DEFAULT_SENDER` in the constructors, `setUp`s and the test functions, and the new `DEFAULT_TEST_CONTRACT` can be obtained by calculating the `CREATE` address for `DEFAULT_SENDER` and nonce 1.

### DIFF
--- a/src/Base.sol
+++ b/src/Base.sol
@@ -14,7 +14,7 @@ abstract contract CommonBase {
     // Default address for tx.origin and msg.sender, 0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38.
     address internal constant DEFAULT_SENDER = address(uint160(uint256(keccak256("foundry default caller"))));
     // Address of the test contract, deployed by the DEFAULT_SENDER.
-    address internal constant DEFAULT_TEST_CONTRACT = 0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f;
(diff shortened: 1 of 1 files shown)


zerosnacks (approved):
Hi @CodeSandwich thanks for flagging this

We are making the same assertion here in Foundry: https://github.com/foundry-rs/foundry/blob/f6133f9e5fbd37864958cc58457f795ce35cd469/testdata/default/core/ContractEnvironment.t.sol#L16

It appears that when the test contract creates its first contract it will be deployed at `0x5615deb798bb3e4dfa0139dfa1b3d433cc23b72f` as seen here

https://github.com/foundry-rs/foundry/blob/f6133f9e5fbd37864958cc58457f795ce35cd469/crates/forge/tests/fixtures/SimpleContractTestVerbose.json#L119-L120

compared to

https://github.com/foundry-rs/foundry/blob/f6133f9e5fbd37864958cc58457f795ce35cd469/crates/forge/tests/fixtures/SimpleContractTestVerbose.json#L79-L80

CodeSandwich:
You're right, this is all correct and the address doesn't need to be fixed; I simply didn't understand what it means. IMO the current docs are just wrong and need to be fixed, but I'll do it in a separate PR.

PR #662, opened 2025-02-27
chore: avoid using Alchemy API keys in tests due to limits

zerosnacks:
We are running into Alchemy API limits (https://github.com/foundry-rs/forge-std/actions/runs/13545994206/job/37926743447?pr=654) that is causing the CI to block, small refactor to avoid using hardcoded API keys.


mattsse (approved):
alchemy 👎

PR #689, opened 2025-07-14
feat(StdMath): Add INT256_MIN_ABS constant to avoid magic number dupl…

reallesee:
Introduces a new constant INT256_MIN_ABS in the stdMath library to store the absolute value of INT256_MIN. The abs() function now uses this constant instead of duplicating the magic number. This change improves code readability, maintainability, and follows the DRY principle. It also prevents potential overflow issues that could occur when trying to calculate -INT256_MIN directly.


DaniPopes:
Thanks but this is unnecessary because it's used only once

PR #698, opened 2025-07-31
ci: add Foundry stable, use matrixes

DaniPopes:


DaniPopes:
right, we should have both, I'll rewrite it

mattsse (approved):

PR #699, opened 2025-07-31
chore(StdAssertions): manually invert conditions

DaniPopes:
Optimize it ourselves.


src/StdAssertions.sol
```diff
@@ -402,13 +402,13 @@ abstract contract StdAssertions {
     }
 
     function assertLt(uint256 left, uint256 right) internal pure virtual {
-        if (!(left < right)) {
+        if (left >= right) {
```
0xrusowsky, line 405: in this case i think both take 2 opcodes, but personally i prefer this one for clarity

0xrusowsky (approved):

PR #730, opened 2025-09-16
chore(`ci`): pin deps in workflow and add `dependabot` to update them weekly

zerosnacks:
Pinning hashes for dependencies in workflows is a security best practice

Excluded from pinning are actions from the `github/*` and `actions/*` given that these are officially managed by Github and are not raised by `zizmor`

By configuring dependabot with `package-ecosystem: "github-actions"` it will open a pull request only for updating pinned hashes (not cargo, etc..): https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot#enabling-dependabot-version-updates-for-actions

The `<hash> #<branch_name>` syntax is what dependabot picks up on

Note: `foundry-toolchain@v1` has been left unpinned as it will help us catch issues more easily and it is in our interest to be up to date. Let me know if this makes sense @grandizzy or if we should pin instead.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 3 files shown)


grandizzy (approved):
👍

PR #830, opened 2026-03-03
Add NatSpec for StdUtils

esorense:
This PR adds missing NatSpec to `stdMath` and key `StdUtils` helpers (`bound`, `boundPrivateKey`, `bytesToUint`, `computeCreate*`, `hashInitCode`, and `getTokenBalances`).

  This is needed because these widely used utility APIs were under-documented, and `bound` semantics are easy to misread without explicit wording. The updated comments now clearly describe the real behavior (including wrap-into-range behavior and deprecation notes).

  Correctness is low-risk and verifiable: this is comment-only, with no logic, signature, or behavior changes. The annotations were written directly from the current implementation and existing revert conditions.


7zkm7b8gw9-web (approved):

zerosnacks (dismissed):
Thanks!

zerosnacks (approved):

PR #666, opened 2025-03-03
chore: unify interface NatSpec

zerosnacks:
The formatting for the interfaces was inconsistent and missing details, this should make it more unified.

Issue #653, opened 2025-02-04
Enhance documentation with NatSpec for increased readability and for autogeneration of related docs


zerosnacks:
To do:

- [ ] Interfaces
- [ ] Base.sol
- [ ] console.sol
- [ ] console2.sol
- [ ] safeconsole.sol
- [ ] Script.sol
- [ ] StdAssertions.sol
- [ ] StdChains.sol
- [ ] StdCheats.sol
- [ ] StdError.sol
- [ ] StdInvariant.sol
- [ ] StdJson.sol
- [ ] StdMath.sol
- [ ] StdStorage.sol
- [ ] StdStyle.sol
- [ ] StdToml.sol
- [ ] StdUtils.sol
- [ ] Test.sol
- [ ] Vm.sol

Goals:

- Document all public facing functions with `@notice`
- Mark private variables and functions with underscore prefix
- Unify layouts to follow common best practices of grouping storage variables
- Should not modify any user facing function or interface

This will make it easier to autogenerate documentation that we can then load into https://book.getfoundry.sh/reference/forge-std/ and keep it up to date

japui-coder:
Enhance documentation with NatSpec for increased readability and for autogeneration of related docs #653

PR #845, opened 2026-04-11
docs: add NatSpec to StdMath and StdError

sylvarissolutions-max:
Part of #653. Two more of the files listed in the NatSpec docs checklist.

## StdMath.sol

Added \`@notice\`, \`@param\`, and \`@return\` tags to the library and all five internal functions:

- \`abs(int256)\`
- \`delta(uint256, uint256)\`
- \`delta(int256, int256)\`
- \`percentDelta(uint256, uint256)\`
- \`percentDelta(int256, int256)\`

I did not rename \`INT256_MIN\` to \`_INT256_MIN\` even though the issue mentions the underscore-prefix convention for private variables. The already-merged StdAssertions.sol work (PR #666 / StdInvariant) left its private constants without the underscore prefix (for example \`FAILED_SLOT\`), so I matched that pattern to keep this PR purely additive. Happy to apply the rename in a follow-up or in this PR if you prefer the underscore convention going forward. Let me know which way you want to settle it.

## StdError.sol

Added \`@notice\` to each of the nine public panic-code constants describing what Solidity panic it corresponds to. Panic codes cross-checked against the Solidity docs.

## What is not in this PR

- No logic changes
- No interface changes
- No file reordering (internal up, private down was also listed in #653 - can do that in a follow-up if you want)
- No tests touched

## Testing

I do not have Foundry set up on this machine so I have not run \`forge fmt --check\` or \`forge test\` locally. The diff is 38 insertions and 0 deletions, all inside NatSpec comment blocks above existing declarations, so behavior is unchanged. CI should cover the formatting and test pass. If CI flags anything I will fix it in this branch.

Happy to split this into two PRs (one per file) if that is easier to review.

### DIFF
--- a/src/StdError.sol
+++ b/src/StdError.sol
@@ -2,14 +2,32 @@
 // Panics work for versions >=0.8.0, but we lowered the pragma to make this compatible with Test
 pragma solidity >=0.8.13 <0.9.0;
 
+/// @notice Pre-encoded Solidity panic error selectors for use in test assertions.
 library stdError {
+    /// @notice Panic caused by `assert(false)` or an assertion failure (0x01).
     bytes public constant assertionError = abi.encodeWithSignature("Panic(uint256)", 0x01);
+
+    /// @notice Panic caused by arithmetic overflow or underflow (0x11).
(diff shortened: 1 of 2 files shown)


zerosnacks (approved):
Thanks!

PR #854, opened 2026-04-29
ci: bump SOLC_LATEST to 0.8.35

zerosnacks:
Solidity 0.8.35 was just released. Bumps `SOLC_LATEST` to 0.8.35 and clears `SOLC_PRERELEASE` (the matrix loop is now guarded so prerelease entries are skipped when empty).


7zkm7b8gw9-web (commented):
Hey so my 9w addresses no longer show up whats going on

PR #860, opened 2026-05-13
docs: add NatSpec `@notice` to public functions

mablr:
Towards #653

Adds `@notice` tags to all public-facing and internal functions enabling forge doc to autogenerate reference documentation for the Foundry Book. No function signatures or behavior changed.


figtracer (approved):
lg

zerosnacks (approved):

PR #878, opened 2026-06-11
fix: quote dependabot schedule time

decofe:
## Summary
- quote Dependabot `schedule.time` so GitHub parses it as a string
- set the fixed weekly run time to Monday at 07:00 UTC

## Testing
- checked `.github/dependabot.yml` contains quoted `time: "07:00"` entries

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,7 @@ updates:
     schedule:
       interval: weekly
       day: monday
-      time: "07:00"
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):

7zkm7b8gw9-web (commented):
Is all my info here?

7zkm7b8gw9-web (commented):
Is all my info here?

PR #879, opened 2026-06-13
docs: clarify JSON key selector syntax

### LINKED ISSUE #394: docs(StdJson): document that key parameter must start with a dot
The current documentation for the JSON cheats does not explain that the `key` parameter must be provided with a leading dot:

https://github.com/foundry-rs/forge-std/blob/e8a047e3f40f13fa37af6fe14e6e06283d9a060e/src/Vm.sol#L213-L239

I was trying to access the key by its simple name, like so:

```solidity
string memory json = '{"bar":"Hello World"}';
string memory bar = vm.parseJsonString(json, "bar");
```

And that didn't work until I added the leading dot, i.e. `.bar`.

Side note: the parameters for the `parseJsonSOMETHING` cheats have been left unnamed.

SSDWGG:
## Summary
- document that JSON `key` parameters use selector syntax
- call out leading-dot object field selectors, such as `.bar`, and `$` for root

Fixes #394

### DIFF
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -1060,6 +1060,8 @@ interface VmSafe {
 
     // ======== JSON ========
 
+    /// JSON `key` parameters use selector syntax: use a leading dot for object fields
+    /// (for example, `.bar`) or `$` for the root object.
     /// Checks if `key` exists in a JSON object.
     function keyExistsJson(string calldata json, string calldata key) external view returns (bool);
 
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -1060,9 +1060,10 @@ interface VmSafe {
 
     // ======== JSON ========
 
+    /// Checks if `key` exists in a JSON object.
+    ///
     /// JSON `key` parameters use selector syntax: use a leading dot for object fields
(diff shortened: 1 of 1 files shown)


src/Vm.sol
```diff
@@ -1022,6 +1022,8 @@ interface VmSafe {
 
     // ======== JSON ========
 
+    /// JSON `key` parameters use selector syntax: use a leading dot for object fields
+    /// (for example, `.bar`) or `$` for the root object.
     /// Checks if `key` exists in a JSON object.
```
onbjerg:
```suggestion
    /// Checks if `key` exists in a JSON object.
    ///
    /// JSON `key` parameters use selector syntax: use a leading dot for object fields
    /// (for example, `.bar`) or `$` for the root object.
```

onbjerg (changes requested):

DaniPopes:
This file is autogenerated, please fix here: https://github.com/foundry-rs/foundry/blob/196c90a30d6a668be65e820cc319a20a8e4a2605/crates/cheatcodes/spec/src/vm.rs#L21

PR #887, opened 2026-07-11
fix(StdStorage): find string/bytes slots (dynamic ABI returns)

### LINKED ISSUE #345: StdStorage Slot find does not find storage slot for short strings.
Example contract: 

```sol
contract MyContract {
    string public exists = "thequickbrownfoxjumpsoverthelaz";
}
```

Script:

```sol
contract MyScript is Script {

    function run() public {
        uint256 deployerPrivateKey = vm.envUint("PRIVATE_KEY");
        vm.startBroadcast(deployerPrivateKey);

        MyContract s = new MyContract();

        uint256 slot = stdstore.target(address(s)).sig("exists()").find();
        console2.log(slot);
        vm.stopBroadcast();
```

Output: 
```
stdStorage find(StdStorage): Slot(s) not found.
```

Yonkoo11:
## Summary

`stdstore.find()` failed on public `string` / `bytes` state vars (short strings in particular): `Slot(s) not found.`

Cause: mutation checks only compared one returndata word at `depth`. For dynamic ABI returns that word is the offset (`0x20`), not the content, so writing storage never registered as a hit. The follow-up `slot value == return word` check has the same mismatch (short-string packing is not the ABI word).

## Change

- Detect single dynamic returns (`returndata` length ≥ 64 and word0 == `0x20`)
- For those, probe slots with a full returndata hash compare
- Skip the raw slot/return equality check in that path
- Static multi-word returns (structs) still compare only the word at `depth` so sibling fields are not false hits

## Tests

- `test_StorageFindShortString` — 31-char public string, expects slot 0 (#345)
- `test_StorageFindLongString` — find must not revert
- `test_StorageFindBytes` — public `bytes`, expects slot 0
- Existing `test/StdStorage.t.sol` suite (incl. fuzz) green locally

## Notes

Closes #345

Prepared with AI assistance; I reviewed and ran the tests above.

### DIFF
--- a/src/StdStorage.sol
+++ b/src/StdStorage.sol
@@ -51,20 +51,62 @@ library stdStorageSafe {
         return (success, result);
     }
 
+    /// @notice Staticcall the configured target and return raw returndata.
+    function callTargetRaw(StdStorage storage self) internal view returns (bool success, bytes memory rdat) {
+        bytes memory cd = abi.encodePacked(self._sig, getCallParams(self));
+        (success, rdat) = self._target.staticcall(cd);
+    }
+
+    /// @notice True when returndata is a single ABI dynamic value (string/bytes): word0 == 0x20 offset.
+    function _isSingleDynamicReturn(bytes memory rdat) private pure returns (bool) {
+        if (rdat.length < 64) {
+            return false;
+        }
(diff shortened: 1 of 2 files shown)


Yonkoo11:
Closing — no maintainer engagement after 2 weeks. The fix is still valid if someone wants to pick it up.

PR #894, opened 2026-08-06
fix(StdStorage): find short string and bytes slots

### LINKED ISSUE #345: StdStorage Slot find does not find storage slot for short strings.

stevencartavia:
Fixes `stdStorage.find()` for non-empty short `string` and `bytes` values of 1–31 bytes.

The implementation matches Solidity’s short-value storage encoding and confirms the candidate slot by temporarily clearing and restoring it. Existing static, packed-slot, and `.depth()` behavior remains unchanged.

Fixes #345.

### DIFF
--- a/src/StdStorage.sol
+++ b/src/StdStorage.sol
@@ -22,6 +22,12 @@ struct StdStorage {
 }
 
 library stdStorageSafe {
+    struct FindCallData {
+        bytes32 result;
+        bytes32 shortBytesStorageValue;
+        bytes32[] reads;
+    }
+
     event SlotFound(address who, bytes4 fsig, bytes32 keysHash, uint256 slot);
     event WARNING_UninitedSlot(address who, uint256 slot);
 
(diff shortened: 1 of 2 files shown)


src/StdStorage.sol
```diff
@@ -118,29 +161,40 @@ library stdStorageSafe {
...
+                bytes32 slot = callData.reads[i];
+                bytes32 prev = vm.load(who, slot);
                 if (prev == bytes32(0)) {
-                    emit WARNING_UninitedSlot(who, uint256(reads[i]));
+                    emit WARNING_UninitedSlot(who, uint256(slot));
                 }
 
-                if (!checkSlotMutatesCall(self, reads[i])) {
+                bool shortBytesFound = callData.shortBytesStorageValue != bytes32(0)
+                    && prev == callData.shortBytesStorageValue && _checkShortBytesSlot(self, slot);
```
figtracer: this exact comparison rejects valid short bytes/string slots with dirty unused padding: a slot ending in `...deadbe06` still returns `0x616263`, but `find()` reverts. please compare only the payload and length byte before probing.

mablr: +1

stevencartavia: [2f24033](https://github.com/foundry-rs/forge-std/pull/894/commits/2f24033dcb37fcc18472f5c52fd8668d5ef9e8a2)

test/StdStorage.t.sol
```diff
@@ -351,6 +351,37 @@ contract StdStorageTest is Test {
...
+        ShortBytesStorage target = new ShortBytesStorage();
+        assertEq(stdstore.target(address(target)).sig(target.exists.selector).find(), 0);
+    }
+
+    function test_StorageFindShortBytesWithDirtyPadding() public {
+        ShortBytesStorage target = new ShortBytesStorage();
+        bytes32 slotValue = vm.load(address(target), bytes32(uint256(1)));
+        vm.store(address(target), bytes32(uint256(1)), slotValue | bytes32(uint256(0xDEADBE) << 8));
+        assertEq(target.data(), hex"616263");
+        assertEq(stdstore.target(address(target)).sig(target.data.selector).find(), 1);
```
mablr, line 366: Could we assert that the raw slot still equals the dirty value after `find()`? `_checkShortBytesSlot` temporarily clears the candidate, and preserving the exact word (including dirty padding) is a core invariant of this fix. The existing restoration test only covers a failed probe, so saving the dirty word here and comparing `vm.load(...)` afterward would cover the successful path before merge.

mablr (commented):
LGTM, just one non-blocking suggestion

test/StdStorage.t.sol
stevencartavia: [f6ec2a3](https://github.com/foundry-rs/forge-std/pull/894/commits/f6ec2a3d27c5aed14d11f335bc2f1775b8be26a6)

mablr (approved):

mattsse (approved):
lgtm

PR #907, opened 2026-08-28
chore: sync state gas field

figtracer:
Sync `Vm.Gas` with Foundry's six-field return ABI by documenting `gasTotalUsed` as regular gas and exposing `gasStateUsed` for net state gas. This keeps Forge Std callers aligned with the state-gas reporting added in foundry-rs/foundry#16387.

AI assistance was used to implement and test this change.

### DIFF
--- a/src/Vm.sol
+++ b/src/Vm.sol
@@ -258,7 +258,7 @@ interface VmSafe {
     struct Gas {
         // The gas limit of the call.
         uint64 gasLimit;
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm but this is technically breaking 
but we already have in on foundry like this
ref https://github.com/foundry-rs/foundry/pull/16442

so for next 1.x foundry release we must publish this as well

